PR #214, opened 2015-04-23
Added support for rotation in geoTIFFs ( fixes #213 )

### LINKED ISSUE #213: Geotiff import does not handle raster rotations
I tried importing a geotiff with rotation and it failed to rotate.
I dug a bit in the code and found that the values are correctly read at [RasterGridFilter](https://github.com/cloudcompare/trunk/blob/master/libs/qCC_io/RasterGridFilter.cpp#L80-88)

Unfortunately the rotation is not applied at [Y coord](https://github.com/cloudcompare/trunk/blob/master/libs/qCC_io/RasterGridFilter.cpp#L116) and  [X coord](https://github.com/cloudcompare/trunk/blob/master/libs/qCC_io/RasterGridFilter.cpp#L122) 

the code should follow GDAL suggestion that is 
```c++
   Xp = padfTransform[0] + P*padfTransform[1] + L*padfTransform[2];
   Yp = padfTransform[3] + P*padfTransform[4] + L*padfTransform[5];
```

it seems to me pretty straightforward to implement.
To test it it's enough to load a geotiff that has the accompanying tfw set to have non-null 2nd and 3d component, as in
```
-0.005191
0.008547
0.008547
0.005191
66666.410
66666.120
```

[link to GDAL reference](http://www.gdal.org/classGDALDataset.html#af9593cc241e7d140f5f3c4798a43a668)

ramassin:
Tested and it seems to work fine.
Don't know which branch you prefer to pull it (if any)

image

### DIFF
--- a/libs/qCC_io/RasterGridFilter.cpp
+++ b/libs/qCC_io/RasterGridFilter.cpp
@@ -113,19 +113,14 @@ CC_FILE_ERROR RasterGridFilter::loadFile(QString filename, ccHObject& container,
 				double z = 0.0 /*+ Pshift.z*/;
 				for (int j=0; j<rasterY; ++j)
 				{
-					double y = adfGeoTransform[3] + static_cast<double>(j) * adfGeoTransform[5] + Pshift.y;
-					CCVector3 P(	0,
-									static_cast<PointCoordinateType>(y),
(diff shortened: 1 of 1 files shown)


cloudcompare:
Thanks!

PR #218, opened 2015-06-03
qRANSAC: change from #include <malloc.h> to #include <stdlib.h> (Mac)

PabloCOVES:
Conditional include directive in order to make the qRANSAC plugin work on Mac OS.
Compiles and works fine on MacOS 10.10.2 and up-to-date ArchLinux.


dgirardeau:
Thanks!

PR #234, opened 2015-10-15
Small fixes

luca-penasa:
a couple of fixes. 

In particular it looks like that if computeMinMax is called on a field before to add it to its point cloud does not work as expected, the min max is not computed and the scalar field is not colored and remains gray...

in mainwindow.cpp I fixed it moving the call, but maybe a scalar field on which the computeMinMax() method has been called should be add-able  to its cloud without requiring to add the field to the cloud before... 

To check this problem just call doActionScalarFieldFromColor() and look at the output. In my case the fields remains gray with min = max =0 in the range of SF display params in the UI.

This is probably a recenly-introduced unintended behavior cause in the past the action was working as expected...

Luca

### DIFF
--- a/plugins/qAnimation/qAnimationDlg.cpp
+++ b/plugins/qAnimation/qAnimationDlg.cpp
@@ -28,6 +28,7 @@
 #include <QSettings>
 #include <QElapsedTimer>
 #include <QProgressDialog>
+#include <QMessageBox>
 
 //standard includes
 #include <vector>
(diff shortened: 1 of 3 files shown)


cloudcompare:
Hi Luca!

In fact, the computeMinMax issue is different: it's because we call 'setValue' on the scalar fields while they have been 'reserved' and not 'resized'. Therefore, the internal number of values stays at 0 (and this is why they are ignored). We should call either 'resize' on the scalar fields first, or call 'addElement' afterwards.

And moving the instruction after the 'addScalarField' works because the cloud will force the scalar field size to be the same as its number of points (i.e. it will call 'resize').

I'll fix it. And I'll change the wrong includes as well.

As always, thanks for the feedback.

luca-penasa:
Thanks to you for your work! and happy to see qFacets released!

Issue #78, opened 2013-04-21
static const var issue


asmaloney:
A recent change highlights a problem with decalring static consts in classes.

DgmOctree has the following in the class declaration:

``` C++
   static const int MAX_OCTREE_LEVEL = 10;
```

Compilers usually treat this as a constant compile-time expression and just inlines the value.  But this means that you can't pass MAX_OCTREE_LEVEL to any function that takes a reference, such as this:

``` C++
   std::min<int>(7,CCLib::DgmOctree::MAX_OCTREE_LEVEL)
```

because the var isn't actually defined anywhere.  This may result in a link error (as it does with GCC).

In general, the solutions are to either (1) declare it in the header:

``` C++
   static const int MAX_OCTREE_LEVEL;
```

and define it in the cpp:

``` C++
#ifdef OCTREE_CODES_64_BITS
    const int CCLib::DgmOctree::MAX_OCTREE_LEVEL = 21;
#else
    const int CCLib::DgmOctree::MAX_OCTREE_LEVEL = 10;
#endif
```

(This won't work for DgmOctree since it uses it to declare arrays later in the class:

``` C++
   //! Cell dimensions for all subdivision levels
   PointCoordinateType m_cellSize[MAX_OCTREE_LEVEL+2];
```

)

(2) Add a definition in the cpp, but leave the initialization with the declaration:

``` C++
  const int CCLib::DgmOctree::MAX_OCTREE_LEVEL;
```

Or (3) simply use an enum:

``` C++
#ifdef OCTREE_CODES_64_BITS
  enum { MAX_OCTREE_LEVEL = 21 };
#else
  enum { MAX_OCTREE_LEVEL = 10 };
#endif
```

Daniel: I know you have an aversion to using enums like that, so I assume you'd prefer solution (2)? :-)

PR #235, opened 2015-10-19
Fix loading the QFACETS plugin when compiled with gcc

ao2:
The QFACETS plugin cannot be loaded when compiled with gcc because of
a missing symbol:

[11:48:20] [Plugin] Cannot load library .../plugins/libQFACETS_PLUGIN_DLL.so:
  (.../plugins/libQFACETS_PLUGIN_DLL.so: undefined symbol: _ZN12ccColorScale9MAX_STEPSE)

The error seems to be about plugins/qFacets/stereogramDlg.cpp using the
symbol ccColorScale::MAX_STEPS as a parameter of a template function
while the symbol is not properly exported.

Apparently gcc is quite strict about static integer members being
_declared_ in the class declaration but then not explicitly _defined_ in
the implementation.

Some references:
http://stackoverflow.com/questions/3025997/defining-static-const-integer-members-in-class-definition
https://gcc.gnu.org/wiki/VerboseDiagnostics#missing_static_const_definition
https://isocpp.org/wiki/faq/ctors#link-errs-static-data-mems

This is the minimal fix to have the plugin loaded successfully, but for
consistency other static member in the class definition may have to be
defined as well, or even made private if they do not need to be exposed.

### DIFF
--- a/libs/qCC_db/ccColorScale.cpp
+++ b/libs/qCC_db/ccColorScale.cpp
@@ -23,6 +23,8 @@
 //Local
 #include "ccLog.h"
 
+const unsigned ccColorScale::MAX_STEPS;
+
 ccColorScale::Shared ccColorScale::Create(QString name)
 {
 	return ccColorScale::Shared(new ccColorScale(name));


cloudcompare:
Ah! I can't count the number of times it has already happened... I really don't get why gcc is so strict about this. Thanks anyway.

dgirardeau:
And for the records this fix doesn't work on Visual :(

asmaloney:
This sounds similar to #78.  Maybe one of those solutions apply?

dgirardeau:
In this particular case I use the value of MAX_STEPS in the header itself (to define the size of the color table array). When I try to leave the declaration of MAX_STEPS in the header but I assign it a value in the cpp, Visual complains that the array size is not defined (and I guess gcc would do the same).

For now I simply added a '#ifndef _MSC_VER' in the cpp flie...

ao2:
@dgirardeau hi, what was the exact error from MSVC?

The ```ifndef``` looks a lot like a workaround, maybe a proper solution can be found.

Indeed, I still left the assignment in the header because gcc too would have complained with the assignment in the cpp file (and that makes sense to me, the value is used in the header after all).

BTW, putting verbatim errors in commit messages helps other developers to see and validate what a fix is exactly about. Don't be afraid to write long commit messages :smile: 

Thanks,
   Antonio

dgirardeau:
Oh the error was very simple:
- if I apply your patch, then I get a duplicate symbol error at linking time (stating that ccColorScale::MAX_STEPS is already defined in cc2DLabel.obj)
- if I leave the declaration in the header file, and I assign the value in the cpp file only, I get a compilation error because the value of MAX_STEPS is not defined and is required for the definition of the 'm_rgbaScale' array (obviously).

The thing that bugs me is that we do exactly the same thing in CC_CORE_LIB (see MAX_OCTREE_LEVEL in DgmOctree.h for instance). Do you get any compilation error for this one on your side?

P.S.: my verbatim error messages are in French ;)

ao2:
OK, the cause of the original error is explained here (see in particular the comment of the user Crazy Eddie):
http://stackoverflow.com/questions/3025997/defining-static-const-integer-members-in-class-definition

The out-of-class definition is needed by the *standard* when the **address** of the variable is used somewhere, in our case that happens in plugins/qFacets/stereogramDlg.cpp calling std::min().

Either Visual C++ is being liberal and allows omitting the out-of-class definition, or maybe the `std::min()` implementation is different, I do not know.

A more standard compliant behavior could be enabled using the `/Za` option as described here:
http://stackoverflow.com/questions/5602425/defining-integral-static-const-in-a-class-whose-h-file-is-included-in-multiple

and here:
https://msdn.microsoft.com/en-us/library/34h23df8.aspx

This way Visual C++ should become more standard compliant and the same code should work on more compilers. But this has to be tested and I don't have Windows here.

@dgirardeau can you test if using `/Za` fixes the problem and allows to remove the ugly `#ifndef`s?

A change like that shoudl do it, but I've never used Visual Studio, so double check please :)

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 8fe6d25..08bf5bf 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,6 +26,7 @@ endif()
 # Default debug suffix for libraries
 if(MSVC)
   set( CMAKE_DEBUG_POSTFIX d )
+  set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /Za")
 endif()

 if( UNIX )
```

Thanks,
   Antonio

dgirardeau:
Sorry for the delay. I finally took the time to test the compilation with '/Za'... and it's a nightmare (> 1500 errors, even 'winnt.h', 'winuser.h' and 'wingdi.h' cause dozens of errors - mostly about the usage of anonymous struct for instance...).

But following the comments on stackoverflow, we can simply avoid using those static const definitions as 'by ref' parameters. I know that it won't prevent this issue to happen again but at least we will be aware of it (hopefully :-)).

asmaloney:
@dgirardeau You could always switch the Windows build to [MinGW](http://www.mingw.org/) so we're all using similar compilers :-)

I actually use it with Qt Creator on Windows for my own work - once it's all set up it's fine.

PR #237, opened 2015-10-21
fix debug compile on linux

luca-penasa:
The line

```cmake
if( NOT CMAKE_CONFIGURATION_TYPES )
	set_property( TARGET ${ARGV0} APPEND PROPERTY COMPILE_DEFINITIONS ${CC_DEFAULT_PREPROCESSORS_RELEASE} )
else()
```
was preventing debug-compile on linux, where most of cmake generators are single-config.  
In the specific the same CC_DEFAULT_PREPROCESSORS_RELEASE options were applied to all the targets, independently by the mode (debug or release) the user is configuring for. Thus if the user expect a debug compile in fact the options were forcing a release mode.

I see there are other places in the same file where CMAKE_CONFIGURATION_TYPES var is used to make decision on what to do... But from a rapid inspection these points seem not critical.

BTW The function
```cmake
function( set_default_cc_preproc ) 
```
Is there a reason for setting the default definitions for each target we are compiling, instead of setting the definitions globally? For example altering the variables:

```cmake
CMAKE_CXX_FLAGS 
    the compiler flags for compiling C++ sources. Note you can also specify switches with ADD_DEFINITIONS(). 
CMAKE_CXX_FLAGS_DEBUG 
    compiler flags for compiling a debug build from C++ sources. 
CMAKE_CXX_FLAGS_RELEASE 
    compiler flags for compiling a release build from C++ sources. 
CMAKE_CXX_FLAGS_RELWITHDEBINFO 
    compiler flags for compiling a release build with debug flags from C++ sources. 
```
E.g. if NOMINMAX and _CRT_SECURE_NO_WARNINGS are always needed we can put them in the main CMakeLists.txt file as 
```cmake
add_definitions(-DNOMINMAX -D_CRT_SECURE_NO_WARNINGS)
```

Well.. if its ok I'll try to prepare a branch with these changes to see if it works correctly

### DIFF
--- a/CMakeInclude.cmake
+++ b/CMakeInclude.cmake
@@ -180,28 +180,25 @@ endfunction()
 
 # Default preprocessors
 function( set_default_cc_preproc ) # 1 argument: ARGV0 = target
-set( CC_DEFAULT_PREPROCESSORS NOMINMAX _CRT_SECURE_NO_WARNINGS ) #all configurations
-set( CC_DEFAULT_PREPROCESSORS_RELEASE NDEBUG ) #release specific
-set( CC_DEFAULT_PREPROCESSORS_DEBUG _DEBUG ) #debug specific
+    set( CC_DEFAULT_PREPROCESSORS NOMINMAX _CRT_SECURE_NO_WARNINGS ) #all configurations
+    set( CC_DEFAULT_PREPROCESSORS_RELEASE NDEBUG ) #release specific
+    set( CC_DEFAULT_PREPROCESSORS_DEBUG _DEBUG ) #debug specific
 
-if (MSVC)
-	#disable SECURE_SCL (see http://channel9.msdn.com/shows/Going+Deep/STL-Iterator-Debugging-and-Secure-SCL/)
-	list( APPEND CC_DEFAULT_PREPROCESSORS_RELEASE _SECURE_SCL=0 )
+    if(MSVC)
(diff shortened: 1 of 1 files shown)


cloudcompare:
Yes please go ahead. All this CMake code has been modified in a quite random way and by various people (or the same person with various states of mind ;). It would be good to get a more consistent behavior.

luca-penasa:
can we move straight to qt5 only compiles? this would do a nice cleaning of cmake files...
I cannot see a good reason for anybody to really need qt4 instead of qt5... 

What do you think? Maintaining both is always more work...

luca-penasa:
I mean... it is not really a big problem maintaining both.. I just wanted to see if it is still important to maintain both in your (or other users) opinion...

cloudcompare:
No! Qt 5 is far from being stable yet. In the 2.6.1 version we had half a dozen bugs with Qt 5 (at least on Windows) . The 2.6.2 now uses Qt 4 again.

The biggest issue is that most of the bugs come from the bad portage of the QGLWidget widget with their new mechanism. And they wont fix it (at least it's still buggy with the 5.5 version) as the aim is to use the new QOpenGLWidget mechanism... But porting the existing code to the new mechanism will take a long time. So for now, let's keep Qt4 :)

Or I can put it differently: I'll drop Qt4 when someone will port the existing QGLWidget to QOpenGLWidget (without too much loss of functionality of course ;).

luca-penasa:
OK! thanks for the explanation!

PR #243, opened 2015-11-07
Use CGAL for Delaunay Triangulation and Constrained Delaunay Triangulation

ORNis:
cc @Wookey

As pointed by Debian FTPMaster, the "not so free" license of Triangle library is not compatible with the LGPL/GPL license scheme adopted by CC.
This commit allow to use CGAL, a well established (and packaged) LGPL/GPL library of geometric processing instead of Triangle where needed.
This implementation is *a priori* on parity with Triangle one with regards of the repetability. It's a little bit faster and maybe it have a small memory overhead.
Alternatively Triangle can still be used (and is the default build choice in the CMakeList.txt)  to allow a smooth transition. But it can be safely removed in order to eventually push a DSFG package of CloudCompare.

Travis script is updated to use CGAL instead of Triangle.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,17 +6,18 @@ compiler:
     - gcc # 4.6
 
 install:
-    # We remove Postgis added by Travis since it prevent us from installing libgdal-dev
+    # We remove Postgis added by Travis since it prevents us from installing libgdal-dev
     - sudo mv /etc/apt/sources.list.d/pgdg-source.list* /tmp/
     - sudo apt-get -qq remove postgis
     - sudo apt-get update -qq
     # Install dependencies
-    - sudo apt-get install -qy cmake libqt4-dev libqt4-opengl-dev libgdal-dev
+    - sudo apt-get install -qy cmake libqt4-dev libqt4-opengl-dev libgdal-dev libcgal-dev
 
 before_script:
     - mkdir build
     - cd build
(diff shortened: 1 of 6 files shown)


cloudcompare:
Many thanks!

wookey:
On 2015-11-07 06:30 -0800, Romain Janvier wrote:
> cc @Wookey
> 
> As pointed by Debian FTPMaster, the "not so free" license of Triangle library
> is not compatible with the LGPL/GPL license scheme adopted by CC.
> This commit allow to use CGAL, a well established (and packaged) LGPL/GPL
> library of geometric processing instead of Triangle where needed.
> This implementation is a priori on parity with Triangle one with regards of the
> repetability. It's a little bit faster and maybe it have a small memory
> overhead.
> Alternatively Triangle can still be used (and is the default build choice in
> the CMakeList.txt) to allow a smooth transition. But it can be safely removed
> in order to eventually push a DSFG package of CloudCompare.
> 
> Travis script is updated to use CGAL instead of Triangle.

OK. I just woke up to this (nearly a year ago!). In the meantime I see
we have 2.6.3 and 2.7.0 released.

Your ubuntu packaging has been updated to 2.6.3. Is there any reason
not to put that in Debian now (with the above fix)?

How is your 2.7.0 packaging coming along? 

Wookey
-- 
Principal hats:  Linaro, Debian, Wookware, ARM
http://wookware.org/

PR #247, opened 2015-11-12
fix #245

### LINKED ISSUE #245: Random subsample is not random for big cloud
Hi

Random sub-sample is not working correctly with big clouds = 150 Mega Points.

Only first aprox. 15 Mpoints are sub-sampled correctly but then probability of point will be removed is higher as should be.

My test:
-generate cloud with 150 M points, points [0, 0, 0], [1, 0, 0] ... [149 000 000, 0, 0]
-then sub-sample randomly to 10 M points
-generate histogram of X values (1000 bins in my case)

You could see that at the end of cloud there is much less points as should be. Cloud is less dense at the and of cloud.

cc random subsample

X axis = bins of X values of sub-sampled cloud
Y axis = counts of points in bin

PeterJCC:
replaced standard rand() with mt19937

### DIFF
--- a/CC/src/CloudSamplingTools.cpp
+++ b/CC/src/CloudSamplingTools.cpp
@@ -28,6 +28,7 @@
 #include "DgmOctreeReferenceCloud.h"
 #include "DistanceComputationTools.h"
 #include "ScalarFieldTools.h"
+#include <random>
 
 //system
 #include <assert.h>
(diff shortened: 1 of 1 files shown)


cloudcompare:
Great! Is it working better this way?

PeterJCC:
Yes, here is histogram subsample 150 Mpoints to 10 Mpoints.
Speed is same.
cc random subsample_mt19937

cloudcompare:
Nice!

It works with Visual 2012 but I get an error from Travis (with gcc and c-lang).

With which compiler are you working on your side?

PeterJCC:
I used Visual Studio Express 2012 for Windows Desktop Version 11.0.50727.42

Travis: gcc compiler flag should be C++0x or C++11 for random function

cloudcompare:
Okay we'll have these flags then.

However it's strange that one must explicitly add these flags if the compiler does support C++x11!

PR #259, opened 2015-12-16
Compile definitions

luca-penasa:
as in commits description, works well on linux, to be checked with a fresh build on win

### DIFF
--- a/CMakeSetCompilerOptions.cmake
+++ b/CMakeSetCompilerOptions.cmake
@@ -1,0 +1,35 @@
+# these are always defined
+add_definitions(-DNOMINMAX -D_CRT_SECURE_NO_WARNINGS)
+
+# cmake should be smart enough to automatically append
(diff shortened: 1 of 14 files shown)


CMakeSetCompilerOptions.cmake
```diff
@@ -0,0 +1,35 @@
+# these are always defined
+add_definitions(-DNOMINMAX -D_CRT_SECURE_NO_WARNINGS)
```
asmaloney: Aren't these Windows things?  Shouldn't this be moved inside `if( MSVC )`?

luca-penasa: I suspect NOMINMAX is also needed on UNIX, while D_CRT_SECURE_NO_WARNINGS is totally msvc stuff.. I'll try to compile on linux without nominmax defined to see if something bad happens :-)

luca-penasa: looks like its msvc-related stuff only

PR #260, opened 2015-12-16
getting rid of the global cmake var EXTERNAL_LIBS_LIBRARIES. It was c…

luca-penasa:
…ollecting just a couple of libraries (qt plus opengl)

But not all the plugins (or other target) need to be linked against OpenGl and not all plugins need to be linked against all the qt libs. I removed this variable in favour of explicit target linking, in the specific:
 - when using qt4 the target is linked against ${QT_LIBRARIES}
 - when on qt5 the linkage is specified by the qt5_use_modules function.
 - Link to openGl is made only where necessary
Furthermore OpenGL include is made when the package has been found.

there is no need to keep a find_mocable_files function i think (?).
CMake has its own automatic wrapping system that automatically wraps cpp files as needed. It is now enabled: see set(CMAKE_AUTOMOC ON) this should work both with qt4 and qt5

### DIFF
--- a/CMakeExternalLibs.cmake
+++ b/CMakeExternalLibs.cmake
@@ -2,20 +2,18 @@
 # Qt
 # ------------------------------------------------------------------------------
 option( USE_QT5 "Check to use Qt5 instead of Qt4" OFF )
+## we will use cmake automoc feature
+set(CMAKE_AUTOMOC ON)
+set(CMAKE_INCLUDE_CURRENT_DIR ON)
+
 if ( USE_QT5 )
 
 	cmake_minimum_required(VERSION 2.8.8)
 
-	set( QT5_ROOT_PATH CACHE PATH "Qt5 root directory (i.e. where the 'bin' folder lies)" )
-	if ( QT5_ROOT_PATH )
-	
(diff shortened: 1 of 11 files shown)


cloudcompare:
Luca, it doesn't work on Windows.

Are you sure about the  `find_package(Qt5 ...`? CMake (3.2) can't find anything on my side. And when looking at Qt's documentation (http://doc.qt.io/qt-5/cmake-manual.html) they use the following syntax;
```
find_package(Qt5Widgets)
```

cloudcompare:
I confirm that replacing the `find_package(Qt5...)` by the following works:
```
	find_package(Qt5Widgets)
	find_package(Qt5Core)
	find_package(Qt5Gui)
	find_package(Qt5PrintSupport)
	find_package(Qt5Concurrent)
```

cloudcompare:
+ 	find_package(Qt5OpenGL)

luca-penasa:
Sorry, seen your comments right now... 
On linux that syntax was working, but it is totally equivalent I think.

It is strange that something was missing opengl includes (it was working fine on linux) because the include is made at the higher cmake level, so all subprojects should find them (?)

PR #261, opened 2015-12-16
openmp test on linux

luca-penasa:
I made a simple test of using openmp, it is enabled on linux, only if openmp is automatically found, but it gives a nice speed up for point picking and segmentation of big clouds... I think at least on linux will be appreciated...

Isn't there any free windows compiler supporting it? maybe there are much more tasks that can get a nice speed up from it...

What do you guys think?

### DIFF
--- a/CMakeExternalLibs.cmake
+++ b/CMakeExternalLibs.cmake
@@ -78,6 +78,18 @@ include_directories(${OpenGL_INCLUDE_DIR})
 #    endif()
 #endif()
 
+# ------------------------------------------------------------------------------
+# OpenMP (only on UNIX for now)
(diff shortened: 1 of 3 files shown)


luca-penasa:
I've followed the discussion in the other thread....

Any example of parallelizing the same for qtconcurrent?  How much is it different? I suspect one would have to rewrite the code more or less heavily...

cloudcompare:
If we use C++x11 lambdas it can be very easy. But it will only compile on 'modern' compilers...

Does OpenMP handles the concurrent access to the variables?

asmaloney:
C++11 is fine for me on the Mac.  I use it for my own work.

PR #270, opened 2016-01-21
ffmpegSupport.cmake message function call fix

pushrbx:
I've just found you guys, and I wanted to create a custom build of the app, but I got a problem with a cmake.
With cmake 3.4.1 I got a error message saying that the message function has been called with incorrect number of arguments. So I've fixed it. Maybe it is working with older versions of cmake if you use the message function with one argument, but I haven't tested it.


cloudcompare:
Thanks.

In fact the issue comes from the fact that there's no DLL file listed in FFMEG_DLL. And 'message' always sends an error when an empty list is provided.

Not sure if it's normal or not in your case ;)

PR #271, opened 2016-01-27
tracing tool

luca-penasa:
I had to make some changes around. Maybe what I've chosen was not the best approach so its up to you guys to see if its ok...

It allows to "upsample" the traced polyline... this is also nice for creating sections of point clouds surfaces. With the new octree-based intersection works great!


cloudcompare:
Oups I should have checked this in a branch before!

It doesn't work well with the new picking mechanism. And anyway there's a lot of things that should be changed... I'm working on it right now.

PR #293, opened 2016-03-03
CMake

ORNis:
Build failed on linux. I fixed it by explicitly define CXX language in the project() macro.
cc @asmaloney for review.


ORNis:
f

PR #364, opened 2016-04-15
Fix (?) LOCALE issue on Unix systems

ORNis:


ccViewer/main.cpp
```diff
@@ -120,15 +120,12 @@ int main(int argc, char *argv[])
 	}
 
 	ccApplication a(argc, argv);
-	
-	//Force 'english' local so as to get a consistent behavior everywhere
-	QLocale::setDefault(QLocale::English);
```
cloudcompare: Why did you removed this? We still need it even on Windows

ORNis: it was commented on the qCC main.cpp

cloudcompare: I mean this one or the one after (setNumberOptions ;)

ccViewer/main.cpp
```diff
@@ -120,15 +120,12 @@ int main(int argc, char *argv[])
...
 
 	// We reset the numeric locale.
 	// See http://qt-project.org/doc/qt-5/qcoreapplication.html#locale-settings
-	QLocale locale = QLocale::system();
-	locale.setNumberOptions( QLocale::c().numberOptions() );
-	QLocale::setDefault( locale );
+#ifdef Q_OS_UNIX
+	setlocale(LC_NUMERIC, "C");
+#endif
```
ORNis: QLocale::setDefault( locale ) ?

cloudcompare: Yep. We need to force the usage of the dot as decimal character whatever the case.

ORNis: If I understand well, on my computer
QLocale locale = QLocale::system(); => FR
locale.setNumberOptions( QLocale::c().numberOptions() ); => dot instead of commas
QLocale::setDefault( locale ); => we make this append.

it's ok, but it does not work on Linux (for ASCII files and my inputs still require commas). Moreover I though it was the trick @asmaloney use to replace the POSIX setLocale() trick I bring back in this PR.

PR #377, opened 2016-04-27
dbtree objects can be selected by custom type + by regex

luca-penasa:


qCC/ccSelectChildrenDlg.h
```diff
@@ -41,11 +41,16 @@ class ccSelectChildrenDlg : public QDialog, public Ui::SelectChildrenDialog
...
 	//! Returns the selected type
 	CC_CLASS_ENUM getSelectedType();
+
 	//! Returns the selected name (if any)
 	QString getSelectedName();
+
 	//! Returns the state of the strict type checkbox
 	bool getStrictMatchState();
 
+    //! if the name must be considerd as regex
```
asmaloney: Your tabs/spacing seem to be off?

Tabs only - tab size of 4.

qCC/db_tree/ccDBRoot.cpp
```diff
@@ -1700,6 +1704,7 @@ void ccDBRoot::selectByTypeAndName()
 	//For generic-only types the match type gets overridden and forced to
 	//false because exclusive match makes no sense!
 	bool exclusive;
+    bool regex;
```
asmaloney: Need to initialize this to `false`.

luca-penasa: Also having other problems with the type selection. What kind of operator do we want to use? 

I mean if we disable the type search the regex/name search is not finding anything right now cause the filtering is always made before name-matching. I'll  try to put together a better proposal...

qCC/ccSelectChildrenDlg.h
luca-penasa: Yep sorry, saw it right now :-(

qCC/db_tree/ccDBRoot.cpp
luca-penasa: bool regex is initialized, as well the exclusive var, into the default case next lines. but maybe is better to get it out of it...

asmaloney: That's the problem - it's _only_ initialized in the default case.

luca-penasa: Ahhh sure! thanks

asmaloney: I would recommend that, as a general practice, _all_ variables are initialized at declaration (when possible).

Then you know it's done, it signals to other developers what the intention is, and it save other devs from having to parse a bunch of code in their head.

It will save a lot of time and bug hunting - I guarantee it!

PR #378, opened 2016-04-27
introduces regex-selection of tree objects

luca-penasa:
this should be a better version, let me know if anybody sees problems with it 

- some line indentation cleaning 
- possibility to use regex for name matching
- when not using object name matching nor type matching all the objects are selected
- one can now select all custom types (e.g. objects from plugins)

### DIFF
--- a/qCC/db_tree/ccDBRoot.cpp
+++ b/qCC/db_tree/ccDBRoot.cpp
@@ -1725,7 +1725,7 @@ void ccDBRoot::selectByTypeAndName()
 
 	// for name matching - def values
 	bool regex = false;
-	QString name (""); // an empty string by default
+	QString name; // an empty string by default
 
 	if (scDlg.getNameMatchIsUsed())
 	{
--- a/qCC/db_tree/ccDBRoot.cpp
+++ b/qCC/db_tree/ccDBRoot.cpp
@@ -26,6 +26,7 @@
 #include <QHeaderView>
 #include <QMimeData>
 #include <QMessageBox>
+#include <QRegExp>
 
 //qCC_db
 #include <ccLog.h>
@@ -66,7 +67,7 @@ static bool CanDetachCloud(const ccHObject* obj)
 		assert(false);
 		return false;
 	}
-	
+
 	ccHObject* parent = obj->getParent();
 	if (!parent)
 	{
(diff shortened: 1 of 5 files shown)


qCC/db_tree/ccDBRoot.cpp
```diff
@@ -1687,37 +1690,58 @@ void ccDBRoot::selectByTypeAndName()
...
-	//some types are exclusive, some are generic, and some can be both
-	//(e.g. Meshes)
-	//
-	//For generic-only types the match type gets overridden and forced to
-	//false because exclusive match makes no sense!
-	bool exclusive;
-	switch (type)
+	// for name matching - def values
+	bool regex = false;
+	QString name = "";
```
asmaloney: In this case, the QString class has a constructor that's responsible for initialization, so you don't really want to do this.  

Built-in types - bool, int, float, etc. - do not initialize themselves - unless you are compiling with debug.  Some - most? - compilers will initialize these in that case.  This is why you sometimes have bugs in release versions that you can't reproduce in debug compiles, and another reason you should always initialize built-ins.

If you did want to initialize a QString to some string, it would probably be better like this:

`QString   name( "foo" );`

luca-penasa: I just wanted to initialize QString to an empty string, I used a "" to be clear that is actually my intention to initialize it as empty, and I did not simply forgot to initialize it.
totally ok to have 
```cpp
QString name("")
```
 or even 
```cpp
QString name;
```
will just work right, the string will be initialized as empty. I found the use of a copy initializer more clear about my intentions :-) If I am not wrong,  in such a case the def constructor will be never called and the closer match converting constructor will be called instead. thus not adding overhead to the QString initialization.

I am referring to this doc: http://en.cppreference.com/w/cpp/language/copy_initialization

I am going to change it to a QString name(""); that is equally clear :+1:

asmaloney: I don't see any point in writing `QString name("");` when it's the same as `QString name;` except that the former _might_ or _might not_ add overhead and the latter is cleaner.  Your intentions are not any clearer writing it like that.  In fact it can be confusing.  I'd be asking myself "did he forget to put some text in there?".

cloudcompare: Moreover in the first case `QString::isNull()` will return false.

Of course if you only use `QString::isEmpty()` it's good (it will work in both cases).

asmaloney: True, but if you're relying on the difference between the two, you might want to rethink the code.

cloudcompare: My point was that the `QString name;` was more generic :D

luca-penasa: ok changed to QString name;

trying to understand why I wrote that is cause in these lines

```cpp
	// for name matching - def values
	bool regex = false;
	QString name; // an empty string by default
```
I am defining what will be the default values for two variables.. Thus I found clearer to state that _name_ is an empty string (not just a default constructed one...) and that _regex_ is false. 

Its like saying "I am not leaving the default values undefined - or def constructed -, but I am defining it as a length zero string... if you need you can write there something else if you need to change the default value".. that was my idea , probably:-)

thanks for the help btw

PR #379, opened 2016-04-30
Move history in CHANGELOG.md

ORNis:
I tried to convert history.txt file into CHANGELOG.md. MD formating seems OK, and I also take care of replacing reference to history.txt file into the build system.


asmaloney:
Are you going to still include the .md file in distributions?  Or just have it online?  I would vote for the latter - most people won't be able to read .md files (though this one seems readable - the extension won't make sens to them).

ORNis:
Hmm you are probably right. maybe "CHANGELOG" without extension make more sense.

asmaloney:
I'm actually in favour of using the CHANGELOG.md file - "github standard" and just not including it in the distribution.

ORNis:
I think github handle text file without extension as MD files. I'm not sure either...  But if @dgirardeau still  wants to ship it, it can be a good compromise. On my side, I will continue to deliver it.

PR #391, opened 2016-05-26
fix transformation of facets

luca-penasa:
- change min-max values in cell fusion dlg
- normal are shown as sqrt(polyline-length)

### DIFF
--- a/libs/qCC_db/ccFacet.cpp
+++ b/libs/qCC_db/ccFacet.cpp
@@ -415,7 +415,11 @@ void ccFacet::drawMeOnly(CC_DRAW_CONTEXT& context)
 		markerContext.display = 0;
 
 		c_unitNormalSymbol->setTempColor(m_contourPolyline->getColor());
-		PointCoordinateType scale = m_contourPolyline->getOwnBB().getMinBoxDim();
+
+        //		PointCoordinateType scale = m_contourPolyline->getOwnBB().getMinBoxDim();
+
+        // maybe using surface sqrt is better?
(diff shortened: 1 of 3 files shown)


plugins/qFacets/cellsFusionDlg.ui
```diff
@@ -299,9 +305,7 @@ Output facet normals may be randomly oriented
    <header location="global">ccOctreeSpinBox.h</header>
   </customwidget>
  </customwidgets>
- <resources>
-  <include location="qFacets.qrc"/>
```
luca-penasa: For some reason qtdesigner removed this. It was not intended but it did that by its own.

cloudcompare: Can you fix this?

libs/qCC_db/ccFacet.cpp
```diff
@@ -415,7 +415,11 @@ void ccFacet::drawMeOnly(CC_DRAW_CONTEXT& context)
 		markerContext.display = 0;
 
 		c_unitNormalSymbol->setTempColor(m_contourPolyline->getColor());
-		PointCoordinateType scale = m_contourPolyline->getOwnBB().getMinBoxDim();
+
+        //		PointCoordinateType scale = m_contourPolyline->getOwnBB().getMinBoxDim();
+
+        // maybe using surface sqrt is better?
+        PointCoordinateType scale = sqrt(m_contourPolyline->computeLength());
```
asmaloney: It looks like your space/tabs are off throughout this PR.

luca-penasa:
I'll put all together in a single commit :-)

PR #395, opened 2016-05-31
Bug fix: OpenGL rendering on retina displays

D-Alex:
Wrongly sized display buffers generated a wrong view
onto the scene. Therefore, all 3D interactions were not in
sync with the internal modelview matrix.


cloudcompare:
Awesome. Thanks for the fix!

PR #396, opened 2016-06-01
[IO/ASCII] Prepends cloud name

### LINKED ISSUE #394: Request --> IO --> ASCII --> cloud name in comments
Hello,

I am using the ASCII export for project export, could it be possible to add the cloud name as a comment line (like the column labels). It would be useful when loading the ASCII files with other soft (python, matlab...).

ex : 
// here_goes_the_name
// X Y Z
0.101 0.22 6.7
0.8 ...

Thanks.

ORNis:
close #394

### DIFF
--- a/libs/qCC_io/AsciiFilter.cpp
+++ b/libs/qCC_io/AsciiFilter.cpp
@@ -180,6 +180,7 @@ CC_FILE_ERROR AsciiFilter::saveToFile(ccHObject* entity, QString filename, SaveP
 	bool writeColors = cloud->hasColors();
 	bool writeNorms = cloud->hasNormals();
 	std::vector<ccScalarField*> theScalarFields;
+
 	if (cloud->isKindOf(CC_TYPES::POINT_CLOUD))
 	{
 		ccPointCloud* ccCloud = static_cast<ccPointCloud*>(cloud);
@@ -210,6 +211,11 @@ CC_FILE_ERROR AsciiFilter::saveToFile(ccHObject* entity, QString filename, SaveP
 	if (saveColumnsHeader)
 	{
 		QString header("//");
+		// We append the cloud name to the header
+		header.append(cloud->getName());
+		header.append("\n");
+
+		header.append("//");
 		header.append(AsciiHeaderColumns::X());
 		header.append(separator);
 		header.append(AsciiHeaderColumns::Y());
@@ -257,7 +263,7 @@ CC_FILE_ERROR AsciiFilter::saveToFile(ccHObject* entity, QString filename, SaveP
 			header.append(separator);
 			header.append(AsciiHeaderColumns::Nz());
 		}
-		
+
 		stream << header << "\n";
 	}
 


libs/qCC_io/AsciiFilter.cpp
```diff
@@ -210,6 +211,11 @@ CC_FILE_ERROR AsciiFilter::saveToFile(ccHObject* entity, QString filename, SaveP
 	if (saveColumnsHeader)
 	{
 		QString header("//");
+		// We append the cloud name to the header
```
asmaloney: This comment doesn't really add any information that isn't obvious in the following line...

Other than that 👍

dgirardeau:
I'm a bit annoyed by this modification "as is" as it may change the current ASCII output without any warning.

Could we add a boolean to control that? (just like 'saveColumnsHeader')?

ORNis:
Yes sir! 'will look at this ASAP!

asmaloney:
If the comments are somehow semantically significant, then maybe they either shouldn't be comments or they should add some structure to make it clear they are not normal comments?

ORNis:
There is no "normal comment", I mean in an ASCII file you can choose '//' as well as '#' to start a comment line, nothing is standard. So in all case they are intrusive and proprietary.

asmaloney:
Ok.  But mixing the two - comments and something that's semantically significant - using the same tokens is usually a bad idea.  (Disclaimer - I've never looked at one of these files, so I don't know what your current format looks like.)

If you are free to change things, then I would consider making one - say "//" a real comment that is always ignored when reading, and the other "#" a structured notation (or whatever you want to call it) that has some significance when reading the file.

ORNis:
My point is that what ever we choose it's potentially have the impact of a  "structured notation" because it could really change the way third party software parse CC files because it change the *header*, and usually people simply ignore it or read it. I do think it's why @dgirardeau want to make this optional.

dgirardeau:
There's no official format or anything. But this header 'comment' as been added a while ago because some people were requesting it for their own needs... (and now CC uses it as well by the way).

But if we add another line in a 'de facto' structured file, it may break their process. This is why, just as the previous 'comment' (which is not output by default), I'd like this new one to be also optional.

ORNis:
@dgirardeau I think we agree

asmaloney:
And my point is that if the file did have a simple structure you wouldn't have to worry about breaking other people's parsing of this file.  It would eliminate the need for adding more and more options to the interface when people (inevitably) want more things.  It would ultimately make this simpler.

This could be as simple as:
```
// This is a comment that is always ignored by parsers
```
and allow some key/value pairs
```
# CloudName: Foo
# Titles: X Y Z
# FormatVersion: 2
# NumPoints: 655321
# Key: Value
```

This would break current versions, but you could just name the new format v2.

ORNis:
Good point ;).
But for now, I think it's worth adding UI complexity than breaking backward compatibility or create a new format (there is already to much 3d formats)... I think it's up to @dgirardeau

dgirardeau:
To tell the truth I don't think it's a good idea to add this option at all ;).

asmaloney:
> there is already to much 3d formats

I agree, but a straight text file like this without any other info can be problematic too because once removed from the context (look at some random file created 5 years ago) all info about it is left as an exercise for the reader.

I believe that any proper data file format - and just calling it ASCII doesn't make it less of a format - has at least a header with pertinent info.

The other information I always find useful would fit there too:

```
# CreatedBy: CloudCompare
# CreatedByVersion: 2.7.0
# CreatedOn: 01-Jun-16 at 10:16 AM
```

CC can list what Keys it understands and uses, but anyone is free to put whatever they want in there without breaking everyone else.  Totally flexible.

Anyway - that's my suggestion based on my experience.  A small bit of pain now to avoid some future pain.

But I leave it to @dgirardeau a.k.a. El Presidente, El Heffe, Head Honcho.

dgirardeau:
El Commandante!

I like this idea. Instead of adding an 'add cloud name' option, we could add a single 'add meta data as comments' or similar option.

And we'll put everything we can think of there (those suggested by Andy, etc.) and it will also acceptable to add other things later (as we don't explicitly announce what we are adding ;).

ORNis:
Ok I close this

PR #412, opened 2016-07-17
fix compile error under ubuntu 

jianboqi:
fix the merge conflicts and compile error under ubuntu


asmaloney:
Again - PRs should only do one thing at a time.  This should be multiple PRs.

And you still need to remove the unnecessary comments (the commented out code and the `//zwm`).

PR #428, opened 2016-09-02
Small typo

whatnick:
Corrected GDAL spelling.
P.S. Osgeo4W distribution of GDAL contains a lot of the CC dependencies will be an great starting point for people building for the first time.


cloudcompare:
Thanks.

Regarding Osgeo4W:
- given its name I don't think that Osgeo4W is available on Linux and Mac, is it?
- and can't you already use the libraries from Osgeo4W packages to compile CC?

whatnick:
Yes Osgeo4W windows only. Given there is no easy linux/macports/homebrew
like package management in windows, osgeo4w provides a short-cut to getting
prebuilt versions of a lot of the dependencies and could be a recommended
path.

PR #432, opened 2016-09-07
Fix unconditional stereo buffer enabling.

### LINKED ISSUE #406: GLX startup problem
Hello, 

I tried to re-install on a fresh Arch Linux machine and I get  
`Could not initialize GLX`
`Aborted (core dumped)`

I can find any logs or any useful info. ccViewer runs fine.

renatoGarcia:
Fix #406

Besides unconditional that enabling was redundant, three lines down an enabling controlled by CC_GL_WINDOW_USE_QWINDOW was present.

With this patch the stereo buffer is enabled in compile time by OPTION_GL_QUAD_BUFFER_SUPPORT cmake option.


cloudcompare:
I guess you are right ;)

PR #441, opened 2016-09-29
Update qhull 2012.1=>2015.2 && qHPR/qhull Intel C/C++ Compiler compilation fix

nine7nine:
Hey CC,

I thought while I wait to hear back from the Qhull maintainers, you could have a look/review of my changes to CC/qHPR Plugin... My branch/fork updates Qhull to 2015.2 (the last release for libqhull, before upstream switched to libqhull_r)... I know there is the possibility that CC may switch to CGAL for qHPR, but updating qhull took no effort and seems to work fine with the qHPR plugin. 

I tried to keep the files exactly as the .../src2012.1 folder had them, removing any unnecessary files and updating the CMakelists.txt to point to ../src2015.2... The only additional file that I added was the file noting that this version is deprecated... I've been using qhull-2015.2 for over a week now and I haven't bumped into any problems, although it would still need to be reviewed...** As i cannot test building the new qhull on Windows or Mac, someone else would need to do that.

The other commit is my fix for the Intel C/C++ Compiler, which is what I am waiting hear back from qhull/qhull on github about. They haven't been quick to reply ~ I think it's been over a week now (?).  Both of my Archlinux source packages apply this commit/patch; gcc and icc builds compile fine with it applied. Unfortunately/again; I can't test on Windows or MacOSX builds, as I only run Linux. the icc fix shouldn't matter on Windows. * However, I have no idea on MacOSX (someone would need to test that... Although, i suspect it won't be a problem).

thanks!


cloudcompare:
Thanks

nine7nine:
Cool. no problem! :)

I'm glad to get the changes included!

PR #452, opened 2016-11-14
add option to set last shift and scale info

vinayan:
add the option to update last shift and scale when using import through a plugin that needs to suppress the shfit&scale dialog. Discussions here

http://www.danielgm.net/cc/forum/viewtopic.php?f=3&t=2081&sid=b9cee2e0adfe4758ec3ab6be540d522b

### DIFF
--- a/libs/qCC_io/ccShiftAndScaleCloudDlg.cpp
+++ b/libs/qCC_io/ccShiftAndScaleCloudDlg.cpp
@@ -457,3 +457,18 @@ int ccShiftAndScaleCloudDlg::addShiftInfo(const ShiftInfo& info)
 
 	return static_cast<int>(m_defaultInfos.size())-1;
 }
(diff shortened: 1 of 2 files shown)


libs/qCC_io/ccShiftAndScaleCloudDlg.cpp
```diff
@@ -457,3 +457,18 @@ int ccShiftAndScaleCloudDlg::addShiftInfo(const ShiftInfo& info)
 
 	return static_cast<int>(m_defaultInfos.size())-1;
 }
+
+void ccShiftAndScaleCloudDlg::SetLastInfo(CCVector3d shift,double scale)
+{
+    try
```
cloudcompare: The try / catch statement is useless here. I'll remove it.

PR #462, opened 2016-11-18
Cleanup

asmaloney:
A series of small changes:

- avoids extra call to `ccPointCloud::clearLOD()` in `ccPointCloud::unalloactePoints()`
- reorder member variables to avoid padding in `struct DisplayDesc`
- converts a couple of old-style casts to `static_cast`
- convert dynamic exception specification (deprecated) to `noexcept`
- prefer `nullptr` to "0" for pointer initialization
- reorder initialization list to match declarations

### DIFF
--- a/libs/qCC_db/ccPointCloud.cpp
+++ b/libs/qCC_db/ccPointCloud.cpp
@@ -2578,7 +2578,7 @@ void ccPointCloud::drawMeOnly(CC_DRAW_CONTEXT& context)
 					unsigned steps = m_currentDisplayedScalarField->getColorRampSteps();
 					assert(steps != 0);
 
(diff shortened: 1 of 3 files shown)


libs/qCC_db/ccPointCloud.cpp
```diff
@@ -431,12 +431,11 @@ void ccPointCloud::clear()
 
 void ccPointCloud::unalloactePoints()
 {
-	clearLOD();
```
cloudcompare: No we have to clear the LOD structure before clearing the colors / SFs. But we can't call 'notifyGeometryUpdate' before releasing them :D

asmaloney: OK - I wondered about that.  One of the reasons for PRs!

Will fix.

PR #480, opened 2016-12-06
{Mac} Adjust font size on macOS

asmaloney:
Fonts are too small on macOS (regular display, not retina), so add small adjustment to make them more readable


dgirardeau:
Wouldn't it be more logical to change the default font size in the display properties? (see 'ccGui::ParamStruct::reset()' in libs/qCC_glWindow/ccGuiParameters.cpp)

e.g. labelFontSize = 10 instead of 8

asmaloney:
Abandoned for other branch based on input.

PR #488, opened 2017-01-02
Strange behavior in sphere fitting 

Luisloez89:
The refinement step is not improving the results. Derivatives are not working as an accumulative element in "GeometricalAnalysisTools::refineSphereLS" so the deviation is being applied in the direction of the vector defined by the latest point of the point cloud and the center of the sphere.


dgirardeau:
Nice! Thank you very much for the fix.

PR #494, opened 2017-01-11
ccCompass

samthiele:
Added plugin for easy digitization of geological structures and structural traces on point clouds


luca-penasa:
super cool!

problem I spotted:
when in "measure orientation froms tructure path" it tries to access color informations also if they does not exists in the picked cloud. A check should be done for that... and can you use any scalar field for that? Is so maybe using intensity would be a good idea..

plugins/qCompass/ccCompassDlg.h
```diff
@@ -0,0 +1,65 @@
...
+	explicit ccCompassDlg(QWidget* parent = 0);
+
+	/*
+	Returns a flag describing the currently selected ccTrace::COST_MODE (used to build the cost function for optimisation)
+	*/
+	int getCostMode();
+
+	bool planeFitMode();
+
+	QMenu *m_cost_algorithm_menu;
```
asmaloney: Do all these member vars need to be public?

plugins/qCompass/ccCompassInfo.h
```diff
@@ -0,0 +1,94 @@
...
+		QDialogButtonBox *buttonBox = new QDialogButtonBox(Qt::Horizontal);
+		buttonBox->addButton(okButton, QDialogButtonBox::AcceptRole);
+		connect(buttonBox, SIGNAL(accepted()), this, SLOT(accept())); //done
+		QVBoxLayout *lt = new QVBoxLayout;
+		lt->addWidget(l);
+		lt->addWidget(buttonBox);
+		setLayout(lt);
+
+		//add text
+		l->setText("<!DOCTYPE html>"
```
asmaloney: Maybe better to store HTML in an HTML file (so it can be easily edited/checked for errors) and read in for display?

plugins/qCompass/ccLineation.h
```diff
@@ -0,0 +1,165 @@
...
+
+	virtual ~ccLineation() {}
+
+	//inherited from ccHObject
+	inline virtual CC_CLASS_ENUM getClassID() const override { return CC_TYPES::POLY_LINE; }
+
+	CCVector3 getDirection()
+	{
+		if (size() != 2)
+			return CCVector3();
```
asmaloney: Add {}

plugins/qCompass/ccCompass.h
```diff
@@ -0,0 +1,167 @@
...
+
+	//link to application windows
+	ccGLWindow* m_window;
+	QMainWindow* m_main_window;
+
+	//2D-circle for selection during plane-mode
+	ccMouseCircle* m_mouseCircle;
+
+	//ccCompass toolbar gui
+    ccCompassDlg* m_dlg;
```
asmaloney: Formatting

plugins/qCompass/ccCompassDlg.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+	algorithmButton->setPopupMode(QToolButton::InstantPopup);
+	//N.B. the menu is added to the algorithm button in the functions that are called with "mode" (plane, trace, lineation) changes - see ccCompass
+
+	//set background color
+	QPalette p;
+	p.setColor(backgroundRole(), QColor(240, 240, 240, 200));
+	setPalette(p);
+	setAutoFillBackground(true);
+
```
asmaloney: Remove blank line

plugins/qCompass/ccCompassDlg.cpp
```diff
@@ -0,0 +1,103 @@
...
+	if (m_grad->isChecked())
+		out = out | ccTrace::MODE::GRADIENT;
+	if (m_dist->isChecked())
+		out = out | ccTrace::MODE::DISTANCE;
+	if (m_scalar->isChecked())
+		out = out | ccTrace::MODE::SCALAR;
+	if (m_scalar_inv->isChecked())
+		out = out | ccTrace::MODE::INV_SCALAR;
+
+	if (out == 0) return ccTrace::MODE::DISTANCE; //default to distance if everything has been unchecked
```
asmaloney: Two lines

plugins/qCompass/ccCompassDlg.h
```diff
@@ -0,0 +1,65 @@
...
+#include <ui_compassDlg.h>
+#include "ccTrace.h"
+
+
+class ccCompassDlg : public ccOverlayDialog, public Ui::compassDlg
+{
+	Q_OBJECT
+
+public:
+
```
asmaloney: Remove blank line

plugins/qCompass/ccCompassInfo.h
```diff
@@ -0,0 +1,94 @@
...
+#include <qtextedit.h>
+
+class ccCompassInfo : public QDialog
+{
+	
+public:
+	explicit ccCompassInfo(QWidget *parent = 0) :
+		QDialog(parent)
+	{
+
```
asmaloney: Blank line

I would also put this into a cpp file.

plugins/qCompass/ccLineation.h
```diff
@@ -0,0 +1,165 @@
...
+static QSharedPointer<ccSphere> c_unitPointMarker(0);
+static QSharedPointer<ccCylinder> c_bodyMarker(0);
+static QSharedPointer<ccCone> c_headMarker(0);
+
+class ccLineation : public ccPolyline
+{
+public:
+	ccLineation(ccPointCloud* associatedCloud)
+		: ccPolyline(associatedCloud)
+	{ /*ctor */ }
```
asmaloney: No need for a comment - doesn't add anything.

plugins/qCompass/ccMouseCircle.cpp
```diff
@@ -0,0 +1,155 @@
+#include "ccMouseCircle.h"
+
+#include <math.h>
+
+ccMouseCircle::ccMouseCircle(ccGLWindow* owner, QString name) : cc2DViewportObject(name.isEmpty() ? "label" : name)
+{
+	this->setVisible(true);
+	this->setEnabled(false);
```
asmaloney: Why specify "this"?

plugins/qCompass/ccMouseCircle.cpp
```diff
@@ -0,0 +1,155 @@
...
+	glFunc->glLineWidth(2);
+	glFunc->glLineStipple(1, 0xAAAA);
+	glFunc->glEnable(GL_LINE_STIPPLE);
+
+	const unsigned char* defaultColor = m_selected ? ccColor::red.rgba : context.textDefaultCol.rgb;
+	glFunc->glColor3ubv(ccColor::red.rgba);
+
+	//get height & width
+	int halfW = context.glW / 2.0f;
+	int halfH = context.glH / 2.0f;
```
asmaloney: Assigning to "int" but using floats. This will cause warnings on clang.

plugins/qCompass/ccMouseCircle.h
```diff
@@ -0,0 +1,52 @@
...
+
+	//deconstructor
+	~ccMouseCircle();
+
+	//get the circle radius in px
+	int getRadiusPx();
+
+	//get the circle radius in world coordinates
+	float getRadiusWorld();
+
```
asmaloney: Extra line

plugins/qCompass/ccTrace.h
```diff
@@ -0,0 +1,192 @@
...
+	Applies the optimized path to the underlying polyline object (allows saving etc.).
+	*/
+	void finalizePath()
+	{
+		//clear existing points in background "polyline"
+		this->clear();
+
+		//push trace buffer to said polyline (for save/export etc.)
+		for (std::deque<int> seg : m_trace)
+			for (int p : seg )
```
asmaloney: Add {}

plugins/qCompass/ccTrace.h
```diff
@@ -0,0 +1,192 @@
...
+	*/
+	bool optimizePath(int maxIterations = 1000000);
+
+	/*
+	Applies the optimized path to the underlying polyline object (allows saving etc.).
+	*/
+	void finalizePath()
+	{
+		//clear existing points in background "polyline"
+		this->clear();
```
asmaloney: Why "this"?

plugins/qCompass/ccTrace.h
```diff
@@ -0,0 +1,192 @@
...
+	Fit a plane to this trace. Returns true if a plane was succesfully fitted.
+
+	@Args
+	*surface_effect_tolerance* = the minimum allowable angle (in degrees) between the average surface normals of the points [if they exist] and the plane. This gets rid of fit planes that simply represent the outcrop surface.
+	*min_planarity* = the minimum allowable planarity (where a perfect plane = 1, while a line or random point cloud = 0)
+
+	@Return
+	The plane that was fitted, or a null pointer (0) if the plane is not considered valid (by the above criterion)
+	*/
+	ccPlane* fitPlane(int surface_effect_tolerance = 10, float min_planarity = 0.75);
```
asmaloney: 0.75 -> 0.75f

plugins/qCompass/ccTrace.h
```diff
@@ -0,0 +1,192 @@
...
+	ccPointCloud* m_cloud=0; //pointer to ccPointCloud object this is linked to (slightly different to polylines as we know this data is sampled from a real cloud)
+	
+	std::vector<std::deque<int>> m_trace; //contains an ordered list of indices which define this trace. Note that indices representing nodes MAY be inserted twice.
+	std::vector<int> m_waypoints; //list of waypoint indices
+	int m_previous=-1; //for undoing waypoints
+
+	ccColor::Rgba m_waypoint_colour = ccColor::green;
+	ccColor::Rgba m_trace_colour = ccColor::yellow;
+
+
```
asmaloney: Extra line

plugins/qCompass/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
...
+    #set( CC_OPENGL_FILTER ON BOOL)
+
+    project( CC_COMPASS_PLUGIN )
+
+	#As we are going to re-use several components of the main qCC project
+	#we define its sources path relatively to the plugin's dir
+	set(CloudCompare_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/../../qCC)
+	
+	#include the whole CC directory (easier)
+    include_directories( ${CloudCompare_SOURCE_DIR} )
```
asmaloney: Formatting

plugins/qCompass/ccTrace.cpp
```diff
@@ -0,0 +1,678 @@
...
+}
+
+int ccTrace::COST_MODE = ccTrace::MODE::DARK; //set default cost mode
+std::deque<int> ccTrace::optimizeSegment(int start, int end, float search_r, int maxIterations, int offset)
+{
+	//check handle to point cloud
+	if (!m_cloud) return std::deque<int>(); //error -> no cloud
+
+	//retreive and store start & end rgb
+	const ColorCompType* s = m_cloud->getPointColor(start);
```
luca-penasa: here you must check if the cloud has colors, if not getPointColor causes a segfault

plugins/qCompass/ccCompassInfo.h
samthiele: I agree completely - though I'm unsure how to do this such that the file gets included in the compiled binary (avoiding relative path issues etc. on distribution)

samthiele:
Thanks for the feedback @asmaloney and @luca-penasa. I have fixed the bug regarding point clouds without colour and the formatting issues.

As for  @luca-penasa's query about using scalar fields to define the cost function - yes, this is possible: select the "Scalar Field" or more normally the "Inverse Scalar Field" cost function in the algorithm dropdown.

plugins/qCompass/ccCompassInfo.h
asmaloney: You should be able to use the [Qt resource system](https://doc.qt.io/qt-5/resources.html). Essentially you add the file to a .qrc file, then to reference the HTML file you use the resource path (":foo/bar.html"). This avoids the distribution issue since it will be included in the compiled binary, but kept as HTML for editing.

(Note from the docs that you may need to use `Q_INIT_RESOURCE() `. I'm not sure quite how resources work with dynamic libs.)

If you need any help, let me know!

plugins/qCompass/ccCompass.cpp
```diff
@@ -0,0 +1,964 @@
...
+//#                                                                        #
+//##########################################################################
+
+#include "ccCompass.h"
+
+
+//Qt
+#include <QtGui>
+
+ccCompass::ccCompass(QObject* parent/*=0*/)
```
luca-penasa: I think you should initialize m_mouseCircle as a nullptr in constructor, or the destructor, when called, will try to delete it causing an invisible segafault on cc exit... at least gdb is telling me this

luca-penasa: I mean, this happens if you never call the qCompass plugin during the cc execution. the pointer remains not initialized and if(ptr) will return true...

plugins/qCompass/ccCompass.qrc
```diff
@@ -0,0 +1,5 @@
+<RCC>
+  <qresource prefix="/CC/plugin/qDummyPlugin" >
```
asmaloney: Might want to change the prefix?

plugins/qCompass/ccCompassDlg.cpp
```diff
@@ -0,0 +1,103 @@
...
+	, Ui::compassDlg()
+{
+	setupUi(this);
+
+	//setup combo box
+	categoryBox->clear();
+	categoryBox->addItems(QStringList() << "S0 (Bedding)" << "S1" << "S2" << "L1" << "L2" << "Custom..."); //default items
+
+	//setup "algorithm" dropdown
+	m_cost_algorithm_menu = new QMenu();
```
asmaloney: Shouldn't this have a parent? `new QMenu(this)`

plugins/qCompass/ccLineation.h
```diff
@@ -0,0 +1,164 @@
...
+#include <ccSphere.h>
+#include <ccCylinder.h>
+#include <ccCone.h>
+#include <GenericIndexedCloudPersist.h>
+#include <ccPointCloud.h>
+
+//static sphere for drawing withs
+static QSharedPointer<ccSphere> c_unitPointMarker(0);
+static QSharedPointer<ccCylinder> c_bodyMarker(0);
+static QSharedPointer<ccCone> c_headMarker(0);
```
asmaloney: Pointers should use `nullptr`, not 0.

plugins/qCompass/ccLineation.h
```diff
@@ -0,0 +1,164 @@
...
+			const CCVector3 start = *getPoint(0);
+			const CCVector3 end = *getPoint(1);
+			return end - start;
+		}
+	}
+
+	ccColor::Rgba colour = ccColor::green;
+
+protected:
+	float m_relMarkerScale = 5.0;
```
asmaloney: 5.0 -> 5.0f

plugins/qCompass/ccLineation.h
```diff
@@ -0,0 +1,164 @@
...
+
+	//overidden from ccHObject
+	virtual void drawMeOnly(CC_DRAW_CONTEXT& context) override
```
asmaloney: This method should be moved to a cpp file.

plugins/qCompass/ccMouseCircle.cpp
```diff
@@ -0,0 +1,155 @@
+#include "ccMouseCircle.h"
+
+#include <math.h>
+
+ccMouseCircle::ccMouseCircle(ccGLWindow* owner, QString name) : cc2DViewportObject(name.isEmpty() ? "label" : name)
```
asmaloney: Initializer on next line.

plugins/qCompass/ccMouseCircle.cpp
```diff
@@ -0,0 +1,155 @@
...
+{
+	setVisible(true);
+	setEnabled(false);
+
+	//setup unit circle
+	float heading;
```
asmaloney: Move declaration into loop since you aren't using it outside loop.
(`float heading = n ...`)

plugins/qCompass/ccMouseCircle.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+	if (glFunc == nullptr)
+		return;
+
+	//test viewport parameters
+	const ccViewportParameters& params = context.display->getViewportParameters();
+	glFunc->glPushAttrib(GL_LINE_BIT);
+
+	float relativeZoom = 1.0f;
+	float dx = 0, dy = 0;
```
asmaloney: 0.0f and declarations on separate lines.

plugins/qCompass/ccMouseCircle.cpp
```diff
@@ -0,0 +1,155 @@
...
+	glFunc->glLineWidth(2);
+	glFunc->glLineStipple(1, 0xAAAA);
+	glFunc->glEnable(GL_LINE_STIPPLE);
+
+	const unsigned char* defaultColor = m_selected ? ccColor::red.rgba : context.textDefaultCol.rgb;
+	glFunc->glColor3ubv(ccColor::red.rgba);
+
+	//get height & width
+	int halfW = (int) (context.glW / 2.0f);
+	int halfH = (int) (context.glH / 2.0f);
```
asmaloney: Don't use C-style casts - use `static_cast`

plugins/qCompass/ccMouseCircle.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+		//is control down
+		if (wheelEvent->modifiers().testFlag(Qt::ControlModifier))
+		{
+			//adjust radius
+			ccMouseCircle::RADIUS -= ccMouseCircle::RADIUS_STEP*(wheelEvent->delta()/100.0);
+
+			//avoid really small radius
+			if (ccMouseCircle::RADIUS < ccMouseCircle::RADIUS_STEP)
+				ccMouseCircle::RADIUS = ccMouseCircle::RADIUS_STEP;
```
asmaloney: {}

plugins/qCompass/ccTrace.h
```diff
@@ -0,0 +1,199 @@
...
+	//inherited from ccHObject
+	inline virtual CC_CLASS_ENUM getClassID() const override { return CC_TYPES::POLY_LINE; }
+
+	/*
+	Adds waypoint to the end of this trace.
+	@Args
+	 *pointIndex* = the index of the point (in the cloud associated with this object) that is the new waypoint
+	*/
+	void pushWaypoint(int pointId) { m_waypoints.push_back(pointId); }//addPointIndex(id);}
```
asmaloney: Remove (debug?) comment.

plugins/qCompass/ccTrace.h
```diff
@@ -0,0 +1,199 @@
...
+	*/
+	int insertWaypoint(int pointId);
+
+	/*
+	Undo last action
+	*/
+	void undoLast()
+	{
+		if (m_previous != -1)
+			m_waypoints.erase(m_waypoints.begin() + m_previous);
```
asmaloney: This is exactly the reason to always use {}. The code says only this line is executed for the `if`, but the indentation seems to imply that the `m_trace` line should be too...

plugins/qCompass/info.html
```diff
@@ -0,0 +1,66 @@
...
+<h2>Exporting data</h2>
+<p> Interpretation generated using this tool is represented internally by Cloud Compare as planes and polygons, and so can be natively stored in CC's .bin files. 
+For ease of use, a <i>save</i> tool has also been included. This tool exports plane and trace data to .csv files, from which data can be easily imported 
+into other software packages for further analysis. Polylines created by the <i>Trace</i> and <i>Lineation</i> tools can also be exported as shapefiles through the normal 
+
+<h2>Further Reading and Citation</h2>
+<p>If you found this tool useful, please cite <i>Thiele et al., 2017</i>. The publication also contains a more detailed 
+description of the methods employed by this plugin.</p>
+<p>
+<i>Thiele, ST., Grose, L., Samsu, A., Micklethwaite, S. & Cruden, S., 2017, 'ccCompass: A computer-assisted approach to structural interpretation of point clouds', Journal TBA </i>
```
asmaloney: "&" -> "\&amp;"

plugins/qCompass/ccCompassInfo.h
```diff
@@ -7,88 +7,14 @@
 #include <qdialogbuttonbox.h>
 #include <qboxlayout.h>
 #include <qtextedit.h>
+#include <qfile.h>
```
asmaloney: These should all use the same &lt;QFile&gt; &lt;QTextEdit&gt; style like you do with QTextStream, not the .h ones

plugins/qCompass/ccCompassInfo.cpp
```diff
@@ -0,0 +1,37 @@
...
+	QDialogButtonBox *buttonBox = new QDialogButtonBox(Qt::Horizontal);
+	buttonBox->addButton(okButton, QDialogButtonBox::AcceptRole);
+	connect(buttonBox, SIGNAL(accepted()), this, SLOT(accept())); //done
+	QVBoxLayout *lt = new QVBoxLayout;
+	lt->addWidget(l);
+	lt->addWidget(buttonBox);
+	setLayout(lt);
+
+	//load text
+	//QIcon(":/CC/plugin/qCompass/icon.png");
```
asmaloney: Remove QIcon comment

samthiele:
Ok cool - thanks for your help Andy :)

dgirardeau:
Okay, nice job Sam! Thanks for the contribution.

I think we will make a new (partial) release soon (version 2.8.1). It will be a great addition.

samthiele:
Awesome - hopefully it's useful!

dgirardeau:
Oops, I missed the 'CC_COMPASS_PLUGIN' name of the project :D

The 'CC_' prefix is reserved for core components. Plugins start with 'QXXX'. I'll rename the project QCOMPASS_PLUGIN to match the default plugin format.

dgirardeau:
I also had to add a 'return' statement at the end of _optimizeSegment_. I guess it's not possible to reach this end of the function?

And I still have 3 warnings at lines 558, 559 and 560 of ccCompass.cpp regarding the "\." string. Do you actually want to detect the point character or the anti-slash and the point character?

samthiele:
You're correct - it shouldn't be possible to reach the end of _optimizeSegment_, though one never knows.... Also I just noticed that _optimizeSegment_ has an unused argument _offset_. This can be removed (along with the associated comment) as it's no longer used.

As for the warnings - I want the string to be \. as it's a regex to find the last "." in the path (i.e. I'm appending text just before the file extension). Should the strings be replaced with "\\." maybe? (Visual Studio seems to be funny with escape characters).

dgirardeau:
That's noted for _optimizeSegment_.

And for the calls to lastIndexOf, we could simply use the '.' (character) instead of a "." or "\." string.

samthiele:
Sure - if that works. I initially used ".", but it was interpreted as regex and hence matched every character. If it matches properly as a char then great!

(to test it, use the tool to make some planes or traces or lineations and then try exporting them using the _save_ button on the plugins toolbar - if it doesn't work then the filenames will come out all funny).

dgirardeau:
Ok I'll do this.

dgirardeau:
Ah! Because I didn't knew what to set, I input a filename with no extension... and I got strange errors in the Console ;)

To make something more robust, I used QFileInfo. And then I modified a lot of things (hopefully for the better). Mostly to avoid crashes when the window or CC is closed before the plugin. I'll let you test that I haven't broken everything ;)

samthiele:
Strange - on my computer it was automatically setting the filename extension to .csv if none was specified. Regardless - it still seems to work fine.

One small thing - a comma should be written to the output file at the end of line 729 of ccCompass.cpp (this was my mistake...)

`*out << object->getMetaData("Strike").toString() << "," << object->getMetaData("Dip").toString() << "," << object->getMetaData("Strike").toFloat() + 90;`

should be changed to

`*out << object->getMetaData("Strike").toString() << "," << object->getMetaData("Dip").toString() << "," << object->getMetaData("Strike").toFloat() + 90 << ",";`

dgirardeau:
Okay that's fixed ;)

PR #506, opened 2017-02-11
Add snap for CloudCompare

mardy:
This change adds the files needed to build CloudCompare as a [snap package](https://snapcraft.io). The package can be generated just by typing `snapcraft` from the project base directory on an Ubuntu machine (I recommend using 16.04), and can be run on any Linux distribution which supports the snap formats (most of them).

Users can then install the package just by typing
```
    snap install cloudcompare
```

I took the liberty of registering the "cloudcompare" name in the snap store, and uploaded version 2.8.0 there. Please let me know which maintainers would like to be able to push packages to the store (I need their e-mail address) so that I'll add them as uploaders; I recommend at least one to be added, in case I'm unavailable.

Please also let me know if this effort is welcome or not: in case it isn't, it's easy for me to publish cloudcompare as "cloudcompare-mardy", so that your project won't be directly associated with the snap package.


rjanvier:
Well, it's a good initiative.
But IMHO it would have been nicer to consult the team before taking a such iniative as there are already some Linux maintainers and packagers. For example, last time I look at snap, I felt that it was promising but not mature enough for the way we want to distribute CC + plugins, so it was decided that .deb should  still be the way to do it  for 2.8. The good news is that I failed to deliver a 2.8.0 deb and then your snap package can help us to provide quickly something to users. The bad news is that this snap package doesn't solve the things that made the 2.8.0 deb delayed: because I'm trying to handle as much plugin as I can, for as much Ubuntu/Deb. version as I can + continous delivery of beta versions and license issues (Ransac_SD). So, the question, before we consider this pull request, is maybe to know if you want to be a full time maintainer or if it is just a "quick and dirty" thing to make some kind of "domain squatting". If you want work with us to improve the way CC is bundled for Linux I guess we would be glad to accept this PR. if  It's just a "fire and forget" thing, maybe the "cloudcompare-mardy" solution you suggested is the best.

mardy:
Hi Romain,
  I understand your concerns, they are indeed valid. There is absolutely
no intention from my side to do a name squatting; besides, the snap
store provides an easy way for people to open a case about name
registrations, so you don't have to worry about that (nor in this case,
nor in other cases which might happen in the future).

Now, let me explain why I decided to contribute this, and how I intend
to follow up; then you'll take your choice.

I've been trying some SfM software; most of them have very complicated
build processes and rarely provide Linux binaries, and for that reason I
decided that since I had to build them anyway, I would make snaps for
them so that other people could try them out more easily. Then I thought
that to complete the loop I could also make snaps of a couple of mesh
visualizers, and I chose CloudCompare and MeshLab.

Indeed, between all the programs, CloudCompare is the one that benefits
less from a snap package, given that your build process is extremely
easy and smooth. But still, not all Linux users are comfortable with the
command line, so I decided to make a snap for it as well.

Now, as you might have already guessed, I don't have the time to
actively follow and participate in all these projects. My mailbox would
simply explode. What I can certainly take care of, is building and
uploading a new snap whenever a new release is made, but I need to be
"pinged" when a new release happens.
A possibility worth exploring is to have snaps being automatically
built; I know for sure that launchpad.net allows that, and it might be
that this is also possible to have github's travis CI do that. If you
are interested, I can investigate. In that case, you could count on me
as being responsible for the timely release of CC's snaps.

As for this merge proposal, as you can see it's built with the default
cmake parameters, meaning that all plugins are disabled. This is enough
for what I need, but you might not like it. If you accept the proposal
that I take care of snapping CC whenever you make a new release, then
please let me know what plugins you consider as essential and I'll
include them in the build.
Otherwise, I'll release the "cloudcompare" name and use
"cloudcompare-mardy".

rjanvier:
Well, thank you. that seems acceptable to me if you take care of building CC with the plugins. You can look in the control file the rules file of the debian package here https://github.com/CloudCompare/CloudCompare/tree/Debian/debian. you'll find what you need.  You'll find also my email adress if you whant to add me as an uploader.

mardy:
Hi Romain, I pushed a new commit where I enable the same plugins as in Debian. It seems to work, but since I'm not familiar with CC, it would be good if you or some other CC expert could give a try at the snap package and let me know if you find any issues. I've uploaded it to the "candidate" channel, which means that in order to install it you need to type
```
snap install cloudcompare --candidate
```
Once you confirm that everything is fine, I'll move that to the "stable" channel (so one won't need the `--candidate` flag to install it). Currently the CC in the stable channel is the one I uploaded without plugins.

I've sent you an invite for the snap store, please let me know if you have any trouble accessing it.

snap/snapcraft.yaml
```diff
@@ -25,14 +25,68 @@ apps:
...
+        -DINSTALL_QPCV_PLUGIN=ON \
+        -DINSTALL_QM3C2_PLUGIN=ON \
+        -DINSTALL_QBROOM_PLUGIN=true \
+        -DINSTALL_QHOUGH_NORMALS_PLUGIN=true \
+        -DEIGEN_ROOT_DIR=$SNAPCRAFT_STAGE/include/eigen3 \
+        -DLIBLAS_INCLUDE_DIR="/usr/include" \
+        -DLIBLAS_RELEASE_LIBRARY_FILE="/usr/lib/liblas.so" \
+        -DCMAKE_INSTALL_PREFIX=$SNAPCRAFT_PART_INSTALL \
+        ..
+      make -j8
```
rjanvier: Could we have a less hardcoded value here? Something like 
export NPROCS=$(grep -c ^processor /proc/cpuinfo)
make -j${NPROCS} should work.

mardy: OK, I'll try that.

mardy:
About libvxl: snaps are meant to be usable in any Linux distribution, not just Ubuntu. For this reason, they contain all the dependencies: the CC snap can be installed on machines without Qt, for example. There are ways to reduce the binary size by stripping common dependencies; for instance, there is a snap providing Qt and we could link to that, but this has the downside of bringing trouble if the Qt versions don't match. For this reason I think it's wiser to have all the deps in the snap; this ensure that when one reports bug, we know exactly what libraries were used in the build.

rjanvier:
Ok, that's the point I didn't get about snap I though that deps from repos would not get embeded in the archive. So if it's mature enough, it's definitly a game changer for us.

rjanvier:
Ok, I have tested  the snap pacakge on an antergos virtual machine by curiosity, it seems to work perfectly.

asmaloney:
Thanks for the contibution @mardy!

dgirardeau:
Thanks guys!

I've updated the 'download' page (http://www.cloudcompare.org/release) to advertise this new 'snap' packaging system.

dgirardeau:
Hi guys,

Do you know how to fix this issue: http://www.cloudcompare.org/forum/viewtopic.php?f=12&t=2376?

mardy:
On 20/03/2017 11:45, Daniel Girardeau-Montaut wrote:
> Hi guys,
> 
> Do you know how to fix this issue:
> http://www.cloudcompare.org/forum/viewtopic.php?f=12&t=2376?

Indeed, the current version of the snap is not asking for the
"removable-media" interface. I this it's actually a good idea to add it.
It should be just a matter of adding it to the list of plus in the
snapcraft.yaml file; please let me know if someone else is going to take
care of that, or if you'd like me to do it.

Then Daniel, I wonder if you could be so kind and convey this
information to the user who asked this (I'm not registered to the
forum): yes, the immediate solution is to copy the files into one's home.
Another option is to remove the snap, and reinstall it using the
"--devmode" option, which effectively disables confinement and will
allow direct access to any part of the filesystem.

rjanvier:
Ok I thanks, I saw that, I'll take care of this.

rjanvier:
@mardy ok, it seems to solve this issue. But the main issue I saw now, is that the removable-media plug is not an "auto-connect" one. It requires an additional manipulation for the end-user (sudo snap connect  cloudcompare:removable-media  cloudcompare:removable-media) and thus it's not very convenient. Is there any other alternative I should be aware of?

mardy:
Yes, the interface needs to be manually connected (but you don't need to use "sudo" if you run "sudo snap login" at least once). There are talks about making the connections more interactive (and with a GUI), but so far there aren't concrete plans, AFAIK.

rjanvier:
Ok, thanks. I guess I should write a proper documentation for the snap package, else this kind of issue will come back very frequently...

PR #507, opened 2017-02-13
ccTrace optimisations

samthiele:
Optimised shortest path estimation done using the trace tool (via the ccTrace class). Significantly improves picking speed for longer traces. Fixed minor bug involving point clouds where no valid path between start & end exists.


dgirardeau:
Thanks Sam ;)

PR #510, opened 2017-02-16
qSort is deprecated and will be removed in the future

asmaloney:
Replace by std::sort


rjanvier:
@asmaloney, @dgirardeau  maybe we could use this https://github.com/CloudCompare/CloudCompare/blob/c009de9d4aa399e695ce5fc333df30701a621744/CC/include/SortAlgo.h in these cases.

dgirardeau:
@rjanvier I agree but I'm not sure it's necessary (there's rarely more than 6 or 8 steps in a color scale, and anyway this method is rarely called)

PR #538, opened 2017-07-07
QString::arg doesn't have "%0" - it starts from "%1"

asmaloney:
(And one error text copy-paste fix: "FBX" -> "ICM")


asmaloney:
BTW - I just ran some test and it looks like these are equivalent:

```
   qDebug() << QString( "%0 %1" ).arg( "foo", "bar" );
   qDebug() << QString( "%1 %2" ).arg( "foo", "bar" );
```

So Qt is handling the case internally despite what the docs say: `Place marker numbers must be in the range 1 to 99.`.

Probably worth fixing anyways.

PR #541, opened 2017-07-08
Qt usage fixes based on clazy analysis

asmaloney:
- String::arg usage
- missing Q_OBJECT macros
- using references & consts to avoid copies

### DIFF
--- a/libs/qCC_db/ccPointCloudLOD.cpp
+++ b/libs/qCC_db/ccPointCloudLOD.cpp
@@ -27,6 +27,8 @@
 //! Thread for background computation
 class ccPointCloudLODThread : public QThread
 {
+	Q_OBJECT
+	
 public:
 	
 	//! Default constructor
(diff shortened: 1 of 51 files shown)


libs/qCC_db/ccPointCloudLOD.cpp
```diff
@@ -27,6 +27,8 @@
 //! Thread for background computation
 class ccPointCloudLODThread : public QThread
 {
+	Q_OBJECT
```
dgirardeau: Do you really need this when there's no signal/slot defined? I avoid putting it on purpose so as to avoid the creation and include of the MOC file.

asmaloney: There are several other reasons listed [here](https://github.com/KDE/clazy/blob/master/src/checks/level1/README-missing-qobject-macro.md).

If it's not included, then using qobject_cast won't work for example.

My view is that it's cleaner to always include it. That way when you do need to use one of those other features you won't be spending time trying to figure out why it doesn't work.

PR #560, opened 2017-07-24
Use system locale to format number of points in several places

### LINKED ISSUE #559: Add digit grouping in open file dialogue
When CloudCompare opens a point cloud file it displays the number of points that are loaded.
A digit grouping would be nice to make number of points readable more easily.

This is not a bug, more a feature request.

Attached a screenshot of the dialogue.

Current behaviour: 
Points: 41272065

Desired behavoiur:
Points: 41.272.065 or
Points: 41 272 065 ...

cloudcompare_screenshot_open_file

asmaloney:
I seem to recall we went back-and-forth on some QLocale problems and I see that `QLocale(QLocale::English)` is used in several paces.

I've used `QLocale::system()` - if it's a problem we can switch this to `QLocale::English`, but it would be better if we could handle it properly for everyone instead of forcing English.

Fixes #559

### DIFF
--- a/libs/qCC_db/ccGBLSensor.cpp
+++ b/libs/qCC_db/ccGBLSensor.cpp
@@ -682,7 +682,7 @@ bool ccGBLSensor::computeDepthBuffer(CCLib::GenericCloud* theCloud, int& errorCo
 			ccProgressDialog pdlg(true);
 			CCLib::NormalizedProgress nprogress(&pdlg,pointCount);
 			pdlg.setMethodTitle(QObject::tr("Depth buffer"));
-			pdlg.setInfo(QObject::tr("Points: %1").arg(pointCount));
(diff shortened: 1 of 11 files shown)


dgirardeau:
Actually, when we use this 'QLocale::English' thing is because we want the 'XXX XXX XXX' format for big numbers (the thing we don't have in the French locale sadly).

But at least adding the 'QLocale::system' version where it was not used won't harm ;)

asmaloney:
If I output `QLocale()` and `QLocale::system()` it tells me they are the same - `QLocale(English, Latin, UnitedStates)` - but if I use `QLocale().toString()` it doesn't add the separators. The docs say it should work, but it doesn't. 

So from what you said, using `QLocale::system()` has no effect when the language is set to French?

At least with this commit, some people will get separators :-)

dgirardeau:
But can't we use the English QLocale everywhere then? I want the separators as well :D

PR #565, opened 2017-08-01
Fix GenericChunkedArray::reserve() under 64-bit environments.

CecilHarvey:
According to comments: "If the new number of elements is smaller than the actual one,
nothing happens." However currently under 64-bit environments reserve() will actually shrink
the array, which causes loss of data.

### DIFF
--- a/CC/include/GenericChunkedArray.h
+++ b/CC/include/GenericChunkedArray.h
@@ -231,15 +231,17 @@ public:
 #ifdef CC_ENV_64
 		try
 		{
-			m_data.resize(capacity * N);
+			if (m_capacity < capacity)
+			{
+				m_data.resize(capacity * N);
+				m_capacity = capacity;
+			}
 		}
 		catch (const std::bad_alloc&)
 		{
 			//not enough memory
 			return false;
 		}
-
-		m_capacity = capacity;
 #else
 		while (m_capacity < capacity)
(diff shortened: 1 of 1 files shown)


dgirardeau:
Have you encountered an actual bug due to this or was the comment simply misleading?

CecilHarvey:
The actual bug was in our own customized version for our specific purpose (I'm unable to give more details on this, sorry).

The bug happened when we are transiting from 32-bit i386 to x86-64.

The 32-bit code does NOT have this problem, so it's very likely that the comment is not "misleading".

dgirardeau:
Okay. Actually I'm quite sure that this "feature" is not used on our side (not shrinking the container when reserving a smaller size). At the very beginning it was just a way to mimic the STL 'vector' container.

Modifying it now may induce some loss of memory. I'll have to double check the code everywhere.

CecilHarvey:
However this does result in various pitfalls, e.g.,
ccBox::setDimensions() will result in loss of all of data in m_triMtlIndexes
(at setDimensions()->buildUp()->multiple calls to operator+=()'s ->ccMesh::reserve() with size from 1 to n).

dgirardeau:
Okay, I took some time to check the code, and I believe it's safe to merge this modification.

Thanks for the contribution.

PR #566, opened 2017-08-01
Use more precise 'log1p(x)' instead of 'log(1 + x)'

asmaloney:
@dgirardeau Not sure what your Windows compiler will do, but it should work since it's part of the standard!


dgirardeau:
log1p seems to be defined on my side as well ;)

PR #570, opened 2017-08-06
Replace libLAS with PDAL

tmontaigu:
Replace the use of libLAS with PDAL for the LasFilter.

PDAL is able to read Las files version 1.0 to 1.4 (libLAS could only read 1.0, 1.1 , 1.2).
The LAS standard version 1.4 defines 5 new point format : 6,7,8,9,10.

- PDAL is not able to read LAS files with point format that have waveform data, wich means that 
point format 4, 5, 9 and 10 are **not** supported.

- One of the notable change  is that in new point formats ( > 6) the classification value takes the full 8 bits and not the 4 first bits only.

There is still something to change in the CloudCompare UI, when opening a LasFile, the CC asks if he wants the user to read the 8bits as the classification value or if he wants to decompose. However that "decomposing" stage is already done by PDAL.

The reading and writing make use of the StreamingAPI  that PDAL offers, so the behaviour is the same than when libLAS was used.
However the tiling process does not use the StreamingAPI (as it didn't seem possible), so instead the process is done in 3 steps:

1. Read the whole file
2. Tile points
 3. Write resulting files


dgirardeau:
That's good!

My only concern is about the 'Tiling' function, as it's only interest is too avoid loading all the points in memory (as it's meant to cut files that are too big to fit into memory. Why can't it be used as before exactly?

contrib/AllSupport.cmake
```diff
@@ -27,6 +29,8 @@ function( target_link_contrib ) # 2 arguments: ARGV0 = project name / ARGV1 = sh
 
 	#liblas support
 	target_link_liblas( ${ARGV0} ${ARGV1} )
+    #PDAL support
```
asmaloney: Formatting for whole PR - from CONTRIBUTING: "Indentation is expected to be made in Tabs only, each of size 4."

libs/qCC_io/LASFilter.cpp
```diff
@@ -32,1408 +32,1167 @@
...
 class LASSaveDlg : public QDialog, public Ui::SaveLASFileDialog
 {
 public:
-	explicit LASSaveDlg(QWidget* parent = 0)
-		: QDialog(parent)
-		, Ui::SaveLASFileDialog()
-	{
-		setupUi(this);
-	}
+    explicit LASSaveDlg(QWidget* parent = 0)
```
asmaloney: For whole PR: use nullptr, not 0 for pointers

libs/qCC_io/LASFilter.cpp
```diff
@@ -32,1408 +32,1167 @@
...
-		return 0;
-	}
-
-	QString fieldName;
-	Type valType;
-	int dataOffset;
-	double scale;
-	double offset;
+    //! Default constructor
+    ExtraLasField(QString name, Id id, double defaultVal = 0, double min = 0.0, double max = -1.0)
```
asmaloney: `double defaultVal = 0.0`

libs/qCC_io/LASFilter.cpp
```diff
@@ -32,1408 +32,1167 @@
...
+        : LasField(LAS_EXTRA,defaultVal,min,max)
+        , fieldName(name)
+        , pdalId(id)
+        , scale(1.0)
+        , offset(0.0)
+    {}
+
+    //reimplemented from LasField
+    virtual inline QString getName() const	{ return fieldName; }
```
asmaloney: Remove extra spacing after const

libs/qCC_io/LASFilter.cpp
```diff
@@ -32,1408 +32,1167 @@
...
+    if (parameters.parentWidget)
+    {
+        pDlg.reset(new ccProgressDialog(true, parameters.parentWidget)); //cancel available
+        pDlg->setMethodTitle(QObject::tr("Save LAS file"));
+        pDlg->setInfo(QObject::tr("Points: %L1").arg(numberOfPoints));
+        pDlg->start();
+    }
+    CCLib::NormalizedProgress nProgress(pDlg.data(), numberOfPoints);
+
+	LasWriter writer;
```
asmaloney: Spacing

libs/qCC_io/LASFilter.cpp
```diff
@@ -32,1408 +32,1167 @@
...
+    IdList dims = table.layout()->dims();
+    for (auto &dimId: dims)
+    {
+        std::cerr << "dim: " << name(dimId) << std::endl;
+    }
+    table.layout()->registerDim(Id::X);
+    table.layout()->registerDim(Id::Y);
+    table.layout()->registerDim(Id::Z);
+
+	unsigned ptsWritten = 0;
```
asmaloney: Spacing

libs/qCC_io/LASFilter.cpp
```diff
@@ -32,1408 +32,1167 @@
...
 		}
-
+		
 		//additional fields
 		for (std::vector<LasField>::const_iterator it = fieldsToSave.begin(); it != fieldsToSave.end(); ++it)
 		{
-			assert(it->sf);
+			std::bitset<8> classFlags;
+			Id pdalId;
+		    assert(it->sf);
```
asmaloney: Spacing. Q_ASSERT()?

libs/qCC_io/LASFilter.cpp
```diff
@@ -32,1408 +32,1167 @@
...
 
-	lasWriter.close();
+		nProgress.oneStep();
 
-	return result;
+		++ptsWritten;
+		return true;
+	};
+
+	//Set offset & scale, as points will be stored as boost::int32_t values (between 0 and 4294967296)
```
asmaloney: Won't continue to point it out. Please fix indentation throughout.

libs/qCC_io/LASFilter.cpp
```diff
@@ -32,1408 +32,1167 @@
...
+        assert(Zdim < 3);
+        Z = Zdim;
+        X = (Z == 2 ? 0 : Z+1);
+        Y = (X == 2 ? 0 : X+1);
+
+        bbMinCorner = bbMin;
+        tileDiag = bbMax - bbMin;
+        tileDiag.u[X] /= width;
+        tileDiag.u[Y] /= height;
+
```
asmaloney: Remove extra blank line

libs/qCC_io/LASFilter.cpp
```diff
@@ -32,1408 +32,1167 @@
...
+        //File extension
+        //        QString ext = (header.Compressed() ? "laz" : "las");
+
+        for (unsigned i = 0; i < width; ++i)
+        {
+            for (unsigned j = 0; j < height; ++j)
+            {
+                unsigned ii = index(i, j);
+                //TODO: Don't forget to change the ext to be able to write .laz
```
asmaloney: Is this something that needs to be done before merging?

libs/qCC_io/LASFilter.cpp
```diff
@@ -32,1408 +32,1167 @@
...
+        int ji = static_cast<int>(floor(Prel.u[Y] / tileDiag.u[Y]));
+        unsigned i = std::min( static_cast<unsigned>(std::max(ii, 0)), w-1);
+        unsigned j = std::min( static_cast<unsigned>(std::max(ji, 0)), h-1);
+        PointViewPtr outputView = tilePointViews[index(i,j)];
+        outputView->appendPoint(*buffer, pointIndex);
+    }
+
+    void writeAll()
+    {
+
```
asmaloney: Remove blank line

libs/qCC_io/LASFilter.cpp
```diff
@@ -32,1408 +32,1167 @@
...
-		Z = Zdim;
-		X = (Z == 2 ? 0 : Z+1);
-		Y = (X == 2 ? 0 : X+1);
+    inline unsigned index(unsigned i, unsigned j) const { return i + j * w; }
 
-		bbMinCorner = bbMin;
-		tileDiag = bbMax - bbMin;
-		tileDiag.u[X] /= width;
-		tileDiag.u[Y] /= height;
+    unsigned w, h;
```
asmaloney: Separate lines for each member variable.

libs/qCC_io/LASFilter.cpp
```diff
@@ -32,1408 +32,1167 @@
...
+			lasFields.push_back(LasField::Shared(new LasField(LAS_FLIGHT_LINE_EDGE, 0, 0, 1))); //1 bit: 0 or 1
+		if (s_lasOpenDlg->doLoad(LAS_SCAN_ANGLE_RANK))
+			lasFields.push_back(LasField::Shared(new LasField(LAS_SCAN_ANGLE_RANK, 0, -90, 90))); //signed char: between -90 and +90
+		if (s_lasOpenDlg->doLoad(LAS_USER_DATA))
+			lasFields.push_back(LasField::Shared(new LasField(LAS_USER_DATA, 0, 0, 255))); //unsigned char: between 0 and 255
+		if (s_lasOpenDlg->doLoad(LAS_POINT_SOURCE_ID))
+			lasFields.push_back(LasField::Shared(new LasField(LAS_POINT_SOURCE_ID, 0, 0, 65535))); //16 bits: between 0 and 65536
+
+		//extra fields
+		for (unsigned i = 0; i < extraNamesToLoad.size(); ++i)
```
asmaloney: range-for?

libs/qCC_io/LASFilter.cpp
```diff
@@ -32,1408 +32,1167 @@
...
-								for (size_t j = 0; j < count; ++j)
-								{
-									const EVLR* evlr = reinterpret_cast<const EVLR*>(vlrData + j*EB_RECORD_SIZE);
-									evlrs.push_back(*evlr);
-									ccLog::PrintDebug(QString("[LAS] Extra bytes VLR found: %1 (%2)").arg(evlr->getName()).arg(evlr->getDescription()));
-								}
-							}
-						}
-					}
+					auto cMin = static_cast<int>(field->sf->getMin());
```
asmaloney: auto -> int

libs/qCC_io/LASFilter.cpp
```diff
@@ -32,1408 +32,1167 @@
...
+					auto cMax = static_cast<int>(field->sf->getMax());
```
asmaloney: auto -> int

libs/qCC_io/LASFilter.cpp
```diff
@@ -32,1408 +32,1167 @@
...
+		if (s_lasOpenDlg->doLoadEVLR(i)) {
+			extraFieldsToLoad.push_back(extraDimensionsIds[i]);
+			extraNamesToLoad.push_back(extraDimensions[i]);
 		}
+	}
 
-		//dialog to choose the fields to load
-		if (!s_lasOpenDlg)
+    bool tiling = s_lasOpenDlg->tileGroupBox->isChecked();
+
```
asmaloney: Remove blank line

libs/qCC_io/LASFilter.cpp
```diff
@@ -32,1408 +32,1167 @@
...
+            break;
+        case 2: //YZ
+            vertDim = 0;
+            break;
+        default:
+            assert(false);
+            break;
+        }
+
+        auto w = static_cast<unsigned>(s_lasOpenDlg->wTileSpinBox->value());
```
asmaloney: auto -> unsigned int (same below)

libs/qCC_io/LASFilter.cpp
```diff
@@ -32,1408 +32,1167 @@
...
-		s_lasOpenDlg->setDimensions(dimensions);
-		s_lasOpenDlg->clearEVLRs();
-		s_lasOpenDlg->setInfos(filename, nbOfPoints, bbMin, bbMax);
-		if (extraDimension)
+		
+		while (!reader.isFinished())
 		{
-			assert(!evlrs.empty());
-			for (const EVLR& evlr : evlrs)
+#if defined(CC_WINDOWS)
```
asmaloney: Is there a way around using Sleep/usleep? To me this is a red flag that something isn't implemented properly w.r.t. threads or concurrency.

Is there a way to pass in a callback to the reader to handle events processing and then simply use `reader.waitForFinished()` here?

libs/qCC_io/LASFilter.cpp
```diff
@@ -32,1408 +32,1167 @@
...
+			pDlg->setMethodTitle(QObject::tr("LAS file"));
+			pDlg->setInfo(QObject::tr("Please wait... writing in progress"));
+			pDlg->setRange(0, 0);
+			pDlg->setModal(true);
 			pDlg->start();
 		}
-		CCLib::NormalizedProgress nProgress(pDlg.data(), nbOfPoints);
+		while (!writer.isFinished())
+		{
+#if defined(CC_WINDOWS)
```
asmaloney: Same comment as above - Sleep seems strange. Can it be avoided in favour of `writer.waitForFinished()` somehow?

libs/qCC_io/LASFilter.cpp
```diff
@@ -32,1408 +32,1167 @@
...
+        return CC_FERR_BAD_ARGUMENT;
+
+    ccGenericPointCloud* theCloud = ccHObjectCaster::ToGenericPointCloud(entity);
+    if (!theCloud)
+    {
+        ccLog::Warning("[LAS] This filter can only save one cloud at a time");
+        return CC_FERR_BAD_ENTITY_TYPE;
+    }
+
+    unsigned numberOfPoints = theCloud->size();
```
asmaloney: So this is just my opinion - I don't think our brief style guide has anything to say about it - but I prefer `unsigned int` to simply `unsigned`. Whichever is chosen, please use consistently throughout.

libs/qCC_io/LASFilter.cpp
```diff
@@ -32,1408 +32,1167 @@
...
-					ccLog::Warning("[LAS] Not enough memory!");
-					delete loadedCloud;
-					ifs.close();
-					return CC_FERR_NOT_ENOUGH_MEMORY;
-				}
-				loadedCloud->setGlobalShift(Pshift);
+		//first point check for 'big' coordinates 
+		if (nbPointsRead == 0)
+		{
+
```
asmaloney: Remove extra blank line

libs/qCC_io/LASFilter.cpp
```diff
@@ -32,1408 +32,1167 @@
...
-					else //otherwise we ignore it for the moment (we'll add it later if necessary)
-					{
-						pushColor = false;
-					}
-				}
+        CCVector3 P(static_cast<PointCoordinateType>(point.getFieldAs<double>(Id::X) + Pshift.x),
+                    static_cast<PointCoordinateType>(point.getFieldAs<double>(Id::Y) + Pshift.y),
+                    static_cast<PointCoordinateType>(point.getFieldAs<double>(Id::Z) + Pshift.z));
+        loadedCloud->addPoint(P);
+
```
asmaloney: Remove extra blank line

libs/qCC_io/LASFilter.cpp
```diff
@@ -32,1408 +32,1167 @@
...
+                {
+                    pushColor = false;
+                }
+            }
+            if (pushColor)
+            {
+                //we test if the color components are on 16 bits (standard) or only on 8 bits (it happens ;)
+                if (!forced8bitRgbMode && colorCompBitShift == 0)
+                {
+                    if (   (red   & 0xFF00)
```
asmaloney: Spacing

libs/qCC_io/LASFilter.cpp
```diff
@@ -32,1408 +32,1167 @@
...
+                           || (green  & 0xFF00)
+                           || (blue & 0xFF00))
+                    {
+                        //the co
```
asmaloney: "{" on next line, remove blank line

asmaloney: auto -> ScalarType

asmaloney: Remove blank line

asmaloney: Remove blank line

libs/qCC_io/LASFilter.cpp
tmontaigu: I forgot this part, It is now done

libs/qCC_io/LASFilter.cpp
tmontaigu: I used a code that I found in the BinFilter.cpp.
By changing the QFuture to a QFutureWatcher and using signals the ugly while loop & sleep can be replaced by waitForFinished()

tmontaigu:
Sorry for the formatting it was a mess 😢 , I reformated the file I hope it is better now.

About the Tiling, I could ask for some guidance to know if there is a way to do it without loading the whole file.

A similar question was asked las month in the mailinglist [(first post)](http://lists.osgeo.org/pipermail/pdal/2017-July/001343.html).

To summerize PDAL works with stages (reader, writer, filter) and you compose a pipeline.
For the tiling process, the pipeline could be : LasReader |  Tiler.
In Streaming "mode" for each stage the function procesOne(point) is called, this function is private which prevents me from making a Tiler that has multiple LasWriter and handle the call the the right writer. And I can't subclass the LasWriter either because I still need to be able to call the LasWriterr::processOne().

There might be other solutions but the mailing list thread makes me think that it is not possible without modifying something in PDAL 🤔 .

libs/qCC_io/LASFilter.cpp
asmaloney: Aha. BinFilter should be looked at too then :-) Thanks.

asmaloney:
Good stuff. Thanks for your work on this!

As for the technical part about what is supported and so on before merging, I leave that to @dgirardeau ...

abellgithub:
Hi,

I wrote the PDAL code you're using here.  I'll look this over and see if I can address your concern noted above.

hobu:
> For the tiling process, the pipeline could be : LasReader | Tiler.

We have a better solution to the tiling problem. We wrote open source software for point cloud web services called [Greyhound](http://greyhound.io) that allows you to access massive collections of point cloud data octree-style however you want. See [this article](http://lidarnews.com/articles/open-source-point-cloud-web-services-with-greyhound/) I wrote for LiDARNews describing how it works.

PDAL can speak to Greyhound for simple data access with its [readers.greyhound](https://www.pdal.io/stages/readers.greyhound.html), but that's not how a rendering client would typically query the data to keep the view filled with data. 

We have web services online for all of the LiDAR for Denmark (400 billion points), Netherlands (640 billion points), Dublin, New York City, Minnesota, Iowa, and so on. The user would be expected to "open" the resource by providing a URL such as http://data.greyhound.io/resource/dublin/info and the software can unpack all it needs to be able to query the data. See the [Greyhound Client development document](https://github.com/hobu/greyhound/blob/master/doc/clientDevelopment.rst) for how a rendering client might hit a Greyhound service. The [PDAL readers.greyhound driver](https://github.com/PDAL/PDAL/tree/master/plugins/greyhound/io) might also be instructive, but it isn't going to incrementally fetch data based on a view frustum as users pan around a GUI.

Finally, you can see two WebGL clients that can consume data from these services -- Potree http://potree.entwine.io and Plasio http://speck.ly We can't do everything with a browser, and we would be very excited if a rich client such as CloudCompare provided support for these point cloud web services.

dgirardeau:
Thanks Howard.

I remember Greyhound, and I still think it's a good idea to integrate it in CC. However, the tiling function was meant to be used on big LAS files (that are generated by some handheld laser scanners typically, but it could also be airborne lasers). And people just want to check their data on the field for instance (therefore I'm not sure a web service is suitable here). Any idea?

hobu:
> And people just want to check their data on the field for instance (therefore I'm not sure a web service is suitable here). Any idea?

[Entwine](https://entwine.io) is what does the tiling/indexing for Greyhound, and that could easily be used with software with no network.

asmaloney:
It's unclear from the entwine page whether it's cross-platform or not. Obviously it works on Linux since they put it in a docker container.

(FWIW there isn't a homebrew for it on macOS, so it would have to be included in CC sources or built separately.)

hobu:
Entwine has been built on Windows, but we don't have packages yet. I could do something through OSGeo4W.

A Homebrew package should be straightforward. I'll take that as a task.

abellgithub:
I looked over the code and I'm afraid I don't have a whole lot to offer at the current time, other than Howard's suggestion above, which provides more flexibility for your application than tiling.  That said, PDAL can do your tiling directly, if that's your goal.  You can simply place a splitter filter between a LAS reader and writer and let it rip.  You can provide a pipeline file (JSON) to a PipelineManager and tell it to execute.  It could all be in a subprocess if you like.  Once the output has been written, you can open the tiled points as you wish and process normally.  You may need to rename the files to suit your scheme, as PDAL doesn't (currently) have a way to provide output names based on input area.

tmontaigu:
I saw that PDAL already had a splitter filter but I did not use because it seems that it only alows to create squared tiles while the tiler in CC allows to create rectangular tiles.

Also, if I'm correct, it does not work in streaming mode which means that the whole cloud is loaded in memory and that's what we want to avoid.

dgirardeau:
I think I'll merge this soon, and either I'll move the 'tiling' option in the other LAS filter (based on LASlib, but it only works on Windows I believe) or we'll simply discontinue this option (people can still use older versions if it's only for tiling?).

PR #576, opened 2017-08-20
Switched the MainWindow UI from inheritance to member variable

asmaloney:
- removes UI header from mainwindow.h
- adds a method to allow adding the "Edit Plane" action to a context menu


dgirardeau:
Already done?! Impressive !

PR #587, opened 2017-08-31
Calculate normals with adaptive radius

pierstitus:
## Old Behavior
After normal calculation with a certain radius the points with too little neighbors within this radius don't get a normal.

## New Behavior
The search radius is increased on a point by point base to be able to compute normals also for points that are more sparse than average.
The values are now hard coded to find the minimum number of points within 16 times the given radius. (3 for plane/ls and 6 for quadric)
The maximum radius could also be turned into an option if people prefer that.


dgirardeau:
Nice and simple, I like it.

Thanks for the contribution!

PR #589, opened 2017-08-31
qCompass updates

samthiele:
Significant updates to qCompass. There will be a few more to come, but the bulk can be added now so we can start ironing out the bugs.


asmaloney:
Thanks Sam. Can you please resolve the conflicts so it can be merged cleanly when the time comes?

(Also if there's any way to break this into multiple PRs it would be much easier/less time-consuming to review.)

samthiele:
That's now done @asmaloney (after a bit of fun with Travis). 

And this update is associated with some fairly significant restructuring (namely pulling a lot of the functionality out of the ccCompass class and into specific "tool" classes and associated data structures), so it would be hard to break it into several PRs. 

Cheers,
Sam

dgirardeau:
Thanks Sam!

PR #609, opened 2017-10-22
The new line width was not clamped like the point size is.

asmaloney:
Mimic the way the line width is clamped to prevent negative (or zero) line widths.


libs/qCC_glWindow/ccGLWindow.cpp
```diff
@@ -64,6 +64,8 @@ static OculusHMD s_oculus;
 
 const float ccGLWindow::MIN_POINT_SIZE_F = 1.0f;
 const float ccGLWindow::MAX_POINT_SIZE_F = 16.0f;
+const float ccGLWindow::MIN_LINE_WIDTH_F = 1.0f;
+const float ccGLWindow::MAX_LINE_WIDTH_F = 10.0f;
```
dgirardeau: I believe we use the MIN_POINT_SIZE_F and MAX_POINT_SIZE_F limits in the 'properties delegate' of CloudCompare's BD tree (dqCC/db_tree/ccPropertiesTreeDelegate.cpp) to fill the combo-boxes for both the point size and the line width. And we can still use the same limit (16) for lines as for points? It's a bit arbitrary anyway.

PR #612, opened 2017-10-25
Fix some grammar

asmaloney:
One would not say "Export clouds info" even for multiple clouds - it's "Export cloud info" for both singular and plural. (Same thing for planes.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ v2.9 - 10/22/2017
 		- 3 algorithms are available: average, median and weighted average
 
 	* New sub-menu 'Tools > Batch export'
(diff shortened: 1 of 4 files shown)


dgirardeau:
Ah, I always have doubts about these plural issues (in French we could put 's' everywhere, even at 'info' ;) ).

asmaloney:
Yes - this one is actually a bit strange because it's imperative. In the singular, you might say "Export (the) cloud's info" - where "the" is implied - but then the plural would be "Export (the) clouds's info" which is more difficult to read (but correct). And then- as you point out "info" is the plural of "info". :-)

I'm not enough of an English geek to able to point at the rules for it though - just know it when I see it :-)

Issue #584, opened 2017-08-26
Point size in real world units/scale points with distance


pierstitus:
## Current Behavior
Currently the point size is set in pixels, which makes many points overlap when zoomed out and shows sparsely spaced points when zoomed in.

## Expected Behavior
Make the points scale up when they're closer to the camera, so they appear to have a constant size in real world units.
This way the pointcloud can look like a nice solid object, without having to change the point size when you zoom.

## Context/Comments
The point size in real world units can be given as a float, and can be estimated automatically from the distance between points. The points can still be rendered in the same way (squares/disks depending on setting), only every point has to be scaled depending on the distance to the camera.

dgirardeau:
I wonder what the cost (in terms of time) would be to change the point size dynamically for each point. Maybe it's possible to do it with shaders?

pierstitus:
There are two cases:
1. Orthogonal projection is quite easy as it the point size is the same for all the points, so can be set with setPointSize (in ccPointCloud.cpp and in ccGLWindow.cpp):
```
glFunc->glPointSize(m_viewportParams.defaultPointSize / m_viewportParams.pixelSize * m_viewportParams.zoom);
```
This works perfectly.

2. Perspective projection needs a dynamic point size per point indeed. I tried it with
```
float pointsize_factor = static_cast<float>(height()) / (0.018f * getFov());
float attenuation[] = {0.0f, 0.0f, 1.0f/(pointsize_factor*pointsize_factor)};
glFunc->glPointParameterfv(GL_POINT_DISTANCE_ATTENUATION, attenuation);
```
But that doesn't seem to work, it shows buggy graphics on my hardware (Intel graphics).

A vertex shader works fine, with the added benefit that a points can be scaled with a scalar field. I didn't yet manage to get both colored and scaled points, as it's difficult to see through the forest of optimizations for different cases in ccPointCloud.cpp

Simple GLSL vertex shader:
```
void main(void)
{
    gl_Position = ftransform();
    gl_PointSize = max( 100.0 / gl_Position.w, 1.0);
}
```

Or an actual one with SF based point size:
```
// camera dependent scaling factor, should be calculated as:
// uf_heightOfNearPlane = static_cast<float>(height()) / (0.018f * getFov());
uniform float uf_heightOfNearPlane;

// point size in world units
attribute float at_pointSize;

void main(void)
{
    gl_Position = ftransform();
[...]

PR #621, opened 2017-11-16
Correctly load shaders when running from build dir

pierstitus:
Also increase verbosity of shader loading messages, as it can be helpful for debugging to know which shader it actually loads.

Just a small improvement while paying around a bit with shaders.

### DIFF
--- a/libs/qCC_glWindow/ccGLWindow.cpp
+++ b/libs/qCC_glWindow/ccGLWindow.cpp
@@ -802,19 +802,19 @@ bool ccGLWindow::initialize()
 				else
 				{
 					ccColorRampShader* colorRampShader = new ccColorRampShader();
-					QString shadersPath = ccGLWindow::getShadersPath();
+					QString shaderPath = ccGLWindow::getShadersPath() + QString("/ColorRamp/color_ramp.frag");
 					QString error;
-					if (!colorRampShader->loadProgram(QString(), shadersPath + QString("/ColorRamp/color_ramp.frag"), error))
+					if (!colorRampShader->loadProgram(QString(), shaderPath, error))
 					{
 						if (!m_silentInitialization)
-							ccLog::Warning(QString("[3D View %1] Failed to load color ramp shader: '%2'").arg(m_uniqueID).arg(error));
+							ccLog::Warning(QString("[3D View %1] Failed to load color ramp shader '%2': '%3'").arg(m_uniqueID).arg(shaderPath).arg(error));
 						delete colorRampShader;
 						colorRampShader = nullptr;
 					}
 					else
 					{
 						if (!m_silentInitialization)
-							ccLog::Print("[3D View %i] Color ramp shader loaded successfully", m_uniqueID);
+							ccLog::Print(QString("[3D View %1] Color ramp shader '%2' loaded successfully").arg(m_uniqueID).arg(shaderPath));
 						m_colorRampShader = colorRampShader;
 						params.colorScaleShaderSupported = true;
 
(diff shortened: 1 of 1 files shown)


libs/qCC_glWindow/ccGLWindow.cpp
```diff
@@ -6212,6 +6212,13 @@ QString ccGLWindow::getShadersPath()
 		
 		shaderPath = (theDir.absolutePath() + "/share/cloudcompare/shaders");
 	}
+	else if ( theDir.dirName() == "qCC" ) // when running from build dir
```
dgirardeau: What's this 'qCC' folder? On which OS are you working? (Linux I guess?)

asmaloney: I would also recommend that this be wrapped in `#ifdef QT_DEBUG`. No reason to have it active in a release version...

pierstitus: Jup, I'm using linux. There it builds by default in <source_dir>/build/qCC/, so it can find the shaders in <sourcedir>/qCC/shaders. (It can build in a different location of course, so it might be better to let cmake copy the files to the build dir and use that, but that's way more hassle.)

QT_DEBUG is an option, but is not strictly related to where it is run from, you can test a release version too by running it in the build dir (actually that's what I do now). It shouldn't get to this if clause anyway, since any installed version will be installed in a 'bin' dir. Saying that I'd say it makes sense to remove the reasonable default shader dir in the else clause.

A similar issue exists with plugins, they are not loaded when running from the build dir, but that's a bit more difficult since every plugin is build in its own directory.

For Mac OS X it is done with the `CC_MAC_DEV_PATHS` variable, plus you have to copy (or symlink) the shaders dir into your build dir yourself.
But if things can be done automatically I'd say that's definitely preferred.

BTW in the Mac case the way to go up one dir seems to be done in a dangerous way: `appPath.remove("MacOS");`. If there is somewhere 'MacOS' in the path it gets removed! (for example if you put your CloudCompare.app in a dir called 'nice MacOS tools' on your USB disk it looks for plugins and shaders in the dir 'nice  tools' :/ ). Should better be done with a QDir like the linux case or with '../'.

How is it handled on Windows? Is it possible to run from the build dir there? Does it load shaders and plugins?

I still think this can be merged as is, but let me know.

dgirardeau: Actually I always use the INSTALL project. On Windows you have the issue that if you don't do that, in the best case you'll miss the necessary DLLs, or in the worst case wrong versions will be used instead!

Is there any reason why you don't use 'make install'?

pierstitus: I like to keep a stable working version installed, so I can always run that, even if I'm testing some experimental feature in my build dir. That way I can better compare the experimental feature, and also I know I'll have a working version around when I need it. That way it's easier to spend some spare time on coding while not messing up the installed version.

In the case of shaders (which I'm now playing around with, see #584) it's even enough to just restart CloudCompare without recompile, which is way quicker. (I can find ways around it of course, and for now the easiest workaround is to have this commit in my feature branch and rebase it out in case of a PR)

The wrong version case can also exist in the Linux case because of the 'reasonable default' else clause which makes it look in /usr. Is there any reason to keep that? If you agree I can remove it in this PR too.

dgirardeau: Ok, I'll let @asmaloney and @rjanvier decide whether this can be merged in the main branch or not (I don't know enough about Linux!). And regarding the '/usr', same thing, I'm clueless :D

rjanvier: to be paragmatic: if you only change he setup in a non "make install" case AND do not change anything to the "make install" case I guess i'm ok with this PR. if you change the behavior of the install case, I'm afraid I won't be agree because we follow the Filesystem Hierarchy standard.

asmaloney:
Closing this because the path handling is now in `ccApplicationBase::setupPaths()` so the commit is no longer valid.

If you want to pursue it, I would recommend mimicking the `CC_MAC_DEV_PATHS` option from the macOS build for Linux to keep the path change out of release builds.

PR #627, opened 2017-12-06
E57Support.cmake: Support linux builds

andersgb:
I was able to build this on ubuntu 16.04, more or less following these steps (some inspiration taken from here http://www.danielgm.net/cc/forum/viewtopic.php?t=1322)

1. Download libE57 1.1.312
2. Apply these changes https://gist.github.com/andersgb/c62bb013b22ee57ee0c92b0752a22df6 
3. Install `libxerces-c-dev` with `apt` if needbe
4. Build libE57 
```
cd E57RefImpl_src-1.1.312 
mkdir build 
cd build 
cmake .. 
make 
sudo make install
```
5. Apply this PR to CloudCompare
6. Build CloudCompare
```
cd CloudCompare
mkdir build
cd build
cmake .. -DOPTION_USE_LIBE57=ON -DLIBE57_INSTALL_DIR=/usr/local/E57RefImpl-1.1.unknown-x86_64-linux-gcc54
make
sudo make install
```

------

As noted in the commit message, this might be breaking windows builds. Consider this an example implementation, I'm open to testing other variants of the cmake-file on Linux if desired.


dgirardeau:
Interesting contribution, thanks!

Could you maybe add some "if LINUX ... elseif WINDOWS ... endif" statements to keep the compatibility with the existing Windows version?

andersgb:
:+1: 
Added platform branching commit, better now? Original behavior should be intact now.

But, looking at this now, it might be that the `FindXercesC` module [1] is actually win32 compatible, which means we don't actually need the branching?


[1] https://cmake.org/cmake/help/v3.1/module/FindXercesC.html

dgirardeau:
Actually I remember there was an issue with FindXercesC (but I can't remember which one exactly).

dgirardeau:
Thanks for the contribution anyway ;)

PR #631, opened 2017-12-14
Fix mesh laplacianSmooth bug

zhangzscn:
when edgesCount is 0，the point coordinates will be nan


dgirardeau:
I wonder how the edge counter can be 0 (I guess there are unused vertices in the mesh?). But thanks for the fix anyway!

zhangzscn:
yes，there are unused vertices in the mesh

Issue #622, opened 2017-11-24
Feature request : Command line mode scalar field from RGB and arithmetic on several scalar fields


jo-chemla:
Hi, 
This is not a bug but a feature request. It would be nice to be able to
 - export RGB to scalar fields via command line mode, 
 - and do operations such as scalar operations on two scalar fields, therefore being able to do `-SF_OP 0 mult 256 -SF_OP 0 add_field 1` to add field 0 to 1 and store in field 1. 
Thanks a lot! 
PS : also, being able to crop a cloud to a box, not only along X, Y or Z direction would also be a great addition - specifying box by center, scale and rotation for example, or box corners, not AABB.

NicolasCadieux:
Hi,
On the second point, it would be nice to have at least the possibility to exact data on two data fields. The first example that comes to mind for me is with LiDAR data.  ex: extract last returns (if return number = number of returns = last returns.)  I don't think we can do that right?

somtri:
Both requests in the original post are in the command line now:

- RGB to scalar field: `-RGB_CONVERT_TO_SF`
- operations between two scalar fields: `-SF_OP_SF {SF 1 name or index} {operation} {SF 2 name or index}`

The PS is still open, though: `-CROP` takes an axis-aligned box only, in the form `Xmin:Ymin:Zmin:Xmax:Ymax:Zmax`, so there is no way to give an oriented box from the command line.

PR #635, opened 2017-12-23
Add scalar field convert to RGB and scalar field set color scale to command-line

jarett-lee:
One of my first times doing a pull request, please make sure to tell me if I'm messing up anywhere.

### DIFF
--- a/qCC/ccCommandLineCommands.h
+++ b/qCC/ccCommandLineCommands.h
@@ -1374,7 +1374,7 @@ struct CommandSFConvertToRGB : public ccCommandLineInterface::Command
 			{
 				cmd.warning(QString("cmd.warning: cloud '%1' has no scalar field (it will be ignored)").arg(pc->getName()));
 			}
-			else if (activeSFIndex >= 0)
+			else if (activeSFIndex < 0)
 			{
 				cmd.warning(QString("cmd.warning: cloud '%1' has no active scalar field (it will be ignored)").arg(pc->getName()));
 			}
--- a/qCC/ccCommandLineCommands.h
+++ b/qCC/ccCommandLineCommands.h
@@ -1325,7 +1325,7 @@ struct CommandSFColorScale : public ccCommandLineInterface::Command
 
 		for (size_t i = 0; i < cmd.clouds().size(); ++i)
 		{
-			ccScalarField* sf = cmd.clouds()[i].pc->getCurrentDisplayedScalarField();
+			ccScalarField* sf = static_cast<ccScalarField*>(cmd.clouds()[i].pc->getCurrentOutScalarField());
 			if (sf)
 			{
 				sf->setColorScale(scale);
--- a/qCC/ccCommandLineCommands.h
+++ b/qCC/ccCommandLineCommands.h
@@ -1321,7 +1321,7 @@ struct CommandSFColorScale : public ccCommandLineInterface::Command
 			return cmd.error(QString("Failed to read color scale file '%1'!").arg(filename));
 
 		if (cmd.clouds().empty())
-			return cmd.error(QString("No point cloud on which to compute SF gradient! (be sure to open one with \"-%1 [cloud filename]\" before \"-%2\")").arg(COMMAND_OPEN, COMMAND_SF_COLOR_SCALE));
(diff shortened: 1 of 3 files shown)


jarett-lee:
One thing that I'm not sure about is my use of `getCurrentDisplayedScalarField()` to get the scalar field.

Looking at the other places in the code, some commands checks for a SF, then displays the first SF if there isn't a displayed SF and other commands use `getCurrentOutScalarField`.

jarett-lee:
Related #622

qCC/ccCommandLineCommands.h
```diff
@@ -1338,6 +1339,52 @@ struct CommandSFColorScale : public ccCommandLineInterface::Command
...
+		{
+			mixWithExistingColors = true;
+		}
+		else if (mixWithExistingColorsStr != "FALSE")
+		{
+			return cmd.error(QString("Invalid boolean value after \"-%1\". Got '%2' instead of TRUE or FALSE.").arg(COMMAND_SF_CONVERT_TO_RGB, mixWithExistingColorsStr));
+		}
+
+		if (cmd.clouds().empty())
+			return cmd.error(QString("No point cloud on which to compute SF gradient! (be sure to open one with \"-%1 [cloud filename]\" before \"-%2\")").arg(COMMAND_OPEN, COMMAND_SF_CONVERT_TO_RGB));
```
dgirardeau: Error message should be updated ;)

qCC/ccCommandLineCommands.h
```diff
@@ -1338,6 +1339,52 @@ struct CommandSFColorScale : public ccCommandLineInterface::Command
...
+
+		for (size_t i = 0; i < cmd.clouds().size(); ++i)
+		{
+			ccHObject* ent = cmd.clouds()[i].getEntity();
```
dgirardeau: 'ent' is not really necessary here. 'pc' can be used instead below.

qCC/ccCommandLineCommands.h
```diff
@@ -1338,6 +1339,52 @@ struct CommandSFColorScale : public ccCommandLineInterface::Command
...
+			ccPointCloud* pc = cmd.clouds()[i].pc;
+			if (pc->getCurrentDisplayedScalarField())
```
dgirardeau: I'm not sure the 'currrent displayed scalar field' is set in the command line mode (as it has no meaning in this mode). If people use the 'SET_ACTIVE_SF' command before yours, then the ```setCurrentScalarField``` method will be called. It sets the _in_ and _out_ scalar field to the same value. Therefore you can use any of those, even though I think it would be more logical to use the _out_ one in your function (as you 'read' the values). In this case you'll have to use ```getCurrentOutScalarField``` method.

qCC/ccCommandLineCommands.h
```diff
@@ -1300,6 +1302,105 @@ struct CommandDropGlobalShift : public ccCommandLineInterface::Command
...
+		}
+
+		if (cmd.clouds().empty())
+			return cmd.error(QString("No point cloud on which to convert SF to RGB! (be sure to open one with \"-%1 [cloud filename]\" before \"-%2\")").arg(COMMAND_OPEN, COMMAND_SF_CONVERT_TO_RGB));
+
+		for (size_t i = 0; i < cmd.clouds().size(); ++i)
+		{
+			ccPointCloud* pc = cmd.clouds()[i].pc;
+			unsigned sfCount = pc->getNumberOfScalarFields();
+			int activeSFIndex = pc->getCurrentOutScalarFieldIndex();
```
jarett-lee: I found that 'SET_ACTIVE_SF' doesn't set the displayed scalar field and `setRGBColorWithCurrentScalarField` checks for a displayed field using `hasDisplayedScalarField`. To solve this, I just used `getCurrentOutScalarFieldIndex` and `setCurrentDisplayedScalarField` to set the active scalar field as the displayed scalar field.

jarett-lee:
I didn't test this very well. I'll comment/push again when I make sure it works correctly.

jarett-lee:
I fixed all the issues I could find and it seems to works correctly, as far as I can tell.

dgirardeau:
Thanks for the contribution!

PR #637, opened 2017-12-27
Make SF gradient command-line actually use the active SF

jarett-lee:
'-SF_GRAD' didn't work with '-SET_ACTIVE_SF'. In the warning message it says "the active one will be used by default", but this was not actually the case and the displayed scalar field on save would be used since the method `getCurrentDisplayedScalarField` was used.


jarett-lee:
I couldn't find a simple way to reset the displayed scalar field, so I left it changed to the active scalar field.

dgirardeau:
Yes, this 'currently displayed' scalar field thing is not really relevant in the command line mode. We should probably avoid it in this mode but it will ask for a little more work ;).

Thanks for the feedback once again.

PR #648, opened 2018-02-22
Be explicit about signed chars

hroncok:
Chars being signed is apparently an implementation detail, see [1].

This was failing the build on various architectures:

    ppc64le aarch64 ppc64 s390x armv7hl

With:

    CC/src/ChamferDistanceTransform.cpp:44:1:
    error: narrowing conversion of '-1' from 'int' to 'char' inside { }
    ...

[1] https://stackoverflow.com/q/31634962/1839451


dgirardeau:
Interesting! Thanks for the fix.

Issue #649, opened 2018-02-22
There is no such thing as qToLittleEndian<double>


hroncok:
## Current Behavior
When building on Big Endian system, the build fails with:

    ../libs/qCC_io/libQCC_IO_LIB.so: undefined reference to `double qbswap<double>(double)'

I've done some googling and this is what I think happens:

https://github.com/CloudCompare/CloudCompare/blob/9f11b27f031068e67592831f839e9dc33fd9dcbb/libs/qCC_io/ShpFilter.cpp#L512-L515

This call is transferred to "do nothing" on Little Endian systems. Hence it builds just fine.
On Big Endian systems however, this translates to `qbswap<double>(double)` and that doesn't exist.
[`qToLittleEndian` only works with integer types](http://lists.qt-project.org/pipermail/interest/2013-September/008704.html).

## Expected Behavior

CloudCompare should build on Big Endian arches as well.

## Steps to Reproduce (for bugs)

Build CloudCompare on ppc64 or s390x.

## Context/Comments

I'm (again) trying to package CloudCompare for Fedora. We have some BE architectures.

## Your Environment
* CC Version: 2.9.1
* OS & Version: Fedora 29
* Graphics card: N/A

dgirardeau:
Commit 301e982282f4d432e4b051c209b1cb238d91dd92 should fix this?

PR #654, opened 2018-02-23
Compilation on Big Endian architectures 2

hroncok:
See https://github.com/CloudCompare/CloudCompare/commit/301e982282f4d432e4b051c209b1cb238d91dd92#r27739603

Fixes https://github.com/CloudCompare/CloudCompare/issues/649


dgirardeau:
Very good, thanks!

PR #660, opened 2018-03-01
Only read WKT from spatial reference

davidcaron:
There was a bug reading and writing a LAS containing a spatial reference.

The pdal documentation specifies that the `a_srs` parameter can be a WKT string.

This seems to fix the issue.


dgirardeau:
Thanks!

PR #661, opened 2018-03-03
Fix invalid include path specification resulting from deprecated ROS sub-directory removal in PCL 

indivisibleatom:
This change should be merged if deemed harmless, else, tracked for changing when CloudCompare bumps up its official supported PCL version.

Deprecated ROS headers are removed from PCL in PointCloudLibrary/pcl@c50d8ada7e6556c4db6c81f6b8541c826fefda25.

The inclusion of these causes a compile error on Ubuntu16.04 (and indeed should for any other platform). 
The proposed change fixes only the compile error. Unfortunately, I do not know enough about cloud compare / PCL to note what, if anything, fails at run-time at the moment.

### DIFF
--- a/plugins/qPCL/PclUtils/utils/my_point_types.h
+++ b/plugins/qPCL/PclUtils/utils/my_point_types.h
@@ -19,7 +19,7 @@
 #define Q_PCL_PLUGIN_MY_POINT_TYPES_H
 
 //PCL
-#include <pcl/ros/register_point_struct.h>
+#include <pcl/register_point_struct.h>
 #include <pcl/point_types.h>
 
 //! PCL custom point type used for reading RGB data


dgirardeau:
Not sure to understand. On Windows, we were using the 1.8 version (and I updated it with the latest git version) and I don't have any compilation error, nor runtime failure.

indivisibleatom:
Hi,
Not sure to understand. On Windows, we were using the 1.8 versión (and I updated it with the latest git versión) and I don't have any compilation error, nor runtime failure. >>
I understand that this file is present on the 1.8 version of PCL. Everything on 1.8 should work perfectly fine, and, I believe we are in agreement here.

However, if you update to the latest git version, you should have a build failure most definitely, as, the header file being included (ros/register_point_struct.h) is no longer present in the latest PCL (please see github change linked in the pull request, which is the PCL change that deletes this file). 

The one reason that I can think of not getting a build error would be if the git pull doesn't make the newly deleted files disappear. I think this should happen, but, just to be sure can you also do a git clean -fd to remove untracked files and directories of PCL and then try building?

dgirardeau:
Ok, it was simply because the 'INSTALL' command of the compilation project only copies new files, and doesn't remove deprecated files.

PR #666, opened 2018-03-09
(LAS) Read and save extra dimensions

davidcaron:
Here is some explanation of the process:

The VLR record describing the extra bytes after each point is mandatory only in the Las format 1.4. Most software write it anyway in versions <= 1.3. the 1.4 specification says:

> This VLR has been added to LAS 1.4 to formalize a process that has been used in prior versions of LAS.

So PDAL doesn't read this VLR if the version is before 1.4. The user has to manually specify the names and datatypes of the extra bytes to read.

In this PR, I read the VLR manually, make a string of the names and data type, and pass that back to the PDAL reader.

### DIFF
--- a/libs/qCC_io/LASFields.cpp
+++ b/libs/qCC_io/LASFields.cpp
@@ -86,12 +86,6 @@ bool LasField::GetLASFields(ccPointCloud* cloud, std::vector<LasField>& fieldsTo
 					break;
 				}
 			}
-
-			//no correspondance was found?
-			if (!outBounds && (fieldsToSave.empty() || fieldsToSave.back().sf != sf))
-			{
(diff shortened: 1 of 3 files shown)


davidcaron:
Currently, when using the gui, the extra dimensions aren't saved by defaut (user has to check the box). And when using the command line, the extra dimensions **are** saved by default. Maybe this should be changed, I don't know what others thing.

The main issue is that for las versions <= 1.3, the extra bytes aren't expected in some softwares (even if correctly described in the VLR) and garbage is loaded.

davidcaron:
Here is a discussion I had with PDAL devs.

https://github.com/PDAL/PDAL/issues/1831

dgirardeau:
Thank you for this! I'm sure a bunch of people will be very interested.

PR #669, opened 2018-03-12
Fix when opening files from argv containing non ASCII

davidcaron:
Otherwise CloudCompare complains "file doesn't exist"


asmaloney:
Thanks David!

PR #672, opened 2018-03-15
{cmake} Use proper subdirs for db_root & devices

asmaloney:
Also remove unused include directory (qxt)


dgirardeau:
Nicer indeed!

PR #673, opened 2018-03-16
Plugin manager

asmaloney:
Pull the plugin UI management out of the main window into its own class (ccPluginManager).

It should be functionally equivalent to the current UI.

I put this in its own directory because I intend to rewrite the "About plugins" dialog and it will live here as well.

There are likely other cleanups with plugin handling that could move into here. I need to look at how CC and ccViewer are using and passing plugin info.


dgirardeau:
Nice, thanks!

PR #674, opened 2018-03-16
Remove useless extra qualification in Qcc_io class LASSaveDlg

ebatsin:
This PR just removes the extra qualification of three methods of `LASSaveDlg` which causes an error when compiling using g++


asmaloney:
Thanks @ebatsin!

PR #675, opened 2018-03-16
{cmake} Give CC's poor cousin some attention

asmaloney:
- move qCC files into an include so they are easier to manage
- remove unused ccMouse3DContextMenu.[h,cpp] from the list
- use the new add_subdirectory for devices


dgirardeau:
Thanks for taking care of this dusty cousin ;)

PR #691, opened 2018-03-25
New snap package

rjanvier:
- Update syntax
- Update Eigen dep
- Add e57 support
- Add PDAL las/laz support...
- .. and don't cry baby, I also added PCL.


rjanvier:
@andy please Squash and merge, since the history is a bit lame.

PR #702, opened 2018-04-05
{plugins} Move plugins to new directory "plugins/core" 

asmaloney:
- created a "3rdParty" directory
- created an "example" directory
- fixed several relative includes (including in plugins I could not build)
- after this you will have to fix at least one cmake var to add the "core" directory in the path: **NANOFLANN_ROOT_DIR**
- in the examples, the INSTALL_QEXAMPLE_PLUGIN becomes INSTALL_EXAMPLE_PLUGIN. etc. (no "q")


plugins/core/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+
+add_subdirectory( ${CMAKE_CURRENT_SOURCE_DIR}/qAdditionalIO )
+add_subdirectory( ${CMAKE_CURRENT_SOURCE_DIR}/qAnimation )
+add_subdirectory( ${CMAKE_CURRENT_SOURCE_DIR}/qBroom )
+add_subdirectory( ${CMAKE_CURRENT_SOURCE_DIR}/qCompass )
+add_subdirectory( ${CMAKE_CURRENT_SOURCE_DIR}/qCork )
+add_subdirectory( ${CMAKE_CURRENT_SOURCE_DIR}/qCSF )
+add_subdirectory( ${CMAKE_CURRENT_SOURCE_DIR}/qCSVMatrixIO )
+add_subdirectory( ${CMAKE_CURRENT_SOURCE_DIR}/qEDL )
+add_subdirectory( ${CMAKE_CURRENT_SOURCE_DIR}/qExampleGLPlugin )
```
dgirardeau: Why not keeping the automated way as before? And do we really want to add the example plugins?

asmaloney: CMake recommends against using GLOBs/auto-inclusion because some build systems don't get alerted about changes. Since this directory won't be changing that frequently I went to explicit adding of directories. The "auto-add" idea will be used in the 3rdParty directory to make it easy for developers to add their plugins there.

About the examples: That's part of step 2. They'll live in the 3rdParty directory (even though they aren't' really 3rd party).

asmaloney: ...or maybe an examples directory? That might make more sense.

asmaloney:
Looks like I also failed to move the Poisson submodule properly. Will fix...

plugins/core/CMakeLists.txt
asmaloney: Is this a deal-breaker? :-)

From the CMake [manual](https://cmake.org/cmake/help/v3.3/command/file.html):
> We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate.

asmaloney: OK thanks. The only two things that need changing per plugin should be:

-include( ../CMakePluginTpl.cmake )
+include( ../../CMakePluginTpl.cmake )

in the CMakeLists.txt file and removing relative paths:

-#include "../ccStdPluginInterface.h"
+#include "ccStdPluginInterface.h"

I changed all the plugins in the repo, so it should just be your "private" ones. CANUPO!

PR #720, opened 2018-04-24
Use QAtomicInteger<unsigned int> instead of QAtomicInt

asmaloney:
This is how m_counter is used and it avoids casting.

Also avoids problems if there are more than 2,147,483,647 calls to oneStep() ;-)


rjanvier:
@asmaloney sorry, I close this for now, since it's superseded by #722

PR #722, opened 2018-04-25
 [CCLib] Remove Qt dependency

rjanvier:
This basically mean to use TBB or PPL (depending on the plateform) for parallel features used in CC Libs instead of Qt:
- [x] ParallelSort.h -> Parallel.h
- [x] Replace QT by functions both present in PPL and TBB (parallel_for, parallel_for_each)
- [x] Remove CC special wrapper around QTAtomic and replace it by std::atomic
- [x] Add changes proposed by @asmaloney  in #720 
- [x] find an equivalent to task_arena for PPL.
- [ ] Wait for non experimental Xenial images on Travis

this is a strict 1:1 replacement for now but this pave the way for better parallelism (cancellation...) using task based parallelism from TBB/PPL


CC/CMakeLists.txt
```diff
@@ -54,14 +50,15 @@ if (COMPILE_CC_CORE_LIB_WITH_TBB)
 	set_property( TARGET ${PROJECT_NAME} APPEND PROPERTY COMPILE_DEFINITIONS USE_TBB )
 endif()
 
-if ( COMPILE_CC_CORE_LIB_WITH_QT )
+if (COMPILE_CC_CORE_LIB_WITH_QT)
```
asmaloney: Shouldn't this whole section be removed?

CC/CMakeLists.txt
```diff
@@ -54,14 +50,15 @@ if (COMPILE_CC_CORE_LIB_WITH_TBB)
...
+	include( ../cmake/CMakeExternalLibs.cmake )
 	include_directories( ${Qt5Widgets_INCLUDE_DIRS} ${Qt5Core_INCLUDE_DIRS} )
 	qt5_use_modules(${PROJECT_NAME} Core Widgets Concurrent)
 	set_property( TARGET ${PROJECT_NAME} APPEND PROPERTY COMPILE_DEFINITIONS USE_QT )
 endif()
 
 # Load advanced scripts
-include( ../cmake/CMakeInclude.cmake )
+include( ../cmake/CMakeInclude.cmake)
```
asmaloney: Why? :-)

CC/include/Parallel.h
```diff
@@ -15,31 +15,33 @@
...
-#ifndef PARALLEL_SORT_HEADER
-#define PARALLEL_SORT_HEADER
-
-#ifdef ParallelSort
-#undef ParallelSort
-#pragma message "Replacing preprocessor symbol 'ParallelSort' with the one defined in Parallel.h"
-#endif
+#ifndef CCCORE_PARALLEL_HEADER
+#define CCCORE_PARALLEL_HEADER
+#include <atomic>
```
asmaloney: Does this belong here?

CC/CMakeLists.txt
rjanvier: no, because there are still some QT artifacts, especially the use of QTBitArray somewhere

CC/include/Parallel.h
rjanvier: eagle eyes: I simply forgot to remove that after experiments

asmaloney: This is why code reviews rock!

CC/src/DgmOctree.cpp
```diff
@@ -3494,7 +3494,15 @@ unsigned DgmOctree::executeFunctionForAllCellsAtLevel(	unsigned char level,
...
-		CCParallelForEach(cells.begin(), cells.end(), LaunchOctreeCellFunc_MT);
+		if(maxThreadCount > 0)
+		{
+
+			CCParallelWithLimitedThreads(CCParallelForEach(cells.begin(), cells.end(), LaunchOctreeCellFunc_MT), maxThreadCount)
+		}
+		else
+		{
+			CCParallelForEach(cells.begin(), cells.end(), LaunchOctreeCellFunc_MT);
+		}
```
asmaloney: I would be against this style _in general_. It's going to get messy quickly, makes it much easier to make mistakes, and is going to be a lot of "no fun" to maintain.

I don't think there's any measurable benefit to it.

If we need a mechanism to "breathe" in these functions, then we should be calling a user-specified generic "yield" function that gets called periodically. Then the user of the API can decide what to do - put up a progress dialog, or simply (in the Qt case) called QApplication::processEvents() to keep the UI responsive.

Or if you don't want to get that fancy, then simply call `std::this_thread::yield` periodically within the executing code which should avoid hogging all the processing.

asmaloney: http://en.cppreference.com/w/cpp/thread/yield

CC/include/Parallel.h
```diff
@@ -35,10 +36,19 @@
 #include <tbb/parallel_for_each.h>
 #include <tbb/parallel_sort.h>
 #include <tbb/critical_section.h>
+#include <tbb/task_arena.h>
+#include <tbb/task_group.h>
 #define CCParallelSort tbb::parallel_sort
 #define CCParallelFor tbb::parallel_for
 #define CCParallelForEach tbb::parallel_for_each
 #define CCCriticalSection tbb::critical_section
+//Limit the number of thread of a parallel template function (or task)
```
asmaloney: Add:
```cpp
/// TODO: Remove this method in favour of providing a proper yield mechanism
```

CC/src/DistanceComputationTools.cpp
```diff
@@ -1088,8 +1088,7 @@ static bool s_cellFunc_MT_success = true;
 static CCLib::DistanceComputationTools::Cloud2MeshDistanceComputationParams s_params_MT;
 
 //'processTriangles' mechanism (based on bit mask)
-#include <QtCore/QBitArray>
-static std::vector<QBitArray*> s_bitArrayPool_MT;
+static std::vector<std::vector<bool>*> s_bitArrayPool_MT; //TODO: RJ:Do the std::vector<bool> really need to be heap allocated?
```
dgirardeau: Wow, a QBitArray is much smaller than a std::vector<bool>! 8 times smaller in the ideal case (assuming bool are coded on 8 bits, and I'm not even sure of that). You'll have to find a better structure to replace that...

asmaloney: std::vector<bool> should be optimized for space - but you'd have to check with your compiler.

http://www.cplusplus.com/reference/vector/vector-bool/

"The storage is not necessarily an array of bool values, but the library implementation may optimize storage so that each value is stored in a single bit."

CC/src/DgmOctree.cpp
dgirardeau: Then we still need an optional parameter to allow the system to breath or not?

Because some users will want to go as fast as possible, and they don't care that CC freezes the OS. But others will want to continue using their PC while the process is ongoing. In this case, calling QApplication::processEvents() only lets CC breath (but not the other applications). And it's hard to call "yield" on a regular basis (which is the only option to give a rather good experience to users wanting to continue working with their PC). If we call yield in the main octree fork for instance, the time between two calls would actually depend on the time spent on each cell, which is highly dependent on the number of cells (and the octree level by the way).

Am I missing something?

rjanvier: I think we should maybe care to keep CC breath for example to allow a more friendly cancellation or a smoother monitoring of the progress but we should not take care of the OS, the OS scheduler should be smart enough for us.

CC/src/DistanceComputationTools.cpp
rjanvier: The loss of performance is even more dramatic than what I exposed in the last message. so std::vector is not an option.

asmaloney: Doesn't look like our options are great:

https://www.academia.edu/20674648/Performance_of_C_bit-vector_implementations

`std::bitset` would be great, but the size is fixed at compile-time:

http://www.cplusplus.com/reference/bitset/bitset/

rjanvier: I also test with boost::dynamic_bitset, and performance are even worst :(, still investigating...

rjanvier: Ok using the right function make boot::dynamic_bitset on parity with QtBitset. The biggest performance bottleneck  with std::vector<bool> is the use of std::fill (see below)

CC/src/DistanceComputationTools.cpp
```diff
@@ -1174,23 +1173,25 @@ void cloudMeshDistCellFunc_MT(const DgmOctree::IndexAndCode& desc)
...
 		{
+			s_currentBitMaskMutex.lock(); //TODO: RJ: use a parallel_queue to avoid implicit lock
 			bitArray = s_bitArrayPool_MT.back();
 			s_bitArrayPool_MT.pop_back();
+			s_currentBitMaskMutex.unlock();
+
 		}
-		s_currentBitMaskMutex.unlock();
-		bitArray->fill(0);
+		std::fill(std::begin(*bitArray), std::end(*bitArray), 0);
```
rjanvier: here is the biggest performance bottleneck...

CC/src/DistanceComputationTools.cpp
```diff
@@ -1177,23 +1173,25 @@ void cloudMeshDistCellFunc_MT(const DgmOctree::IndexAndCode& desc)
...
-	QBitArray* bitArray = nullptr;
+	std::vector<bool>* bitArray = nullptr;
 	if (s_useBitArrays_MT)
 	{
-		s_currentBitMaskMutex.lock();
 		if (s_bitArrayPool_MT.empty())
 		{
-			bitArray = new QBitArray();
+			bitArray = new std::vector<bool>;
 			bitArray->resize(s_intersection_MT->mesh->size());
```
asmaloney: This can just be:
```
bitArray = new std::vector<bool>(s_intersection_MT->mesh->size());
```
It default initializes to `false`, so the `std::fill` can be moved into the "else". There might be a better way to do that as well.

CC/src/DistanceComputationTools.cpp
asmaloney: (See comment above first)

You might try `(*bitArray).assign(bitArray->size(), false);` instead of `std::fill`, but we're getting into micro-optimizations that are perhaps not useful because they will depend on the compiler/architecture, etc..

CC/include/Parallel.h
```diff
@@ -15,31 +15,50 @@
...
+#include <tbb/parallel_sort.h>
+#include <tbb/critical_section.h>
+#include <tbb/task_arena.h>
+#include <tbb/task_group.h>
+#define CCParallelSort tbb::parallel_sort
+#define CCParallelFor tbb::parallel_for
+#define CCParallelForEach tbb::parallel_for_each
+#define CCCriticalSection tbb::critical_section
+//Limit the number of thread of a parallel template function (or task)
+#define CCParallelWithLimitedThreads(task, num_threads) { \
```
rjanvier: @dgirardeau I'm nearly done with that. As I do not have Windows, could you please check how to limit concurrency with PPL? With TBB I use task_arena but there is no such feature in PPL.

CC/src/DgmOctree.cpp
```diff
@@ -3305,7 +3305,7 @@ unsigned DgmOctree::executeFunctionForAllCellsAtLevel(	unsigned char level,
 
 #ifdef ENABLE_MT_OCTREE
 
-	//cells that will be processed by QtConcurrent::map
+	//Those cells will be processed in parallel
```
asmaloney: "These" :-)

rjanvier: :) sorry

dgirardeau:
Ok I tested the Windows parallel stuff. It works (even the version with a limited number of threads). However it's slightly slower (like 5%).

rjanvier:
Well it's not really what I have expected. I don't know if it worth it...

PR #727, opened 2018-04-27
Replace QbitArray by a simple std::vector

rjanvier:
based on dicussion on #722 and on slack, It's a proposal to replace QbitArray with a simple std::vector<bool>. 
In libstd++ (gcc), it seems to have been implemented efficiently (1 bool = 1 bit) and at least since [g++4.6.3](https://gcc.gnu.org/onlinedocs/gcc-4.6.3/libstdc++/api/a01047_source.html), so it's ok since we required at least g++ 4.8 for a full C++11 support. There is **no loss** of performance with this implementation on Linux.
 @asmaloney, @dgirardeau could you please check if it's ok on your plateform?


asmaloney:
My testing shows that this is equivalent to QBitArray on macOS.

One hour to compute normals (on ~10 M points), then 2.5 seconds to compute mesh to point cloud distance using the QBitArray or std::vector<bool> :-)

PR #739, opened 2018-05-12
Fix las gps

### LINKED ISSUE #738: LAS gps_time (CC standard field: Time) not being recognized in CC v.2.10.alpha
## Current Behavior
When I load a version 1.2 LAS file into CC v2.9.1 the standard field gps_time is recognized as being present with the data. When I try to load the exact same LAS file into CC v2.10.alpha the Time field is disabled in the Open LAS Dialog.

## Expected Behavior
Have the Time field be recognized during LAS import.

## Steps to Reproduce (for bugs)
I can provide the specific LAS file if needed.

## Context/Comments
The version 1.2 LAS file is being created by a Python script (using laspy) where the script firstly creates several (currently up to 16) individual LAS files from the raw lidar data due to the employed multiprocessing approach. These LAS files contain 3 standard fields (Time, Return Num, Intensity) as well as 4 additional fields (new laspy dimensions), all numeric. The individual files are then merged together using lasmerge (run as a subprocess from Python). I don't believe there is a problem with the Python/laspy and lasmerge steps because the final merged LAS file is properly read into CC v2.9.1.

## Your Environment
* CC Version: 2.10.alpha
* OS & Version: Windows 7 x64
* Graphics card: NVIDIA Quadro K2200

tmontaigu:
This should fix #738


dgirardeau:
Thanks ;)

PR #751, opened 2018-05-31
Replace obsolete CMake macro qt5_use_module()

lowsfer:
I'm on CMake 3.11.2 and failed to run cmake because the obsolete qt5_use_module is no longer recognized. Updated the CMakeLists.txt files to make it work.


dgirardeau:
Thanks for the contribution!

PR #753, opened 2018-06-02
Enable CloudCompare to also link against shared library of E57Format

Tones29:
Adds the option to also link against a shared library of E57Format, e.g. provided here: https://github.com/Tones29/libE57Format.


dgirardeau:
Thanks

PR #754, opened 2018-06-04
Set CMAKE_INSTALL_LIBDIR to default, if undefined

Tones29:
Fixes a bug introduced with #746 

CMAKE_INSTALL_LIBDIR is not set by default in some environments, e.g. remains unset with cmake 3.5.1 on ubuntu 16.04. This causes all CloudCompare libs to be installed to `/cloudcompare` instead of `/usr/local/lib/cloudcompare` by default.


rjanvier:
thanks!

PR #759, opened 2018-06-26
UI french translation

Jean-Roc:
420/1599 strings done, some have a comment (search for "?") where the existing translation is far from its english counterpart or simply where I didn't understood the meaning


dgirardeau:
Thanks, that's already useful. I need to let user choose the langage in the Options dialog...

PR #761, opened 2018-07-03
Update in inno-setup for close the CloudCompare in uninstall process.

Mun1z:
We are developing a solution using cloud compare, our testers found strange behavior when uninstalling cloud compare, they were able to uninstall the cloud compare with it being open, this made the executable and the binaries that were being used at the moment were not removed during the deinstallation, with this change in the inno-setup scripts it will close the cloud compare if it is open during the uninstalling, thus removing completely from the machine.


dgirardeau:
Ah ah you have funny users ;)

Thanks for the contribution anyway.

PR #765, opened 2018-07-07
Add the options to constrain ICP rotation via the command line

arminklauser:
New option to constrain ROTATION for the ICP algorithm via the command line -ROT XYZ/X/Y/Z/NONE


dgirardeau:
Thanks for the contribution!

PR #766, opened 2018-07-09
Fix missing include if TBB is not available

nmoehrle:


asmaloney:
Thank you for the fix Nils!

PR #771, opened 2018-07-25
Implemented Structure Normal Estimation

samthiele:
Implemented a (work in progress) algorithm for estimating the normal vector to traces created by geological structures. More to come, but is usable (and stable?) enough to be useful for people as is.


dgirardeau:
Nice update!

PR #779, opened 2018-08-15
Normals whiskers

SalDiAngelus:
Normal whiskers can be enabled by right clicking on a point cloud in the db hierarchy window. They don't support coloring.
Simplified point cloud rendering by only using the fixed function pipeline with material shaders and VBOs. It should support all features rendering options.
Removed the color ramp shader.

Current issues:
I had to disable the optimized path of mesh drawing. Something with the new point cloud rendering causes a crash when glDrawArrays is called for a mesh.
The LOD structure crashes for me. I'm not sure what causes it yet.


SalDiAngelus:
Meant to pull to my fork sorry.

PR #783, opened 2018-09-20
Fix build on archs where char is unsigned

### LINKED ISSUE #709: Debian package fails to build for some architectures
Please find the build logs at
https://buildd.debian.org/status/package.php?p=cloudcompare

You might maybe not want to support one or another architecture (because you don't have it, or
can't even buy it anymore).

frediz:
For that specific source file, just add -fsigned-char for compilation.
This closes #709.

### DIFF
--- a/CC/CMakeLists.txt
+++ b/CC/CMakeLists.txt
@@ -60,6 +60,9 @@ if ( COMPILE_CC_CORE_LIB_WITH_QT )
 	set_property( TARGET ${PROJECT_NAME} APPEND PROPERTY COMPILE_DEFINITIONS USE_QT )
 endif()
 
+# Some archs have char unsigned
+set_source_files_properties(src/ChamferDistanceTransform.cpp PROPERTIES COMPILE_FLAGS -fsigned-char)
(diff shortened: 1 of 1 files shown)


asmaloney:
Does this change have any consequences for the code in `ChamferDistanceTransform.cpp` or are you just trying to "get it to compile"?

Adding a specific compile flag for one file is a Bad Idea. If there's a problem with that one file then it should be addressed properly.

frediz:
It was an implementation of the proposal done here : https://github.com/CloudCompare/CloudCompare/issues/709#issuecomment-383362501 
Since the code just expects signed char from what I see, it just tells gcc to use signed chars which corresponds to what the code needs.
A better solution would just be to be more precise and be explicit and use signed char instead of char.
And I actually see that this has been done on master (I was checking the logs provided in the bug)...
but the bug #709 has not been closed then.
So, sorry for the noise, nothing to be done.
Thanks for your time.

asmaloney:
No problem!

Thanks for taking the time to look into it.

PR #784, opened 2018-09-25
Add the option to constrain ICP translation+rotation via the command line

arminklauser:
New option to constrain TRANSLATION for the ICP algorithm via the command line -TRANS X/Y/Z/XY/XZ/YZ/XYZ/NONE
Changed the code to make option ROTATION working


CC/src/RegistrationTools.cpp
```diff
@@ -81,7 +87,7 @@ void RegistrationTools::FilterTransformation(	const ScaledTransformation& inTran
 				//simpler/faster to ignore this (very) specific case!
 			}
 		}
-		else if (filters & SKIP_RXZ) //keep only the rotation component around Y
```
dgirardeau: I don't believe you can actually release two rotations axes at the same time. Especially since in the way it is currently implemented, some elements of the matrix set in the first block will be overwritten by the next block. And you'll end with rubbish...

qCC/ccCommandLineCommands.h
```diff
@@ -3570,15 +3569,15 @@ struct CommandICP : public ccCommandLineInterface::Command
 				{
 					QString rotation = cmd.arguments().takeFirst().toUpper();
 					if (rotation == "XYZ")
-						rotFilter = SKIP_NONE;
+						rotFilter = 0;
```
dgirardeau: Too bad. Why not keeping the enum version?

CC/src/RegistrationTools.cpp
dgirardeau: Up?

asmaloney:
@arminklauser:

If you want to revive this PR, you will need to update it due to changes in ccCommandLineCommands and also address Daniel's concerns.

Please feel free to reopen it or create a new PR if you would like to continue.

Thank you!

PR #788, opened 2018-10-06
Remove overrides and clean up virtuals in PointCloudTpl.h

asmaloney:
As discussed in Slack, this template should not be overriding anything in the base class, so remove all override keywords. In addition, prune back the number of virtual functions to ones that are used.

This is kind of a fragile class since it's easy to unintentionally override or shadow members and functions in the base class, or mask the ones here in a class that inherits from this template. The structure confuses clang's static analysis tools, so it's difficult to figure out what's noise and what's correct.

At least one issue remains: clear() is in classes at all three levels, so it is not working as intended. This will require some discussion.

### DIFF
--- a/CC/include/PointCloudTpl.h
+++ b/CC/include/PointCloudTpl.h
@@ -49,10 +49,9 @@ namespace CCLib
 			deleteAllScalarFields();
 		}
 
-		//**** inherited form GenericCloud ****//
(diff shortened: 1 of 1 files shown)


dgirardeau:
Thanks Andy (sorry for the wau, I'm finally back from the US ;) )

PR #790, opened 2018-10-07
Fix sensor depth map writing crash

### LINKED ISSUE #729: Writing DepthMap asserts on sensor data
## Current Behavior
Attempting to write a depth map for a sensor asserts in debug build in `ccGBLSensor::projectColors`.

## Steps to Reproduce (for bugs)

* load file with sensor data (I used `pump_scan3.ptx` in the test data)
* select the sensor  ("TLS/GBL" in this case")
* create the depth buffer using "Edit > Sensors > TLS/GBL > Create/show depth buffer"
* save the sensor as type "Depth Map"

## Your Environment
* CC Version: 2.10 alpha
* OS & Version: macOS

asmaloney:
ccGBLSensor was allocating an incorrect type which was causing issues with deleting the object

Although the types seem to be compatible, there's an issue deleting a CCShareable directly in DepthMapFileFilter::saveToFile().

Fixes #729


dgirardeau:
Nice catch!

PR #807, opened 2018-11-11
Update Build.md PDAL

tmontaigu:
Update the Build.md file and cmake files to change the references from liblas to PDAL


rjanvier:
@asmaloney I think you are still using liblas for OS X bundle?

dgirardeau:
I guess it's ok as Andy uses a plugin anyway

PR #811, opened 2018-11-16
Save las point format & version

tmontaigu:
Currently  when a file is saved as LAS, the version & point format (an id that gives the defaults dimensions present in the file), it is saved with the default values chosen by PDAL (version 1.2 & point format 3)

Which means that if you open a LAS file with a version & point format different than the defaults,
when saving, the file version & point format changes.

This is mostly a problem when opening LAS files with version >= 1.4 & point format >= 6 because
it would downgrade the file and even sometime CC could not save the file as some fields in point format >= 6 are not compatible with point format 3.

This PR adds the code to save the LAS file version & point format as meta-data and try to load them when saving a file in order to preserve the original version & point format of the loaded file.

I also did a bit of 'refactoring'.


dgirardeau:
Nice addition!

PR #812, opened 2018-11-28
Started main window translation to brazilian portuguese

fabriciolelis:
Main window translation to Brazilian Portuguese 90% done.


fabriciolelis:
@asmaloney How can I help you with the plugins translation?

asmaloney:
I need to find the time to go through and clean up what gets translated in all of the plugins and some other files that are not currently included in the translation.

This means removing things that should not be added like "%", "...", any numbers, etc., and fixing some spelling/grammar mistakes in the English.

I hope to do that in the next couple of days, then I will upload new versions of the .ts files that includes a bunch of new things to translate.

So you could continue to submit updated ts files for Portuguese and I will merge them. (If you have a lot of changes that are not merged before I do this large update to the ts files, it might make it more difficult at your end.)

Thank you!

PR #820, opened 2018-12-17
Improve ShpFilter

tmontaigu:
Hey, sorry in advance for the big PR 😅
I had problem loading some shape files in CC (type Polyline M & Point Z) and ended up refactoring the code.

This PR fixes the loading of "Polyline M" files (2D polylines with measure
scalar field) and the loading of "Point M" & "Point Z" files (Single point
records with measure scalar field & Z)

I took the opportunity to add some support for the Multipatch (Meshes) shape
type, with this PR CC can read & write shape files for which the parts of a
"multi patch shape file" is "triangle fan" or "triangle strip"
(The file multipatch.m in the test data is an example of such a thing)

This PR also adds test of the reading & writing of the ShpFilter using the 
QTest framework.

The following files (which are present in libs/qcc_io/Tests/Data/shp) do not work with current CC:
 - linem.shp (Polyline with measures scalar field, Fails with unhandled type)
 - multipatch.shp (Multipatch shape that is a triangle mesh, Fails with
   unhandled type)
 - pointz.shp (Shape where each point (x, y, z, m) has its own record, Fails
   with "Exception Invalid `vector<T>` subscript")

In terms on coding the refactor boils down to:
 - trying th split some of the big functions into smaller ones
 - using QDataStream to read & write files which makes handling endianness
 easier
 - Using enum class instead of regular enums


libs/qCC_io/Tests/CMakeLists.txt
```diff
@@ -1,13 +1,22 @@
...
+
+set(TEST_LIBRARIES Qt5::Test Qt5::Core CC_CORE_LIB QCC_DB_LIB QCC_IO_LIB)
+
+if (WIN_32)
+    SET(CMAKE_WIN32_EXECUTABLE False)
+    set(TEST_LIBRARIES ${TEST_LIBRARIES} Qt5::WinMain)
+endif ()
+
+if (OPTION_USE_SHAPE_LIB)
+    message("Use hapefils")
```
asmaloney: Remove debug message?

libs/qCC_io/Tests/TestShpFilter.cpp
```diff
@@ -505,7 +498,7 @@ void TestShpFilter::testWriteMultpatchFile() const
 
 	error = filter.saveToFile(&container, tmpMultipatch, saveParams);
 	QVERIFY(error == CC_FERR_NO_ERROR);
-	readMultipatchFile(tmpMultipatch);
+	//readMultipatchFile(tmpMultipatch);
```
asmaloney: Remove commented-out code?

libs/qCC_io/ShpFilter.cpp
```diff
@@ -1,1846 +1,1892 @@
...
+//#  This program is distributed in the hope that it will be useful,       #
+//#  but WITHOUT ANY WARRANTY; without even the implied warranty of        #
+//#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the          #
+//#  GNU General Public License for more details.                          #
+//#                                                                        #
+//#          COPYRIGHT: EDF R&D / TELECOM ParisTech (ENST-TSI)             #
+//#                                                                        #
+//##########################################################################
+
+#ifdef CC_SHP_SUPPORT
```
dgirardeau: Are you using space-based tabulations maybe? It's hard to see the differences 😢

tmontaigu: I use tabs for indentation, so it should be tabs 🤔

dgirardeau: Do you have a clue why Git(hub) believes everything has changed then?

tmontaigu: It may be because I changed a lot of stuff.
I tried to do a pull request on my own [fork,](https://github.com/tmontaigu/CloudCompare/pull/1/files) the diff does not look as if everything changed, but is very hard to follow

asmaloney: I think it's just the number of changes in that file.

FWIW this is just too large for me to review. In the future, a series of smaller, logical, atomic PRs would be much more helpful (i.e. one might be "move enums into header" - that size). Testing stuff could be a completely separate series of PRs.

I can't speak for the others, but for me something that takes 5 minutes is much more likely to happen than something that's going to take me 1+ hours.

I do appreciate the contribution! Just want to give feedback since I believe we need more code reviews.

tmontaigu: Then, maybe I could try do redo the changes in multiple PR's

asmaloney: If you can, I think it would help in terms of understanding what's in this big change.

Maybe start with the "cleanup/style" type things you did in preparation for the guts of your change if you can? These are usually easy to review.

I would start by separating the testing code out to do after, then look at the larger change for the smaller logical pieces - might be "Reformat case statements for readability", "Move enums into header and change into class-based enum", "Move member initialization into header",  or "Document undocumented methods".

If there's anything I can help with, please let me know.

tmontaigu: So I should make the following PRs:
1) General code modernization (enum class, nullptr, etc)
2) Modernization of the file reading (Use QDataStream, split reading functions, fix the 2 bugs I had when reading .shp files)
2) Modernization of the file writing (same idea as above but will introduce more changes)
3) Add Reading & Writing for Multipatch shapes
4) Add tests

asmaloney: I think that sounds great.

When I'm reviewing code, I like to be able to focus on the one thing that it's trying to accomplish. Separating it like this would help a lot.

Thanks!

PR #822, opened 2018-12-23
Use QDataStream to read SHP files

tmontaigu:
The main change is to use QDataStream to read the input shp file, which in my opinion makes the code easier to maintain.

Also adds POLYLINE_M & POLYGON_M as supported shape type

### DIFF
--- a/libs/qCC_io/ShpFilter.cpp
+++ b/libs/qCC_io/ShpFilter.cpp
@@ -56,6 +56,8 @@ using FieldIndexAndName = QPair<int, QString>;
 
 //Specific value for NaN
 static const double ESRI_NO_DATA = -1.0e38;
+static int32_t ESRI_SHAPE_FILE_CODE = 9994;
+const size_t ESRI_HEADER_SIZE = 100;
 
 //semi-persistent settings
 static double s_dbfFielImportScale = 1.0;
@@ -64,6 +66,7 @@ static double s_dbfFielImportScale = 1.0;
 static bool s_save3DPolysAs2D = false;
 static int  s_poly2DVertDim = 2;
 static bool s_save3DPolyHeightInDBF = false;
+
 
 //! ESRI Shapefile's shape types
 enum class ESRI_SHAPE_TYPE : int32_t {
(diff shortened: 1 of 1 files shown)


libs/qCC_io/ShpFilter.cpp
```diff
@@ -551,26 +575,32 @@ CC_FILE_ERROR LoadPolyline(	QFile& file,
 
 		if (!scalarValues.empty())
 		{
-			ccScalarField* sf = new ccScalarField("Measures");
-			if (!sf->reserveSafe(vertCount))
+			bool allNans = std::all_of(
```
asmaloney: I like it!

libs/qCC_io/ShpFilter.cpp
```diff
@@ -816,62 +836,49 @@ CC_FILE_ERROR LoadCloud(QFile& file,
...
-		{
-			file.read(header, 16);
-			double x = qFromLittleEndianD(*reinterpret_cast<double*>(header));
-			double y = qFromLittleEndianD(*reinterpret_cast<double*>(header + 8));
-			CCVector3 P(static_cast<PointCoordinateType>(x + Pshift.x),
-						static_cast<PointCoordinateType>(y + Pshift.y),
-						0);
-			cloud->addPoint(P);
-		}
+		double  x, y;
```
asmaloney: One line per declaration.

libs/qCC_io/ShpFilter.cpp
```diff
@@ -816,62 +836,49 @@ CC_FILE_ERROR LoadCloud(QFile& file,
...
-				file.read(header, 8);
-				double z = qFromLittleEndianD(*reinterpret_cast<double*>(header));
-				const CCVector3* P = cloud->getPoint(i);
-				const_cast<CCVector3*>(P)->z = static_cast<PointCoordinateType>(z + Pshift.z);
-			}
-			cloud->invalidateBoundingBox();
+			double z;
+			shpStream >> z;
+			const CCVector3* P = cloud->getPoint(i);
+			const_cast<CCVector3*>(P)->z = static_cast<PointCoordinateType>(z + Pshift.z);
```
asmaloney: @dgirardeau Is there a better way of doing this rather than this const_casting? To me const casting is usually a sign that something's not right...

libs/qCC_io/ShpFilter.cpp
```diff
@@ -1624,69 +1545,51 @@ CC_FILE_ERROR ShpFilter::loadFile(const QString& filename, ccHObject& container,
...
-			//Byte 4: Content Length
-			int32_t recordSize = qFromBigEndian<int32_t>(*reinterpret_cast<const int32_t*>(header + 4)); //Record numbers begin at 1
-			recordSize *= 2; //recordSize is measured in 16-bit words
-			fileLength -= 8;
-			pos += 8;
+			ccLog::Warning("[SHP] Something went wrong reading the file");
+			return CC_FERR_READING;
+		}
+		int32_t recordNumber, recordSize, shapeTypeInt;
+		shpStream.setByteOrder(QDataStream::BigEndian);
```
asmaloney: Does this format really switch between "endianness"?

tmontaigu: As strange as it may seem, yes the endianness switches throughout the file parts

asmaloney: Yuck.

That seems like an odd design decision :-)

libs/qCC_io/ShpFilter.cpp
```diff
@@ -154,6 +157,66 @@ static inline bool hasMeasurements(ESRI_SHAPE_TYPE shapeType)
...
+
+CC_FILE_ERROR ShapeFileHeader::readFrom(QDataStream &sin)
+{
+	sin.setByteOrder(QDataStream::BigEndian);
+
+	int32_t fileCode;
+	sin >> fileCode;
+	if (fileCode != ESRI_SHAPE_FILE_CODE)
+	{
+		ccLog::Warning("[SHP] wrong file code (%d), is this a shape file?", fileCode);
```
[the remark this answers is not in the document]
tmontaigu: They could be translated, but then all the different warnings emmited in by the ShpFilter should be translated

libs/qCC_io/ShpFilter.cpp
dgirardeau: The ccPointCloud class has the 'point' method that returns a non-const pointer on a point.

asmaloney: OK, thank you. I assumed it was there somewhere, but I couldn't find it because I was looking for a non-const getPoint().

tmontaigu: The 'point' function is protected, should it be made public ?

dgirardeau: Ah! Sorry I forgot that. Actually until recently it was on purpose that point couldn't be changed easily (because it invalidates the bounding box, the octree, etc.).

Then I think that we should either add a dediated method (like 'setPoint(index, P)') ... or keep the actual code as it clearly shows that this is something "dangerous" ;)

tmontaigu: Another solution could be to read the xs, ys , zs(if any) into a std::vector and building the cloud after instead of doing it in place like it is done now

dgirardeau: It could work if there are not too many points. But frankly, I think it's perfectly ok in this case to keep the const_cast and don't bother that much ;)

PR #825, opened 2018-12-28
ShpFilter: Write using QDataStream

tmontaigu:
It should be the biggest PR of the series,

The changes are:
 - use QDataStream to write shp files
 - some fixes to make sure that when reading ScalarFields are not added to the cloud if all values are nan.
 - Closed polyline are saved as polygon (there is no option that the user can change atm)
 - Polylines/Polygons can be saved with/without Z and/or Measure scalarfield

### DIFF
--- a/libs/qCC_io/ShpFilter.cpp
+++ b/libs/qCC_io/ShpFilter.cpp
@@ -59,6 +59,7 @@ static const double ESRI_NO_DATA = -1.0e38;
 
 static const int32_t ESRI_SHAPE_FILE_CODE = 9994;
 static const size_t ESRI_HEADER_SIZE = 100;
+static const size_t ESRI_FILE_LENGTH_OFFSET = 24;
 
 //semi-persistent settings
 static double s_dbfFielImportScale = 1.0;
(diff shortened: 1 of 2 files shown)


libs/qCC_io/ShpFilter.cpp
```diff
@@ -164,6 +165,201 @@ static inline bool isESRINoData(double m)
...
+	}
+	return minMax;
+}
+
+static bool bBoxOfHObjectContainer(const ccHObject::Container& objects, CCVector3d &bbMinCorner, CCVector3d bbMaxCorner)
+{
+	bool isValid = false;
+	for (ccHObject *obj : objects)
+	{
+		CCVector3d minC, maxC;
```
asmaloney: Prefer one line per declaration.

libs/qCC_io/ShpFilter.cpp
```diff
@@ -623,62 +824,63 @@ CC_FILE_ERROR SavePolyline(ccPolyline* poly, QFile& file, int32_t& bytesWritten,
...
+
+	recordSize = sizeofPolyLine(outputShapeType, numPoints, numParts);
+	out.setByteOrder(QDataStream::BigEndian);
+	out << recordNumber << recordSize;
+
+	qint64 recordStart = out.device()->pos();
+	out.setByteOrder(QDataStream::LittleEndian);
+	out << static_cast<int32_t>(outputShapeType);
+
+	CCVector3d bbMing, bbMaxg;
```
asmaloney: Prefer decls on separate lines.

libs/qCC_io/ShpFilter.cpp
```diff
@@ -893,10 +1044,18 @@ static CC_FILE_ERROR LoadCloud(QDataStream &shpStream,
 				ScalarType s = isESRINoData(m) ? NAN_VALUE : static_cast<ScalarType>(m);
 				sf->addElement(s);
 			}
-			sf->computeMinAndMax();
-			int sfIdx = cloud->addScalarField(sf);
-			cloud->setCurrentDisplayedScalarField(sfIdx);
-			cloud->showSF(true);
+			bool allNans = std::all_of(sf->begin(), sf->end(), [](ScalarType s) { return std::isnan(s); });
```
asmaloney: Would this be useful to add to ccScalarField?

(Not suggesting it for this PR - just asking.)

libs/qCC_io/ShpFilter.cpp
```diff
@@ -907,130 +1066,73 @@ static CC_FILE_ERROR LoadCloud(QDataStream &shpStream,
...
+	out << recordNumber << recordSize;
 
-	//Byte 36: NumPoints (The total number of points)
-	int32_t numPoints = static_cast<int32_t>(cloud->size());
-	{
-		int32_t numPointsLE = qToLittleEndian<int32_t>(numPoints);
-		file.write((const char*)&numPointsLE, 4);
-		bytesWritten += 4;
-	}
+	CCVector3d bbMing, bbMaxg;
```
asmaloney: Prefer decls on separate lines.

libs/qCC_io/ShpFilter.cpp
tmontaigu: Maybe, the check has to be made because CC doesn’t like scalar fields full of NaNs, so we  need to make sure the scalar field contains a valid value before attaching it to an object

dgirardeau: Or we could add a more generic version that takes a scalar value as input? (but I agree, not in this PR)

PR #827, opened 2018-12-31
Fix Linux bug with the shader path

rjanvier:
fix from @asmaloney. At first glance it seems to do the job.
Please merge it if it's harmless for other OSes.


rjanvier:
thanks @asmaloney

dgirardeau:
Thanks to both of you ;)

PR #830, opened 2019-01-03
[SHP] Add Unit tests

tmontaigu:
The last PR

Adds some unit tests for the loading & saving of .shp files.
A new cmake option `BUILD_TESTING` is added (`ON` by default) to activate the building of the tests


asmaloney:
I'm guessing that BUILD_TESTING is in CTest and that's what turns it on by default?

My (not-so-strong) preference would be to have it OFF by default - especially if we start adding other tests - though I can understand the arguments for having it default to ON.

tmontaigu:
Yes CTest adds the BUILD_TESTING and turns it ON  by default,
according to this [stackoverflow](https://stackoverflow.com/questions/41832908/set-build-testing-to-off-by-default-using-ctest) it should be possible to make it OFF by default

asmaloney:
Travis doesn't seem to be running the tests:

https://travis-ci.org/CloudCompare/CloudCompare/jobs/475093954#L3043

tmontaigu:
I'll need to figure out why 🤔

asmaloney:
Tests are failing on clang because of this: https://github.com/travis-ci/travis-ci/issues/8613

.travis.yml
```diff
@@ -45,9 +46,11 @@ before_script:
   -DINSTALL_QPOISSON_RECON_PLUGIN=ON
   -DINSTALL_QSRA_PLUGIN=ON
   -DINSTALL_QSSAO_PLUGIN=ON
+  -DBUILD_TESTING=ON
   ..
 script:
 - if [[ $TRAVIS_BRANCH != "beta_builds" ]]; then make -j2; fi
```
asmaloney: To fix the clang test issue (https://github.com/travis-ci/travis-ci/issues/8613), add the following line before running ctest:

`- if [[ $CXX == "clang++" ]]; then export LD_LIBRARY_PATH=/usr/local/clang/lib:$LD_LIBRARY_PATH; fi`

asmaloney:
Thanks again Thomas!

PR #832, opened 2019-01-08
Cleanup ccColorType

asmaloney:
General cleanup:
- "using" instead of "typedef"
- use "constexpr" in place of "extern const"s
- pass some args by const ref
- reformat conversion functions for readability

`ccColorType.h` is indirectly included in about *80%* of CloudCompare's cpp files, so the fewer includes the better for compilation speed.

- move Generator::Random() definition into cpp
	- `<random>` is very heavy (a dozen includes and 6700 lines of templates in llvm version), so get `<random>` out of the header
	- it is not used very frequently in the codebase
	- it is unlikely to be inlined by the compiler anyways
- move Convert::hsv2rgb() definition into cpp
	- this method is unused and this allows us to move `<cmath>` into the cpp
- because the headers were moved, a couple of other files required new includes
	- there may be other changes required for Windows

### DIFF
--- a/common/ccApplicationBase.cpp
+++ b/common/ccApplicationBase.cpp
@@ -15,6 +15,8 @@
 //#                                                                        #
 //##########################################################################
 
+#include <clocale>
+
 //Qt
 #include <QDir>
 #include <QStandardPaths>
--- a/libs/qCC_db/ccColorTypes.cpp
+++ b/libs/qCC_db/ccColorTypes.cpp
@@ -15,39 +15,72 @@
 //#                                                                        #
 //##########################################################################
 
+#include <cmath>
+#include <random>
+
 #include "ccColorTypes.h"
+
 
 namespace ccColor
 {
-	// Predefined colors (default type)
-	QCC_DB_LIB_API const Rgb white						(MAX, MAX, MAX);
-	QCC_DB_LIB_API const Rgb lightGrey					(static_cast<ColorCompType>(MAX*0.8), static_cast<ColorCompType>(MAX*0.8), static_cast<ColorCompType>(MAX*0.8));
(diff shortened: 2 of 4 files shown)


bartoszek:
There're a few places where using `constexpr` results in an error in gcc-8.2.1@archlinux-4.20.
```
/home/bartus/AUR/cloudcompare-git/src/cloudcompare/libs/qCC_db/ccColorTypes.h:122:61: error: ‘(2.55e+2 * 8.0000000000000004e-1)’ is not a constant expression
  constexpr Rgb lightGrey     (static_cast<ColorCompType>(MAX*0.8), static_cast<ColorCompType>(MAX*0.8), static_cast<ColorCompType>(MAX*0.8));
```
```
In file included from /build/cloudcompare-git/src/cloudcompare/CC/src/CCMiscTools.cpp:22:
/build/cloudcompare-git/src/cloudcompare/CC/include/CCConst.h:43:40: error: ‘(1.8e+2 / 3.1415926535897931e+0)’ is not a constant expression
 constexpr double CC_RAD_TO_DEG = (180.0/M_PI);
                                  ~~~~~~^~~~~~
/build/cloudcompare-git/src/cloudcompare/CC/include/CCConst.h:46:39: error: ‘(3.1415926535897931e+0 / 1.8e+2)’ is not a constant expression
 constexpr double CC_DEG_TO_RAD = (M_PI/180.0);
```
the rest works fine.

asmaloney:
I can't see the problem either.

I don't know how you can get more "const" than `(2.55e+2 * 8.0000000000000004e-1)` or `(1.8e+2 / 3.1415926535897931e+0)`.

bartoszek:
Me either. It looks like even `constexpr double CC_RAD_TO_DEG = 180.0/3.14;` throws an error - realy strange...

Issue #837, opened 2019-01-13
qPCV in command line


foobarbecue:
It would be lovely if we could do ambient occlusion from the command line, using the qPCV plugin!

foobarbecue:
I started working on this, and then realized it introduces a dependency between the command line parser and a plugin, which might not be installed... is there a policy on this? I guess we need some preprocecessor stuff in ccCommandLineCommands.h to check what plugins are available?

foobarbecue:
I guess I can follow the pattern used by BUNLER_IMPORT?

dgirardeau:
If you want to do it, then you have to use the new 'registerCommands' mechanism for plugins.

See how the M3C2 plugin implements it for instance (qM3C2Plugin::registerCommands):
https://github.com/CloudCompare/CloudCompare/blob/master/plugins/core/qM3C2/qM3C2.cpp#L104

To summarize, you have to create a ccCommandLineInterface::Command object that does the work, and register it to the main command line mechanism (this is automatically call on startup when loading the plugin).

foobarbecue:
Thanks! See PR.

PR #845, opened 2019-02-17
Add -PCV option to command line interface for ambient occlusion

foobarbecue:
Addresses #837 . Good enough for my personal use right now. It doesn't expose options like number of rays yet, and only works for a single mesh (qPCV supports arbitrary number of meshes and point clouds). I could add those features sometime soon.

### DIFF
--- a/plugins/core/qPCV/PCVCommand.cpp
+++ b/plugins/core/qPCV/PCVCommand.cpp
@@ -8,11 +8,11 @@
 #include <ccPointCloud.h>
 #include <ccProgressDialog.h>
 
-static const char COMMAND_PCV[] = "PCV";
-static const char COMMAND_PCV_N_RAYS[] = "N_RAYS";
-static const char COMMAND_PCV_IS_CLOSED[] = "IS_CLOSED";
-static const char COMMAND_PCV_180[] = "180";
-static const char COMMAND_PCV_RESOLUTION[] = "RESOLUTION";
+constexpr char COMMAND_PCV[] = "PCV";
+constexpr char COMMAND_PCV_N_RAYS[] = "N_RAYS";
+constexpr char COMMAND_PCV_IS_CLOSED[] = "IS_CLOSED";
+constexpr char COMMAND_PCV_180[] = "180";
+constexpr char COMMAND_PCV_RESOLUTION[] = "RESOLUTION";
 
 PCVCommand::PCVCommand() :
 	Command("PCV", COMMAND_PCV)
--- a/plugins/core/qPCV/PCVCommand.cpp
+++ b/plugins/core/qPCV/PCVCommand.cpp
@@ -77,7 +77,8 @@ bool PCVCommand::process(ccCommandLineInterface& cmd)
 			qPCV::tr("No mesh is available. The CLI PCV implementation currently only supports a single mesh."));
 	}
 
-	ccPointCloud* pc = ccHObjectCaster::ToPointCloud(cmd.meshes()[0].mesh->getAssociatedCloud());
+	ccGenericPointCloud* meshCloud = cmd.meshes()[0].mesh->getAssociatedCloud();
+	ccPointCloud* pc = ccHObjectCaster::ToPointCloud(meshCloud);
 	ccGenericMesh* mesh = cmd.meshes()[0].mesh;
 
 	int point_count = pc->size();
--- a/plugins/core/qPCV/PCVCommand.cpp
+++ b/plugins/core/qPCV/PCVCommand.cpp
@@ -9,6 +9,10 @@
 #include <ccProgressDialog.h>
 
 static const char COMMAND_PCV[] = "PCV";
+static const char COMMAND_PCV_N_RAYS[] = "N_RAYS";
+static const char COMMAND_PCV_IS_CLOSED[] = "IS_CLOSED";
+static const char COMMAND_PCV_180[] = "180";
+static const char COMMAND_PCV_RESOLUTION[] = "RESOLUTION";
 
 PCVCommand::PCVCommand() :
 	Command("PCV", COMMAND_PCV)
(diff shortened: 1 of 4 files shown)


plugins/core/qPCV/PCVCommand.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+	cmd.print(QObject::tr("Size is %1").arg(point_count));
+
+	ccProgressDialog pcvProgressCb(true);
+	pcvProgressCb.setAutoClose(false);
+
+	PCV::Launch(n_rays, pc, mesh, meshIsClosed, mode360, resolution, resolution,
+		&pcvProgressCb, QString(mesh->getName()));
+
+	cmd.meshes()[0].basename += QObject::tr("_PCV");
```
asmaloney: No need to add translation for this.

plugins/core/qPCV/PCVCommand.h
```diff
@@ -0,0 +1,35 @@
...
+//##########################################################################
+
+#include "ccCommandLineInterface.h"
+
+class PCVCommand : public ccCommandLineInterface::Command
+{
+public:
+	PCVCommand();
+
+	~PCVCommand()
```
asmaloney: This should be `~PCVCommand() override = default;`

plugins/core/qPCV/PCVCommand.cpp
```diff
@@ -0,0 +1,53 @@
...
+static const char COMMAND_PCV[] = "PCV";
+
+PCVCommand::PCVCommand() :
+	Command("PCV", COMMAND_PCV)
+{
+}
+
+bool PCVCommand::process(ccCommandLineInterface& cmd)
+{
+	static int n_rays = 256;
```
asmaloney: You didn't address this. Why are these `static`? As far as I can tell they should be `const` or `constexpr`.

foobarbecue:
@asmaloney , thanks for your patience & careful code reviews! I'm pretty new to C++, as you can probably tell. I changed the destructor declaration and got rid of unnecessary "static" in 3e53c2bd235fc5b726acd8249df9d1475e2b1b20 .

foobarbecue:
What about docs to go with this? Should I update https://www.cloudcompare.org/doc/wiki/index.php?title=Command_line_mode ?

asmaloney:
I'll have to let @dgirardeau answer the question about docs.

As for your latest commit, normally when you've asked for a PR, you're "done" and ready to merge - you don't make changes to add new things like this. It means the whole PR needs review again.

plugins/core/qPCV/PCVCommand.cpp
```diff
@@ -0,0 +1,101 @@
...
+#include <ccGenericMesh.h>
+#include <ccHObjectCaster.h>
+#include <ccPointCloud.h>
+#include <ccProgressDialog.h>
+
+static const char COMMAND_PCV[] = "PCV";
+static const char COMMAND_PCV_N_RAYS[] = "N_RAYS";
+static const char COMMAND_PCV_IS_CLOSED[] = "IS_CLOSED";
+static const char COMMAND_PCV_180[] = "180";
+static const char COMMAND_PCV_RESOLUTION[] = "RESOLUTION";
```
asmaloney: In C++11, these are normally `constexpr`. So:

```cpp
constexpr char COMMAND_PCV_RESOLUTION[] = "RESOLUTION";
```

plugins/core/qPCV/PCVCommand.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+	// Initialize to match PCV::Launch defaults
+	unsigned nRays = 256;
+	bool meshIsClosed = false;
+	bool mode360 = true;
+	unsigned resolution = 1024;
+
+	while (!cmd.arguments().empty())
+	{
+		QString arg = cmd.arguments().front();
```
asmaloney: This should be `const` and you should use [takeFirst](https://doc.qt.io/qt-5/qlist.html#takeFirst):

```cpp
const QString arg = cmd.arguments().takeFirst();
```

Then remove the calls to `pop_front()` below.

plugins/core/qPCV/PCVCommand.cpp
```diff
@@ -0,0 +1,101 @@
...
+			mode360 = false;
+		}
+		else if (ccCommandLineInterface::IsCommand(arg, COMMAND_PCV_N_RAYS))
+		{
+			cmd.arguments().pop_front();
+			bool conversionOk = false;
+			nRays = cmd.arguments().takeFirst().toFloat(&conversionOk);
+			if (!conversionOk)
+			{
+				return cmd.error(QObject::tr("Invalid parameter: value after \"-%1\"").arg(COMMAND_PCV_N_RAYS));
```
asmaloney: Where you are using `QObject::tr()`, please use `qPCV::tr()` so the translations are grouped properly for the translators.

plugins/core/qPCV/PCVCommand.cpp
```diff
@@ -0,0 +1,101 @@
...
+		}
+	}
+
+	if (cmd.meshes().empty())
+	{
+		return cmd.error(
+			QObject::tr("No mesh is available. The CLI PCV implementation currently only supports a single mesh."));
+	}
+
+	ccPointCloud* pc = ccHObjectCaster::ToPointCloud(cmd.meshes()[0].mesh->getAssociatedCloud());
```
asmaloney: This should be split for readability:
```cpp
ccGenericPointCloud *meshCloud = cmd.meshes()[0].mesh->getAssociatedCloud();
ccPointCloud* pc = ccHObjectCaster::ToPointCloud( meshCloud );
```

plugins/core/qPCV/PCVCommand.cpp
```diff
@@ -0,0 +1,101 @@
...
+	ccGenericMesh* mesh = cmd.meshes()[0].mesh;
+
+	int point_count = pc->size();
+
+	cmd.print(QString("Size is %1").arg(point_count));
```
asmaloney: If this is a user-facing string, it should be translated with `qPCV::tr()`. I don't know how much of the command-line stuff is translated (I'm guessing not much?) but we have to start somewhere :-)

plugins/core/qPCV/PCVCommand.cpp
foobarbecue: Addressed in d0da40e306c750c901df52c65bdc1aecdc79e88e

plugins/core/qPCV/PCVCommand.cpp
foobarbecue: Addressed in df3b2f59dada2391d69afcacf88e0154701c5061

plugins/core/qPCV/PCVCommand.cpp
foobarbecue: Addressed in 8d2738dbc565bda4581428028cdbc7c3b818d624

plugins/core/qPCV/PCVCommand.cpp
foobarbecue: Addressed in df3b2f5

plugins/core/qPCV/PCVCommand.cpp
foobarbecue: Re-addressed in 545a8db082f63efe2d72f32d86b4b1640babaffe -- I realized two of these strings are not qPCV specific -- they're used in other CLI commands. One of them (an implementation detail) is. There was a fourth string, which reported the number of points in the point cloud, but the progress bar window also displays that, so I removed it since it seems like unnecessary duplication.

plugins/core/qPCV/PCVCommand.cpp
foobarbecue: We don't want to remove the argument from the list until we've verified it's an option of PCV. Otherwise, we'll eat up the next command. I think we should leave this as is, unless you think we should take the argument out of the list and then put it back in if it doesn't match any of the PCV options.

foobarbecue: I'm double-checking this now and will add the const.

foobarbecue: Addressed in 3355f5c188c0084476d9daff63504bd7214c527c

foobarbecue:
> As for your latest commit, normally when you've asked for a PR, you're "done" and ready to merge - you don't make changes to add new things like this. It means the whole PR needs review again.

Yeah, sorry! I had a change of heart and felt this thing shouldn't be in master without the options implemented. I'll think more about defining the scope of my PRs before creating them in the future.

plugins/core/qPCV/PCVCommand.h
foobarbecue: Addressed in 3e53c2

plugins/core/qPCV/PCVCommand.cpp
foobarbecue: They should never have been static. They can't be const now that I've added reading options from the command line.

plugins/core/qPCV/PCVCommand.cpp
asmaloney: Ah - you're absolutely correct. I misunderstood the way this is implemented & thought it was only receiving its own commands.

Based on that - one last thing from me :-)

```cpp
const QString arg = cmd.arguments().front();
```
should actually use a reference so it isn't copying each string:
```cpp
const QString& arg = cmd.arguments().front();
```

foobarbecue: Addressed in 2838528cc720969a1e553c04f46ee471cd44a2cd

asmaloney:
Thanks for working through that with me Aaron!

I'll leave it now for @dgirardeau in case he has more input.

dgirardeau:
Okay good! And thanks for the contribution!

I'll probably add the support for any number of clouds AND meshes (not sure why you restricted that to clouds only ;). But now it should be easy.

dgirardeau:
@foobarbecue Regarding the command line documentation on the wiki, can you create an account on the wiki? I think I'll have to validate it then and you'll be able to add the description for this new option.

foobarbecue:
@dgirardeau Sure, I'd like to sign up, but I couldn't find a "create account" link. When I tried going to Special:CreateAccount (where I think it usually is), I got:  
image

dgirardeau:
My bad, I must have restricted it! (it's crazy as a small forum and a small wiki as those of CloudCompare seem to be of interest for so many spammers :| ).

You should receive a mail soon ;)

foobarbecue:
Cool, I added a row to the table. Wasn't sure if it's in any kind of order so I just put it at the bottom. https://www.cloudcompare.org/doc/wiki/index.php?title=Command_line_mode&type=revision&diff=47189&oldid=47188

As for why it only supports a single mesh right now it's just selfish / lazy -- that's the format of the data I wanted to use it on and I'm not really familiar enough with cc types yet to be confident that I can write it to cover all the edge cases. If you want to extend it that would be cool, otherwise I'll get around to it eventually.

PR #852, opened 2019-02-22
Change the QMessagebox on MainWindow

fabriciolelis:
Update the mainwindow class to enable the translation of the windows that appear on the main screen. For example, when leaving the application.


dgirardeau:
Thanks!

PR #853, opened 2019-02-24
Create CloudCompare_ja.ts

YoichiSeino:
We, CCCP (Japanese translation team), translated a part of the UIs (not all of the UIs) of CloudCompare to japanese. Please check our translation file. Thank you.


asmaloney:
Thank you very much for this contribution!

PR #856, opened 2019-02-26
These command line options take ints, not floats

asmaloney:
Avoid extra conversion and potential rounding errors


asmaloney:
@foobarbecue Can you confirm these arguments are supposed to be ints, not floats?

dgirardeau:
Good catch! (I confirm ;) ).

PR #863, opened 2019-03-12
escape question marks to fix accidental trigraph

daniel-lawrence-lu:
Currently the literal `??-` gets interpreted as a [trigraph](https://en.wikipedia.org/wiki/Digraphs_and_trigraphs#C). My compiler `g++ (GCC) 8.2.1 20181127` emits a warning for this line.

Hence it has to be [escaped](https://stackoverflow.com/questions/1586834/escape-sequence-for-in-c), which fixes the warning and avoids `arch` being erroneously set to `~bit` rather than `??-bit`.


asmaloney:
Thank you Daniel!

PR #866, opened 2019-03-16
Small plugin reorganization

asmaloney:
- split into GL, IO, and Standard directories
- change cmake options so they all begin with PLUGIN_
- replace unused SAMPLE_PROJECT_VERSION_MAJOR/MINOR with proper VERSION on project()

(You will need to update your cmake config to turn on the plugins you want again.)


plugins/core/Standard/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
+
+add_subdirectory( ${CMAKE_CURRENT_SOURCE_DIR}/qAnimation )
```
dgirardeau: do you really need to add ```${CMAKE_CURRENT_SOURCE_DIR}/```?

asmaloney: Nope! I'd copy-pasted all of those from the original file - not sure why they had that. Probably my misunderstanding cmake when I did that a couple of years ago.

PR #872, opened 2019-03-19
Convert QString to std::string in utf8 for PDAL

### LINKED ISSUE #869: An error occurred when loading las file whose path contains multi-byte character [v2.11 alpha]
## Current Behavior

 When trying to open a las file with multi-byte character in path, CC shows an Error dialog "PDAL exception 'bad conversion' "
## Your Environment
* CC Version: 2.11 alpha
* OS & Version: Windows 10
* Graphics card: nVidia GeForce GTX 1060 6GB
* Qt Version (if compiling): 

Character Set: Mulitbyte

tmontaigu:
Convert QString into utf8 encoded std::string before feeding them to  PDAL.

This fixes #869 (on windows at least)

### DIFF
--- a/libs/qCC_io/LASFilter.cpp
+++ b/libs/qCC_io/LASFilter.cpp
@@ -471,8 +471,7 @@ CC_FILE_ERROR LASFilter::saveToFile(ccHObject* entity, const QString& filename, 
 	writerOptions.add("scale_y", lasScale.y);
 	writerOptions.add("scale_z", lasScale.z);
 
-	std::string utf8Filename = filename.toUtf8().toStdString();
-	writerOptions.add("filename", utf8Filename);
+	writerOptions.add("filename", filename.toStdString());
 	writerOptions.add("extra_dims", "all");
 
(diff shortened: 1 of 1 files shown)


libs/qCC_io/LASFilter.cpp
```diff
@@ -471,7 +471,8 @@ CC_FILE_ERROR LASFilter::saveToFile(ccHObject* entity, const QString& filename,
 	writerOptions.add("scale_y", lasScale.y);
 	writerOptions.add("scale_z", lasScale.z);
 
-	writerOptions.add("filename", filename.toLocal8Bit().toStdString());
+	std::string utf8Filename = filename.toUtf8().toStdString();
```
asmaloney: You shouldn't need the `toUtf8()` part.

According to [the docs](https://doc.qt.io/qt-5/qstring.html#toStdString):

> Returns a std::string object with the data contained in this QString. The Unicode data is converted into 8-bit characters using the toUtf8() function.

PR #873, opened 2019-03-20
{plugin} Create qCoreIO plugin and move some file formats into it

asmaloney:
Included the following:

   - HeightProfileFilter
   - MAFilter
   - MascaretFilter
   - ObjFilter
   - OFFFilter
   - PDMSFilter
   - PTXFilter
   - SimpleBinFilter
   - STLFilter
   - VTKFilter

None of these require extra third party libraries.

The remaining formats either require additional libraries or they are tightly coupled to the main source and will be handled separately.

Also made some style fixes to the files that were moved.


asmaloney:
**Note:** I plan to address the order of items in the dropdown in the file dialog in a future commit.

PR #874, opened 2019-03-21
If LAS minor_version was 0, let PDAL choose new version when writing

tmontaigu:
CloudCompare remembers the minor_version of a LAS file read to save the file in the same version.

Howerver a problem arise when reading (then saving)  a LAS 1.0 file ([see CC's forum](https://www.danielgm.net/cc/forum/viewtopic.php?f=9&t=3535&sid=2e1d57a48c982d06f4203a6718f04633)) as PDAL can read LAS 1.0 but not write it. (https://github.com/PDAL/PDAL/issues/2196)

In that case let PDAL use its default for the minor_version


dgirardeau:
Thanks for the quick fix!

PR #876, opened 2019-03-21
Move E57 to its own plugin

asmaloney:
(Also add E57 to the travis build)

One small difference to other plugins: required cmake is 3.1 (as opposed to 3.0). This is because that's when they added the `FindXercesC` module.

Note that if XercesC isn't found, the XercesC options show up in the "advanced" cmake options. (This is just the way they did it in the FindXercesC module for some reason.)

I hope this XercesC stuff works on Windows because it really cleans up the cmake code...


dgirardeau:
The FindXercesC module works on my side this time!

dgirardeau:
Wait I still miss my XERCES_STATIC_LIBRARY definition :p

dgirardeau:
I'll fix that after merging

PR #877, opened 2019-03-22
Move FBX to its own plugin

asmaloney:
1. I cannot compile this since I don't have the FBX SDK, so it's a "best effort" until Daniel can verify/fix it.

2. I had to move the command line option into the plugin. I (think I) changed it from `-FBX_EXPORT_FMT` to `-FBX -EXPORT-FMT` to be more consistent with plugin command line options. This is noted in the changelog.

3. The git changes below might be a little confusing since it mapped _contrib/FBXSupport.cmake_ → _plugins/core/IO/qFBXIO/CMakeLists.txt_


dgirardeau:
Okay works for me now

PR #882, opened 2019-03-25
{plugins} Simplify the I/O plugin API 

asmaloney:
- The `FileIOFilter` base class now takes a struct as an argument containing all the static info about a filter - extensions, features (import/export), etc.. See `FooFilter` in the `ExampleIOPlugin` and the comments in `FileIOFilter::FilterInfo`.
- The use of `FileIOFilter::FilterInfo` means that the following virtual functions in I/O filters are no longer virtual/required:
   - importSupported
   - exportSupported
   - getFileFilters
   - getDefaultExtension
   - canLoadExtension
- Removes a lot of boilerplate code and makes it easier to add properties/features to I/O plugins
- Because it changes every single I/O plugin, this PR is unavoidably large...

### DIFF
--- a/libs/qCC_io/AsciiFilter.cpp
+++ b/libs/qCC_io/AsciiFilter.cpp
@@ -70,14 +70,17 @@ AsciiOpenDlg* AsciiFilter::GetOpenDialog(QWidget* parentWidget/*=0*/)
 	return s_openDialog;
 }
 
-bool AsciiFilter::canLoadExtension(const QString& upperCaseExt) const
+
+AsciiFilter::AsciiFilter() :
+	FileIOFilter( {
+				  "_ASCII Filter",
+				  QStringList{ "txt", "asc", "neu", "xyz", "pts", "csv" },
+				  "asc",
+				  QStringList{ "ASCII cloud (*.txt *.asc *.neu *.xyz *.pts *.csv)" },
+				  QStringList{ "ASCII cloud (*.txt *.asc *.neu *.xyz *.pts *.csv)" },
+				  Import | Export
+				  } )
 {
-	return (	upperCaseExt == "ASC"
-			||	upperCaseExt == "TXT"
-			||	upperCaseExt == "XYZ"
-			||	upperCaseExt == "NEU"
-			||	upperCaseExt == "PTS"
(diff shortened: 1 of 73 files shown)


libs/qCC_io/AsciiFilter.cpp
```diff
@@ -70,14 +70,17 @@ AsciiOpenDlg* AsciiFilter::GetOpenDialog(QWidget* parentWidget/*=0*/)
 	return s_openDialog;
 }
 
-bool AsciiFilter::canLoadExtension(const QString& upperCaseExt) const
+
+AsciiFilter::AsciiFilter() :
+	FileIOFilter( {
+				  "_ASCII Filter",
```
dgirardeau: Is there a reason for the underscore before 'ASCII'?

libs/qCC_io/FileIOFilter.h
```diff
@@ -293,6 +296,65 @@ class FileIOFilter
...
+		
+		DynamicInfo = 0x0008,	//< FilterInfo cannot be set statically (this is used for internal consistency checking)
+	};
+	Q_DECLARE_FLAGS( FilterFeatures, FilterFeature )
+	
+protected:
+	struct FilterInfo
+	{
+		//! ID used to uniquely identify the filter (not user-visible)
+		QString mID;
```
dgirardeau: We don't have this 'm' convention in CloudCompare for members, so it would be better to not mix conventions ;)

libs/qCC_io/AsciiFilter.cpp
dgirardeau: Oh, it seems it depends on whether the filter is defined in qCC_IO or plugins? Isn't it redundant with the 'FromPlugin' flag?

asmaloney: > Isn't it redundant with the 'FromPlugin' flag? 

I was originally going to encode this info in the name - "_"  for core and "+" for core plugin - but making it explicit is better.

> Is there a reason for the underscore before 'ASCII'? 

The ID is supposed to be unique, so I was going to "reserve" "_" for internal (core) IDs. I will likely change the plugins to underscore once I start using the IDs.

PR #890, opened 2019-03-30
Remove problematic disconnect in ccSectionExtractionTool

asmaloney:
In _ccSectionExtractionTool::linkWith()_:

1) If the parameter "win" is nullptr, then _ccOverlayDialog::linkWith()_ will set m_associatedWin to nullptr which we dereference

2) If it isn't, then we are disconnecting "win" because m_associatedWin is assigned it (or equal to it) in _ccOverlayDialog::linkWith()_

3) It should probably have read `oldWin->disconnect( this )`, but this case is handled in _ccOverlayDialog::linkWith()_


dgirardeau:
Good catch

Issue #883, opened 2019-03-26
{cmake} Consider using GENERATE_EXPORT_HEADER


asmaloney:
**Describe the feature you would like**

For each of our libraries, we currently use some header magic for Windows DLL export/import.

GCC/clang have an equivalent - setting the [visibility attribute](https://gcc.gnu.org/wiki/Visibility).

Using this can (among other things):

- produce smaller libs
- allow better optimization within the libraries
- allow the libraries to load faster

Cmake provides [GENERATE_EXPORT_HEADER](https://cmake.org/cmake/help/v3.0/module/GenerateExportHeader.html) which automatically generates header files (e.g. qCC_io.h) with the right #ifdefs to handle this cross-platform.

**Proposal**

I suggest we migrate to using cmake's `GENERATE_EXPORT_HEADER` to replace our own headers.

**Additional context**

I _think_ it's pretty straightforward to switch - we just need to add the cmake code for `GENERATE_EXPORT_HEADER` from the docs. We may need to change some macro names or include file names as well.

As a test, I experimented with the `visibility` attribute by hand with `libQCC_IO_LIBd.dylib`. The resulting lib is slightly smaller (0.6%! - but it indicates there's some optimizing going on) and the number of exported symbols was reduced from 1527 to 625.

asmaloney:
Another option for our Qt-based libraries - maybe even easier - would be to use [Qt's macros](https://doc.qt.io/qt-5/sharedlibrary.html). I use this in my own work. The header would look something like this:

```c++
#include <QtCore/QtGlobal>

#if defined( FOO_LIBRARY_BUILD )
#  define FOO_EXPORT Q_DECL_EXPORT
#else
#  define FOO_EXPORT Q_DECL_IMPORT
#endif
```

Then I think we'd just need to set this in cmake:
```cmake
set(CMAKE_CXX_VISIBILITY_PRESET hidden)
```

PR #891, opened 2019-03-30
Use Qt's macros for DLL import/export in qCC_db

asmaloney:
This implements the 2nd strategy from issue #883 - using Qt's macros.

1) Allows macOS and Linux to use visibility attributes (some benefits listed in the issue)
2) Reduces OS special cases in cmake and simplifies the header file
3) I needed to de-inline the NormsIndexesTableType constructor and mark it for export to link properly on macOS

Once this works properly on Windows & Linux, the same method should be applied to our other shared libs.


dgirardeau:
Ok it seems to work with Visual Studio!

Do you plan to make other changes for other libs?

asmaloney:
Yes - I have IO done and ready to go which I will do after this is merged.

For the "CC" lib I will have to use strategy 1 from #883 (use cmake's stuff) because Qt is optional there. So I will do another PR for that because it will be quite different.

PR #893, opened 2019-03-30
Use cmake to generate DLL import/export header for CCLib

### LINKED ISSUE #883: {cmake} Consider using GENERATE_EXPORT_HEADER

asmaloney:
This implements the 1st strategy from issue #883 - using cmake's generate_export_header.

Note: I needed to use `CC_CORE_LIB_API` on the entire ScalarField class instead of just specific functions for linking on macOS.

When this works on Windows, it will fix #883.


dgirardeau:
What happens if someone want to include/use CC_CORE_LIB in another project / without CMake?

asmaloney:
I'm not sure I understand the question completely.

Libraries that are intended to be built & used by others provide a way to build them (unless they are header-only) and they specify what build tool is needed. These days that's normally cmake which is what we use...

Is this different somehow?

dgirardeau:
As CC_CORE_LIB is LGPL, it is used by other projects that don't use CMake. Therefore one need to compile CC_CORE_LIB (indeed with CMake at this time), then copy the headers AND the 'cc_core_lib_export.h' file (which is generated in the build directory of CMake) and the libraries/DLLs.

It's almost the same as before, but they need to manage this 'cc_core_lib_export.h' file in addition (in order to 'regroup' it with the other header files).

asmaloney:
Based on that, the correct solution is to `install` that file with the headers, but it doesn't look like any of the header files are installed properly for the case you describe. In fact the cmake file is kind of CC-specific.

Shouldn't this library be its own repo if it's intended to be built and used separately?

dgirardeau:
It would be a solution indeed. But for now, users just need to use CMake directly on the CC_CORE_LIB CMake script, compile it and 'grab the libraries' (it works)

asmaloney:
Is it unrealistic to ask them to copy one more file until we can put this in its own repo and do it properly?

How many people do you think are using this as a separate lib?

What if I generated the file in CC/include instead and add a .gitignore for it?

dgirardeau:
I like the last idea (generating it in the CC/include folder (it solves my main issue).

I'm not sure how much people are using it like that, but I know at least 4 separate projects that do it "seriously".

asmaloney:
Updated commit to generate the file in the CC/include folder.

dgirardeau:
Not sure why but I now have a 'link' error:
```CC_CORE_LIB.lib(CC_CORE_LIB.dll) : error LNK2005: "public: __cdecl std::vector<float,class std::allocator<float> >::~vector<float,class std::allocator<float> >(void)" (??1?$vector@MV?$allocator@M@std@@@std@@QEAA@XZ) déjà défini(e) dans Mouse3DInput.obj```

**Comment by dgirardeau** [comment]
Oh and I have another one when compiling qCC_db: ```LNK2005	"public: __cdecl std::vector<float,class std::allocator<float> >::vector<float,class std::allocator<float> >(class std::vector<float,class std::allocator<float> > const &)" (??0?$vector@MV?$allocator@M@std@@@std@@QEAA@AEBV01@@Z) already defined in ccDepthBuffer.obj	QCC_DB_LIB	E:\These\C++\CloudCompare\build_x64_msvc_2017\libs\qCC_db\CC_CORE_LIB.lib(CC_CORE_LIB.dll)```

**Comment by asmaloney** [comment]
Did you run "clean", run "cmake", and try to build? (i.e. totally clean build)

std::vector allocator? Hmm...  Not a lot to go on :-)

Are you building CCLib statically or as a DLL?

**Comment by dgirardeau** [comment]
The second error actually appeared after I made a 'clean' build :D

And I compile CC_CORE_LIB as a DLL. And yes, sadly, this std::vector error is all I have :(

**Comment by asmaloney** [comment]
The only visibility I changed directly in this PR is that of `ScalarField` because I was getting link errors otherwise.

Can you please try reverting those changes to see if that's the culprit?

**Comment by dgirardeau** [comment]
Ok. I'm currently working on another branch, but I'll test that just after.

**Comment by asmaloney** [comment]
I suspect it has to do with that inherited `std::vector`.

We explicitly turn off the warning you probably should be getting about it:

```c++
//To get rid of the really annoying warnings about template class exportation
#pragma warning( disable: 4251 )
```

asmaloney:
OK - I'm assuming that's the issue. The general consensus from everything I'm reading is "don't export anything STL - you will feel pain".

Here's an [analysis of the warning C4251](http://www.unknownroad.com/rtfm/VisualStudio/warningC4251.html) that's turned off.

This guy [describes the problem in detail](https://www.codesynthesis.com/~boris/blog/2010/01/18/dll-export-cxx-templates/).

He points at this [Microsoft article](https://docs.microsoft.com/en-us/cpp/cpp/general-rules-and-limitations?view=vs-2019) (the last point).

So we can try to use an explicit instantiation just before the class like Microsoft recommends?

At the top, outside the `CCLib` namespace:

```c++
template class CC_CORE_LIB_API std::vector<ScalarType>;
```

dgirardeau:
It's actually worse in a way. It conflicts with definitions of std::vector in third party libraries (PoissonRecon, PCL, etc.)

asmaloney:
Maybe that needs to be `extern` and then explicitly instantiated in the cpp:

```c++
extern template class CC_CORE_LIB_API std::vector<ScalarType>;
```

In the cpp:

```c++
template class std::vector<ScalarType>;
```

dgirardeau:
So none of the solutions above worked, apart from restoring the previous code (with only a few methods exported).

asmaloney:
OK thanks. Using the previous code results in link errors on macOS (and presumably Linux if it uses an so). I can re-examine that, but we might be stuck "between compilers" here.

This is a painful way for to debug this since I can't compile CC on Windows at the moment. :-) What happened with the last solution I proposed (the 'extern' one)? That seems to be the recommendation...

dgirardeau:
the 'extern' symbol can't be used with '__declspec(dllexport)'

dgirardeau:
And how come this code was working before? What's the difference on macOS?

asmaloney:
So before this change, _all_ the symbols are exported on macOS/Linux. This change (like the others) makes it more like Windows and only exports certain symbols.

Windows by default hides all symbols and you tell it which to export. macOS/Linux does the opposite and exports them all unless you tell it not to.

dgirardeau:
The error on clang is not very informative either, isn't it? (on Travis)
```../../qCC_db/libQCC_DB_LIB.so: undefined reference to `typeinfo for CCLib::ScalarField'```

Do you get more info on your side?

dgirardeau:
Apparently, it's the exported virtual methods that can cause this? Like ```CC_CORE_LIB_API virtual void computeMinAndMax();```?

asmaloney:
Yes - we need to export the vtable. Not sure how to do that without exporting the class as a whole :-)

asmaloney:
Based on [this from Microsoft](https://docs.microsoft.com/en-us/cpp/cpp/using-dllimport-and-dllexport-in-cpp-classes?view=vs-2019):

> If you have a class in which you are using selective member import/export with virtual functions, the functions must be in the exportable interface or defined inline (visible to the client).

It seems like we're doing it incorrectly (before this PR).

Once I fix this by inlining `computeMinAndMax()` in this PR, it works for me.

dgirardeau:
Let's go with it then!

dgirardeau:
By the way, before closing this PR, is it possible to rename the file? (it doesn't match the other filenames). Or is it automatically generated by CMake?

asmaloney:
> is it possible to rename the file?

Yes!

> it doesn't match the other filenames

I thought that was good since it's a generated file and obviously different.

But what do you want to name it?

dgirardeau:
CCCoreLibExport.h?

asmaloney:
Thanks for the discussion/work through of this issue!

I learned "Some Stuff"™.

PR #896, opened 2019-04-01
Fix operator precedence

asmaloney:
```c++
(warningIfOutside && z < -1.0e-6 || z > 1.0 + 1.0e-6)
```

is parsed as:
```c++
((warningIfOutside && z < -1.0e-6) || z > 1.0 + 1.0e-6)
```
since `&&` has a higher precedence than `||`. I don't think this is the intent.

It's always best to include parentheses to avoid this problem - and as a bonus it makes it more human-compatible!


dgirardeau:
Nice catch! (I added the boolean later and I didn't pay attention)

PR #899, opened 2019-04-04
textColorButton fix

StOriJimmy:
textColorButton typo


asmaloney:
Thank you - nice catch!

PR #901, opened 2019-04-05
check the size m_pointsVisibility instead of capacity

StOriJimmy:
I think the m_pointsVisibility should be checked by size rather than capacity since I didn't see any function like shrink_to_fit called. When we segment the segmented point cloud again (i.e. click segment button at “xx - Cloud.remained”), this assert triggered!


dgirardeau:
Agreed!

PR #920, opened 2019-05-19
Ensure that RPATH is absolute

jnumm:
GNUInstallDirs provides a default value for CMAKE_INSTALL_LIBDIR
and allows a user to override it as either absolute or relative
(eg. -DCMAKE_INSTALL_LIBDIR=/usr/lib64 or -DCMAKE_INSTALL_LIBDIR=lib64).
CMAKE_INSTALL_FULL_LIBDIR is the absolute calculated from it.

This popped up in Debian [#928986](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=928986): the RPATH incorrectly being a relative path caused libQCC_IO_LIB.so not to be found.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,10 +28,8 @@ set( INSTALL_DESTINATIONS CloudCompare )
 # RPATH Linux/Unix: (dynamic) libs are put in a subdir of prefix/lib,
 # since they are only used by qCC/ccViewer
 if( UNIX AND NOT APPLE )
-	if( NOT CMAKE_INSTALL_LIBDIR )
-		set( CMAKE_INSTALL_LIBDIR ${CMAKE_INSTALL_PREFIX}/lib CACHE PATH "CloudCompare lib dir" )
-	endif( NOT CMAKE_INSTALL_LIBDIR )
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -7,6 +7,8 @@ project( CloudCompareProjects )
 option(BUILD_TESTING "" OFF)
 include(CTest)
 
+include( GNUInstallDirs )
```
dgirardeau: Is it defined/existing on Windows?

jnumm: Yes, it exists on Windows. I just ran a test compilation with Windows 8, cmake 3.9, Visual Studio 2017.

asmaloney: Since it's only used in the Linux case, please wrap in `if ( UNIX AND NOT APPLE )`.

rjanvier: Thanks for the contribution. I’m very Happy to see a Debian contributor trying to fix issue upstream. Please take into account @asmaloney remark. else it LGTM.

jnumm: Per asmaloney's request, I moved the `include`.

rjanvier:
thanks.

jnumm:
Happy to help. Thanks to all the reviewers!

PR #922, opened 2019-05-27
cmake: fix multiple else in one block error

NeroBurner:
Multiple else branches for one `if` statement. There should be `elseif` checks

```cmake
if( WIN32 )
  ...
else( APPLE )
  ...
else( UNIX )
```


asmaloney:
Nice catch!

Is it still legal syntax somehow? The [docs](https://cmake.org/cmake/help/v3.14/command/if.html) imply that it isn't, but then I would expect an error from cmake.

Anyways - the `else( UNIX )` should be changed to `elseif( UNIX )` as well:

> Per legacy, the else() and endif() commands admit an optional <condition> argument. If used, it must be a verbatim repeat of the argument of the opening if command.

NeroBurner:
sry, did not update the master branch from upstream...

Issue #848, opened 2019-02-19
CCConst.h:43:40: error: ‘(1.8e+2 / 3.1415926535897931e+0)’ is not a constant expression


swiss-knight:
I'm trying to compile CC from sources on Ubuntu 18.04 and I finally encountered this error:

```
In file included from /opt/CloudCompare/CC/include/FastMarching.h:23:0,
                 from /opt/CloudCompare/CC/include/FastMarchingForPropagation.h:24,
                 from /opt/CloudCompare/CC/src/AutoSegmentationTools.cpp:22:
/opt/CloudCompare/CC/include/CCConst.h:43:40: error: ‘(1.8e+2 / 3.1415926535897931e+0)’ is not a constant expression
 constexpr double CC_RAD_TO_DEG = (180.0/M_PI);
                                  ~~~~~~^~~~~~
/opt/CloudCompare/CC/include/CCConst.h:46:39: error: ‘(3.1415926535897931e+0 / 1.8e+2)’ is not a constant expression
 constexpr double CC_DEG_TO_RAD = (M_PI/180.0);
                                  ~~~~~^~~~~~~
CC/CMakeFiles/CC_CORE_LIB.dir/build.make:62: recipe for target 'CC/CMakeFiles/CC_CORE_LIB.dir/src/AutoSegmentationTools.cpp.o' failed
make[2]: *** [CC/CMakeFiles/CC_CORE_LIB.dir/src/AutoSegmentationTools.cpp.o] Error 1
CMakeFiles/Makefile2:987: recipe for target 'CC/CMakeFiles/CC_CORE_LIB.dir/all' failed
make[1]: *** [CC/CMakeFiles/CC_CORE_LIB.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....

```

## Your Environment
* CC Version: master from sources
* OS & Version: Ubuntu 18.04 64 bit
* Graphics card: On CPU (Intel)
### Other relevant details about the environment: 
* gcc : 7.3.0
* cmake suite: 3.13.20181205

Here is the output log of `cmake-gui` before build:

```
cmake policies active:
cmake policies active:
__cplusplus is 201103
  --> Do not link with Boost.Thread
Boost version: 1.68.0
Boost include dirs: /usr/local/include
Boost libraries:    
Build type: Release
USING CXXFLAGS = '-std=c++17 -std=c++11 -fPIC -std=c++11 -fPIC -O3 -DNDEBUG'
[...]

Issue #835, opened 2019-01-11
constexpr with floating operation, invalid with gcc and CGAL


bartoszek:
There is a small problem with `constexpr` and CGAL. Code calling CGAL has to be devoided of a float operation in `constexpr` as there is a problem to kept FENV in GCC compiler.
_problem introduced in_ #832 
_CGAL rationale_ https://github.com/CGAL/cgal/issues/3180

rjanvier:
cgal marked it as "won't fix", so I afraid as long as we link with cgal it will be a "won't fix" too i.e. no float in constexpr, at least for GCC. 
cc @asmaloney .

asmaloney:
I'm not finding much of a rationale there.

Having this compiler flag propagate to all user code is problematic - you cannot use something as simple (and allowed by the standard) as `constexpr double foo = 1.0 / 42.0;` in your code anymore.

If we end up including other libraries that use constexpr as well, what do we do?

rjanvier:
Wz can also replace cgal by the news QHULL revision that do constrained Delaunay triangulation. I don't know how it compare vs cgal in term of speed and robustness. The other by product could be in license compatibility. Part of cgal we use are GPL so it's viral...

asmaloney:
This one? 

https://github.com/qhull/qhull

rjanvier:
yes Qhull 2018.0.1, we already embed one version of qhull (for qhpr)

bartoszek:
[`naive` patch for linux build.](https://aur.archlinux.org/cgit/aur.git/plain/constexpr.patch?h=cloudcompare-git)

AlexandrParkhomenko:
oops, I try add this




gcc (GCC) 8.2.1 20181127 compilation error

**How to reproduce**
MAKE and out:
CloudCompare/libs/qCC_db/ccCameraSensor.cpp:56:54:
 error: ‘(5.4500000000000002e+0 * 1.0e-3)’ is not a constant expression
  constexpr float focal_mm  = static_cast<float>(5.45 * 1.0e-3); // focal length (real distance in meter)


**Expected behaviour**

using here "const" or dont using multiply "*"

**Additional context**


Please apply this patch:
https://aur.archlinux.org/cgit/aur.git/tree/constexpr.patch?h=cloudcompare-git


**Your environment**
 - CC Version (or git hash/tag): 2.10.2.r206.g7699efd9
 - OS & Version: https://aur.archlinux.org/packages/cloudcompare-git/
[...]

PR #923, opened 2019-06-12
replace constexpr for floating point operations

christian-rauch:
This fixes a compilation error related to `constexpr` with floating point operations which is caused by propagating flag `-frounding-math` from CGAL (https://github.com/CGAL/cgal/issues/3180). See https://github.com/CloudCompare/CloudCompare/issues/835, https://github.com/CloudCompare/CloudCompare/issues/848.

This is a workaround for the issue until a proper solution (replacing CGAL, not propagating `-frounding-math`) is found.

### DIFF
--- a/CC/include/CCConst.h
+++ b/CC/include/CCConst.h
@@ -40,10 +40,10 @@ constexpr double M_PI_2 = (M_PI/2.0);
 constexpr double SQRT_3 = 1.7320508075688772935274463415059;
 
 //! Conversion factor from radians to degrees
-constexpr double CC_RAD_TO_DEG = (180.0/M_PI);
+const double CC_RAD_TO_DEG = (180.0/M_PI);
 
 //! Conversion factor from degrees to radians
-constexpr double CC_DEG_TO_RAD = (M_PI/180.0);
+const double CC_DEG_TO_RAD = (M_PI/180.0);
 
 //! Numerical threshold for considering a value as "zero"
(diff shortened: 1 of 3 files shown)


asmaloney:
This is a [CGAL issue](https://github.com/CGAL/cgal/issues/3180) and needs to be fixed there. 

The fact that they propagate `-frounding-math` means that they will break any code like this:

```cpp
constexpr double foo = 1.0 / 42.0;
```

This not only prevents us from using valid C++, but it also prevents us from including any other code/libraries that use the language in this way.

So it needs to be fixed at the source of the problem, not patched around. Please submit issues [over there](https://github.com/CGAL/cgal/issues).

I would be happy to take a PR for a "proper solution" like replacing CGAL.

christian-rauch:
I know that this is a CGAL issue but CloudCompare introduced CGAL as a dependency and they are not willing to fix this issue.

What would be the best approach to replace CGAL?

asmaloney:
@rjanvier will have to give some guidance since he did this work involving CGAL.

In [the other issue](https://github.com/CloudCompare/CloudCompare/issues/835), he suggests replacing our constrained Delaunay triangulation with [qhull](https://github.com/qhull/qhull).

The CGAL work was added in 4bcb3f0fac4cce0f13d2d2354ddddd51458e7ee6, so you might be able to use that as a reference point.

(Looks like you can also compile CC without CGAL, so that might be an option for you.)

asartori86:
I dumped on the same issue. "fixed" using `clang++` instead of `g++`

PR #924, opened 2019-06-12
flatpak support

christian-rauch:
This PR adds [flatpak](https://flatpak.org) support as an alternative to snap.

Quick start (see `README.md` for details):
1. install flatpak with Flathub repository (https://flatpak.org/setup/)
2. install the KDE runtime and SDK
3. `flatpak-builder --user --install /tmp/build_cc_app flatpak/manifest.yaml`

Most plugins are included. 3 plugins are disabled because of missing dependencies or compilation errors.


rjanvier:
Thanks a lot Christian. 
I didn't test it but it's clearly a very good thing. It seems harmless for the other part of CC so the PR could be merged. Nevertheless, I think I'm not 100% in favor of merging this into the main repo since it's totally independent from the code (the same thing applies for the snap _a posteriori_). 
So maybe we could create a cloudcompare-flathub / cloudcompare-linux-packages repo with only the sources of the flathub/snap packages (and not the sources of CC) under the CloudCompare organization on GitHub ? Do you think it's a good idea?

christian-rauch:
I added the manifest to the main repo, so that the flatpak app can be generated from the current source. This way, it can be installed as flatpak app without needing to wait for a release.
It also simplifies maintaining the manifest, e.g. when issues get resolved and plugins can be enabled or when dependencies change.

However, for releasing the app to flathub, a dedicated repository needs to be created anyway at: https://github.com/flathub.

As a side note:
I am building and releasing from `master` to solve some other compilation issues. Is there a time estimate when the next stable release (version 2.11) will become available?

rjanvier:
It's exactly I want to avoid: to have a high coupling between packaging and the software. ideally issues in the CC repos should only deal about compilation and usage not about peculiarities of X or Y packaging system. 
I think in the manifest, you can replace the "- type: dir" by a "-type: git" this way users could always compile CC with your manifest even if it's not bundled with CC sources.
For the ETA and the creation of an "official" CC account on flathub, you should see with @dgirardeau. We could always do a "point release" if needed.

dgirardeau:
Ah ah don't involve me in this, I don't understand anything about Linux :D 

Ok I'm - partially - kidding. I can do anything that may help flatpak users as long as it's not too complicated ;). So how do we proceed?

rjanvier:
Just read how to proceed. no need of account, just need to push the manifest file on a repo somewhere. see
https://github.com/flathub/flathub/wiki/App-Submission
@christian-rauch I think you are free to push CC on flathub. @dgirardeau seems ok with it. You can refer to this thread in your PR to flathub.

christian-rauch:
I have a PR pending for CloudCompare: https://github.com/flathub/flathub/pull/1037

@dgirardeau 
It was suggested that desktop files like icons, launchers and appdata metadata is upstreamed. Would you be willing to create SVG-icons and generic launchers that can be used for snap, flatpak and regular distro packages (dep, rpm)?
To upstream the appdata metadata file, you can use the file that I included in the flathub PR: https://github.com/flathub/flathub/blob/5800e7f2eab5db3a2d9768cf4968c505dd5370e5/net.danielgm.CloudCompare.appdata.xml

If there are no big changes coming up for CloudCompare (and no blocking issues), it would be useful to have a point release to have a proper tag for the flathub release.

bilelmoussaoui:
Hey, I'm one of the flathub "third-party maintainers" (Just a random geek who tries to make it easier for Linux users to get the best and the latest software easily). You can ping me in the future if you need any help related to Flatpak or Linux packaging in general.

Regarding the appdata, desktop file and the icons, they should definitely be upstreamed as they are not related to Flathub or Flatpak but part of the Freedesktop standards. 

The appdata file is basically an XML file that describes your application for any Linux software center out there. The desktop file is used to create a launcher for the application.

The specs for appdata: https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html#sect-Quickstart-DesktopApps
The specs for desktop files: https://specifications.freedesktop.org/desktop-entry-spec/latest/

dgirardeau:
@christian-rauch We do have SVG icons: 
- https://github.com/CloudCompare/CloudCompare/blob/master/qCC/images/icon/cc_icon.svg
- https://github.com/CloudCompare/CloudCompare/blob/master/qCC/images/icon/cc_viewer_icon.svg
(however they rely on a strange font that might not be available everywhere...)

And can you clarify what you mean by 'upstream the appdata metadata file'?

Otherwise we can probably release the official 2.10.3 version (that's a minor release but with interesting bug fixes).

christian-rauch:
@dgirardeau I think with upstreaming the launchers, icons and the [appdata meta file](https://raw.githubusercontent.com/flathub/flathub/217c820710f256715a8ad12bbdb203a06330a979/org.cloudcompare.CloudCompare.appdata.xml), it is meant that they become part of this repo and get installed to `share/applications`, `share/icons` and `share/metainfo` respectively. (@bilelmoussaoui Is this correct?)

asmaloney:
>Otherwise we can probably release the official 2.10.3 version (that's a minor release but with interesting bug fixes).

@dgirardeau I don't want to hijack this thread, but there are several things in master that should be cherry picked to the 2.10.x branch before releasing 2.10.3. (And some are listed in the changelog as 2.10.3, but aren't in the branch.)

bilelmoussaoui:
@christian-rauch Yes. Those files can be translated (once or if you have i18n support) and should be part of the build system. This allow you to easily change the appdata file when there's a new release. Flathub requires a release tag in order to show which release is hosted on Flathub & Software center on Linux can show from which to which is the user is upgrading, the changelog & which version is installed.

dgirardeau:
@christian-rauch Not sure how to do that sorry! @rjanvier and the others: Is it something specific to this kind of tools or can it be done on any Linux distrib/setup?

rjanvier:
@dgirardeau to generate the correct appdata meta file on the fly for each build/version it's maybe possible on all plateform with a config file https://cmake.org/cmake/help/v3.2/command/configure_file.html. But it's also possible to do it with Linux only command line tools and guard it into the cmake with an `if(Linux)...`. for the l18 thing I don't know exactly how to do that but as we only support a "small" set of languages; it's maybe preferable to translate it once for now.

dgirardeau:
Sorry I am back. It's not clear to me what should be the next step(s).

On my side I'm definitely not able to customize anything for the linux world ;)

christian-rauch:
@dgirardeau It would be useful to have the launchers and SVG icons installed during the CMake install step. You can use the launchers (*.desktop files) from my flathub PR.

rjanvier:
@christian-rauch would you like to propose a dedicated PR please?

christian-rauch:
I am surprised that this PR is still open after this time :-)

I am getting contributions to the flatpak configuration on flathub (https://github.com/flathub/org.cloudcompare.CloudCompare). I agree with some comments in this PR that the flatpak definition and the project itself should be separate repos.

The remaining desktop integration files (icons, starters, ...) can also be maintained in the flathub repo. There is already a snap package definition with icon files in this repo. But if the snap definition gets removed eventually (because separate code and package definition) then using it would also fail.

I am going to close this, as I have no intentions to get the flatpak definition upstreamed. But I have two requests :-)

1. Could the desktop integration files from the snap package be moved into some official folder in the repo? This way the icons would be "upstream approved" and could be resued by other package formats. The `.desktop` files in the snap use some templated format (${SNAP}/meta/gui). Could you move these files too into an "official" folder and have the icon and executable path configured via CMake? Alternatively, the `.desktop` files in the flatpak just use `CloudCompare` as the executable and `org.cloudcompare.CloudCompare` for the icon file name, without an absolute path. And this seems to work too.

2. Could you advertise the flatpak in parallel to the snap package on the release page too? :-)

dgirardeau:
1. Yes, this is definitely possible, do you want to make a PR for that?
2. I guess ;) ... We just need to add the command` flatpak install flathub org.cloudcompare.CloudCompare` as instruction?

Issue #598, opened 2017-09-10
Set up a Travis build for ScalarType of "double"


asmaloney:
By default `ScalarType` is a float, but some people compile CC with `ScalarType` as a double.

We should set up a Travis build for this case.

Bedzior:
Hey, I have just come by this issue. Until recently I have only occasionally used CloudCompare, but now I would love to make a contribution.

I'll try posting a merge request soon.

Bedzior:
@asmaloney hopefully #936 resolved this one :)

tmontaigu:
Closing this, as it was done in the PR linked above

asmaloney:
Thanks @tmontaigu. I just looked at that PR and it's not doing what I expect it to do - i.e. it's not building when `ScalarType` is double.

Take a look at any of the Travis builds (like [this one](https://travis-ci.org/CloudCompare/CloudCompare/builds/653528935?utm_source=github_status&utm_medium=notification)). See how the `SCALAR_DOUBLE=ON` builds only take ~2 minutes? That's because it's not actually building.

For some reason in this PR, we find this: ` && [[ $SCALAR_DOUBLE != "ON" ]]; ` which turns off the build. ??

tmontaigu:
Indeed, I looked at the logs took quickly
It seems to be a mistake that the build stops. What the author wanted to do, is disabling tests when building with scalar

According to the 3rd point in the PR :
> a modification to Travis' build script - turning off tests for double comparisons in TestShpFilter, as the test data seems to be intended for float only and I don't feel competent enough to do anything with it.

Bedzior:
Sorry for this egregious oversight. The build script was obviously supposed to run independent of `ScalarType`'s actual type.

Here's the PR resolving this issue: #1042. I have conditionally turned the tests off (as, again, they seem type-specific). One type had to be fixed, but it seemed straightforward.

asmaloney:
No problem. Thanks for the fix!

PR #936, opened 2019-08-03
Building with ScalarType as double (with fixes)

Bedzior:
I am posting this pull request despite having some doubts.

This PR contains the following changes:
- the actual substance of #598, the issue which prompted the changes with type fixes and static casts to supply proper types;
- a fix to a compilation error that occurred with `ScalarType = double` - namely, [change in parameter order in AutoSegmentationTools](CC/src/AutoSegmentationTools.cpp) - I was surprised this went on unnoticed!
- a modification to Travis' build script - turning off tests for double comparisons in `TestShpFilter`, as the test data seems to be intended for `float` only and I don't feel competent enough to do anything with it.

I kindly ask for suggestions regarding improving this PR.

### DIFF
--- a/CC/include/CCTypes.h
+++ b/CC/include/CCTypes.h
@@ -23,6 +23,12 @@
 using PointCoordinateType = float;
 
 //! Type of a single scalar field value
+#if defined SCALAR_TYPE_DOUBLE
+using ScalarType = double;
+#elif defined SCALAR_TYPE_FLOAT
 using ScalarType = float;
+#else
+static_assert(false, "type for ScalarType has not been declared");
+#endif //SCALAR_TYPE_DOUBLE
 
 #endif //CC_TYPES_HEADER
--- a/libs/qCC_db/ccGBLSensor.cpp
+++ b/libs/qCC_db/ccGBLSensor.cpp
@@ -1051,8 +1051,11 @@ bool ccGBLSensor::fromFile_MeOnly(QFile& in, short dataVersion, int flags)
 	ccSerializationHelper::CoordsFromDataStream(inStream, flags, &m_deltaTheta, 1);
 	if (dataVersion < 38)
 	{
-		ccSerializationHelper::ScalarsFromDataStream(inStream, flags, &m_sensorRange, 1);
-		ccSerializationHelper::ScalarsFromDataStream(inStream, flags, &m_uncertainty, 1);
+		ScalarType sensorRange{}, uncertainty{};
+		ccSerializationHelper::ScalarsFromDataStream(inStream, flags, &sensorRange, 1);
+		ccSerializationHelper::ScalarsFromDataStream(inStream, flags, &uncertainty, 1);
+		m_sensorRange = static_cast<PointCoordinateType>(sensorRange);
+		m_uncertainty = static_cast<PointCoordinateType>(uncertainty);
 	}
 	else
 	{
--- a/libs/qCC_glWindow/ccRenderingTools.cpp
+++ b/libs/qCC_glWindow/ccRenderingTools.cpp
@@ -31,6 +31,7 @@
 #include <QLabel>
 #include <QVBoxLayout>
 
+#include <numeric>
 
 void ccRenderingTools::ShowDepthBuffer(ccGBLSensor* sensor, QWidget* parent/*=0*/, unsigned maxDim/*=1024*/)
 {
@@ -44,8 +45,8 @@ void ccRenderingTools::ShowDepthBuffer(ccGBLSensor* sensor, QWidget* parent/*=0*
 	}
 
 	//determine min and max depths
-	ScalarType minDist = 0.0f;
-	ScalarType maxDist = 0.0f;
+	PointCoordinateType minDist = 0.0f;
+	PointCoordinateType maxDist = 0.0f;
 	{
 		const PointCoordinateType* _zBuff = depthBuffer.zBuff.data();
 		double sumDist = 0.0;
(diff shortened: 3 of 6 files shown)


CC/include/CCTypes.h
```diff
@@ -23,6 +23,12 @@
 using PointCoordinateType = float;
 
 //! Type of a single scalar field value
+#if defined SCALAR_T_DOUBLE
+using ScalarType = double;
+#elif defined SCALAR_T_FLOAT
 using ScalarType = float;
+#else
+static_assert(false, "type for ScalarType has not been undeclared");
```
dgirardeau: Do you mean 'declared' instead of undeclared?

CC/include/CCTypes.h
```diff
@@ -23,6 +23,12 @@
 using PointCoordinateType = float;
 
 //! Type of a single scalar field value
+#if defined SCALAR_T_DOUBLE
```
dgirardeau: I would be more explicit and use 'SCALAR_TYPE_DOUBLE' or 'SCALAR_TYPE_FLOAT'

CC/src/AutoSegmentationTools.cpp
```diff
@@ -129,8 +129,8 @@ bool AutoSegmentationTools::extractConnectedComponents(GenericIndexedCloudPersis
 }
 
 bool AutoSegmentationTools::frontPropagationBasedSegmentation(	GenericIndexedCloudPersist* theCloud,
-																PointCoordinateType radius,
 																ScalarType minSeedDist,
+																PointCoordinateType radius,
```
dgirardeau: Nice catch! Luckily this method is not used anymore.

Out of consistency with what is generally done in CC, I would have kept the radius before the seed parameter (so I would have inverted them in the header file and not in the source file).

libs/qCC_db/ccGBLSensor.cpp
```diff
@@ -1051,8 +1051,8 @@ bool ccGBLSensor::fromFile_MeOnly(QFile& in, short dataVersion, int flags)
 	ccSerializationHelper::CoordsFromDataStream(inStream, flags, &m_deltaTheta, 1);
 	if (dataVersion < 38)
 	{
-		ccSerializationHelper::ScalarsFromDataStream(inStream, flags, &m_sensorRange, 1);
-		ccSerializationHelper::ScalarsFromDataStream(inStream, flags, &m_uncertainty, 1);
+		ccSerializationHelper::ScalarsFromDataStream(inStream, flags, static_cast<ScalarType*>(&m_sensorRange), 1);
```
dgirardeau: Actually there was an error here. m_sensorRange and m_uncertainty are declared as PointCoordinateType. Therefore one should use CoordsFromDataStream to read them.

libs/qCC_glWindow/ccRenderingTools.cpp
```diff
@@ -59,8 +60,8 @@ void ccRenderingTools::ShowDepthBuffer(ccGBLSensor* sensor, QWidget* parent/*=0*
 			}
 			else if (*_zBuff > 0)
 			{
-				maxDist = std::max(maxDist, *_zBuff);
-				minDist = std::min(minDist, *_zBuff);
+				maxDist = std::max(maxDist, static_cast<const ScalarType>(*_zBuff));
```
dgirardeau: I believe in this method, there was another mistake. minDist, maxDist and coef should all be of 'PointCoordinateType' type.

CC/src/AutoSegmentationTools.cpp
Bedzior: I was expecting this to be the case. I intended to keep the interface intact to avoid compatibility issues, but consistency is sound argument.

CC/include/CCTypes.h
Bedzior: 🤦‍♂

libs/qCC_db/ccGBLSensor.cpp
Bedzior: This would somehow conflict with what seems to be some logic here. The `dataVersion` variable seems to delimit a specific range of data format versions, where the data is to be read differently.

Making the suggested changes, I could basically remove the `if` statement surrounding this together with `else` case.

Bedzior: Hey @dgirardeau, any movements on this?

dgirardeau: Sorry I was in vacations. To be on the safe side, we should then 'cast' the m_sensorRange and m_uncertainty variables to ScalarType ones before saving them. And same thing, read them with the ScalarsFromDataStream and temporary ScalarType variables.

Bedzior: Would the following be correct?
```cpp
ScalarType sensorRange{}, uncertainty{};
ccSerializationHelper::ScalarsFromDataStream(inStream, flags, &sensorRange, 1);
ccSerializationHelper::ScalarsFromDataStream(inStream, flags, &uncertainty, 1);
m_sensorRange = static_cast<PointCoordinateType>(sensorRange);
m_uncertainty = static_cast<PointCoordinateType>(uncertainty);
```

As I understand, the temporary `ScalarType` values are there to secure us from reading too much from the buffer (into wrong memory), but I might not be getting this part of your request correctly:
> we should then 'cast' the m_sensorRange and m_uncertainty variables to ScalarType ones before saving them

dgirardeau: Yes I think it's good like that.

dgirardeau:
Thanks for the contribution

PR #954, opened 2019-09-11
Track progress in each stage of OBJ saving

EpicWink:
Update the save progress bar for each of the mesh attributes saved, updating the dialogue label for each. The dialogue is kept open between stages

Previously, the dialogue would stay uninitialised (black) while the vertices, vertex normals, materials and texture coordinates are written, leading to a frustrating experience where the user could think that CloudCompare had crashed during the saving of a large file as OBJ.

By default, the progress-bar dialogue is closed when the progress-bar becomes full, and then re-opened when sufficient iterations have passed. This is confusing, and is prevented by setting `autoClose` to `false`.

### DIFF
--- a/plugins/core/IO/qCoreIO/src/ObjFilter.cpp
+++ b/plugins/core/IO/qCoreIO/src/ObjFilter.cpp
@@ -103,18 +103,17 @@ CC_FILE_ERROR ObjFilter::saveToFile(ccHObject* entity, const QString& filename, 
 	if (!file.open(QFile::Text | QFile::WriteOnly))
 		return CC_FERR_WRITING;
 
-	unsigned numberOfTriangles = mesh->size();
-
-	//progress
+	//progress (start with vertices)
 	QScopedPointer<ccProgressDialog> pDlg(nullptr);
 	if (parameters.parentWidget)
 	{
 		pDlg.reset(new ccProgressDialog(true, parameters.parentWidget));
 		pDlg->setMethodTitle(QObject::tr("Saving mesh [%1]").arg(mesh->getName()));
-		pDlg->setInfo(QObject::tr("Triangles: %1").arg(numberOfTriangles));
+		pDlg->setInfo(QObject::tr("Writing %1 vertices").arg(nbPoints));
+		pDlg->setAutoClose(false); //don't close dialogue when progress bar is full
 		pDlg->start();
 	}
-	CCLib::NormalizedProgress nprogress(pDlg.data(), numberOfTriangles);
+	CCLib::NormalizedProgress nprogress(pDlg.data(), nbPoints);
 
 	QTextStream stream(&file);
 	stream.setRealNumberNotation(QTextStream::FixedNotation);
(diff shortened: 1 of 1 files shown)


EpicWink:
Let me know if you would prefer if the progress bar built up over the course of all stages, only getting full when the last stage is completed.

plugins/core/IO/qCoreIO/src/ObjFilter.cpp
```diff
@@ -145,14 +146,26 @@ CC_FILE_ERROR ObjFilter::saveToFile(ccHObject* entity, const QString& filename,
 		if (withTriNormals)
 		{
 			NormsIndexesTableType* normsTable = mesh->getTriNormsTable();
+
+			//reset save dialog
+			unsigned numTriangleNormals = normsTable->currentSize();
+			if (parameters.parentWidget)
```
dgirardeau: can you test 'pDlg' directly? (instead of parameters.parentWidget). Seems more straightforward.

plugins/core/IO/qCoreIO/src/ObjFilter.cpp
```diff
@@ -164,12 +177,24 @@ CC_FILE_ERROR ObjFilter::saveToFile(ccHObject* entity, const QString& filename,
 		//per-vertices normals
 		else //if (withVertNormals)
 		{
+			//reset save dialog
+			if (parameters.parentWidget)
```
dgirardeau: same here

plugins/core/IO/qCoreIO/src/ObjFilter.cpp
```diff
@@ -208,12 +243,27 @@ CC_FILE_ERROR ObjFilter::saveToFile(ccHObject* entity, const QString& filename,
 		TextureCoordsContainer* texCoords = mesh->getTexCoordinatesTable();
 		if (texCoords)
 		{
+			//reset save dialog
+			unsigned numTexCoords = texCoords->currentSize();
+			if (parameters.parentWidget)
```
dgirardeau: same thing

plugins/core/IO/qCoreIO/src/ObjFilter.cpp
```diff
@@ -239,6 +289,16 @@ CC_FILE_ERROR ObjFilter::saveToFile(ccHObject* entity, const QString& filename,
 		}
 	}
 
+	//reset save dialog for triangles
+	unsigned numTriangles = mesh->size();
+	if (parameters.parentWidget)
```
dgirardeau: wait for it... same thing here ;)

dgirardeau:
> Let me know if you would prefer if the progress bar built up over the course of all stages, only getting full when the last stage is completed.

No I'm good this way, it makes the code simpler, and as you said, the important thing is that the user realizes that CC is still working ;)

EpicWink:
> Thanks for the contribution!
> Just a minor concern: have you estimated the additional time necessary to update the progress dialog for each vertex/triangle/normal/etc.?

No I haven't: I will tomorrow (later today). I suspect it will be quite significant (serialising a single line of OBJ vs large amounts of Qt calls), in which case I'll pull out my old friend the modulo operator.

plugins/core/IO/qCoreIO/src/ObjFilter.cpp
EpicWink: You missed the next one 😄

EpicWink:
Did the timings:
* With progress-bar updating: 25.4 seconds
* Without progress-bar updating 25.2 seconds

Less than 1% change

Configuration
* CPU: Intel i7 7700HQ @ 3.7GHz
* Memory: Crucial 1x8GB and 1x16GB @ 2.4GHz, CL 17
* Storage: Intel SSDPEKKW256G7 256GB SSD
* Mesh:
  - 3835000 faces
  - 1997000 vertices
  - 2862000 texture coordinates
  - 297000000 pixels of texture (most in use), 25 materials
  - 313 MB on disk, with 80MB of textures (JPG)

Issue #382, opened 2016-05-01
Smooth zoom


foobarbecue:
The recent addition of the ability to change the zoom increment was very welcome. However, it would be nice if we could click and drag to zoom smoothly. I'd be happy with a key combo or middle mouse click+drag.

(I work on landscape-scale models where there are a few areas where high detail is needed so I'm often working at the km scale and then the cm scale on the same point cloud / mesh.)

cloudcompare:
Okay I added this to the TODO list.

NeroBurner:
Any updates on this? Where would I have to start to add this functionality to the viewer?

dgirardeau:
I guess you would have to change how the zoom is handled, typically with the wheel mouse for now: https://github.com/CloudCompare/CloudCompare/blob/master/libs/qCC_glWindow/ccGLWindow.cpp#L4520

PR #962, opened 2019-10-03
GLWindow: handle middle mouse button press, use for zoom

NeroBurner:
- handle the middle mouse button press
- when the middle mouse button is pressed use the delta along the y axis
  to zoom the view (like the mouse wheel)

fixes https://github.com/CloudCompare/CloudCompare/issues/382

### DIFF
--- a/libs/qCC_glWindow/ccGLWindow.cpp
+++ b/libs/qCC_glWindow/ccGLWindow.cpp
@@ -3846,6 +3846,14 @@ void ccGLWindow::mousePressEvent(QMouseEvent *event)
 			emit leftButtonClicked(event->x(), event->y());
 		}
 	}
+	if (event->buttons() & Qt::MiddleButton)
+	{
+		//middle click = zooming
+		if (m_interactionFlags & INTERACT_SIG_MB_CLICKED)
+		{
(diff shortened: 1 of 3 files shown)


libs/qCC_glWindow/ccGLWindow.cpp
```diff
@@ -4248,6 +4256,14 @@ void ccGLWindow::mouseMoveEvent(QMouseEvent *event)
 			}
 		}
 	}
+	else if ((event->buttons() & Qt::MiddleButton)) // zoom
+	{
+		//middle button = zooming
+		float pseudo_wheelDelta_deg = -dy;
```
dgirardeau: Isn't this producing a warning about type conversion ? Can you maybe use static_cast<float>(-dy)?

dgirardeau:
Thanks for this contribution.

Can you also add a line about this new feature in the CHANGELOG.md file?

libs/qCC_glWindow/ccGLWindow.cpp
NeroBurner: did not warn for me (qtcreators clang-tidy and clazy warnings), but added it anyways

NeroBurner:
added to changelog

PR #970, opened 2019-10-15
riegl: add rdbio plugin

NeroBurner:
- support extensions rdbx, vxls, ptch and mtch
  - rdbx, vxls and mtch are read solely as pointcloud files
  - ptch files additionally create plane patches
- improved loading speed by preallocating memory, if allocation failes
  try to load as much as possible into memory (or until user aborts)
- add basic README with information where to get rdblib


dgirardeau:
Thanks for the contribution ;)

PR #971, opened 2019-10-15
Lib e57 plugin cleanups

NeroBurner:
update BUILD.md instructions on how to build with `PLUGIN_IO_QE57`
update instructions to use [libE57format](https://github.com/asmaloney/libE57Format) instead of reference implementation

also don't install the libE57 header and static library, those are not needed in the installation folder


NeroBurner:
only tested on Ubuntu 18.04 :see_no_evil:

dgirardeau:
Hopefully it will work on Windows as well ;)

PR #976, opened 2019-10-24
Segmentation support for ReferenceCloud

WargodHernandez:
I found incorrect results while using ManualSegmentationTools::segment on a reference cloud as segment adds the local index of the reference cloud instead of the global index. I separated the functions to avoid having an if check per point.

I also perform a dynamic_cast in segment on cloud to ReferenceCloud and if successful forward to the new function.

### DIFF
--- a/CC/src/ManualSegmentationTools.cpp
+++ b/CC/src/ManualSegmentationTools.cpp
@@ -57,7 +57,7 @@ ReferenceCloud* ManualSegmentationTools::segment(GenericIndexedCloudPersist* aCl
 		{
 			if (!Y->addPointIndex(i))
 			{
-				//not engouh memory
+				//not enough memory
 				delete Y;
 				Y = nullptr;
 				break;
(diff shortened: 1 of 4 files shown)


WargodHernandez:
The reason I want to segment reference clouds is to create an iterative best plane routine, starting with a larger cloud fitting a plane, segment out the major outliers and then create a new best fit plane
then repeat until the desired RMS value is obtained or some minimum point count is reached or a maximum iteration count is exceeded.

I was performing this with the original segmenting but having the point cloud and all scalar fields cloned for each iteration was annoyingly slow.

After I get the iterative plane fitting routine cleaned up I'll create another PR (right now it's working, but is not up to coding standards)

CC/src/ManualSegmentationTools.cpp
```diff
@@ -133,6 +133,39 @@ bool ManualSegmentationTools::isPointInsidePoly(const CCVector2& P,
...
+	//for each point
+	for (unsigned i = 0; i < cloud->size(); ++i)
+	{
+		const ScalarType dist = cloud->getPointScalarValue(i);
+		//we test if its associated scalar value falls inside the specified interval
+		if ((dist >= minDist && dist <= maxDist) ^ outside)
+		{
+			if (!Y->addPointIndex(cloud->getPointGlobalIndex(i)))
+			{
+				//not engouh memory
```
dgirardeau: enough?

PR #977, opened 2019-10-25
Added getEquation() planeEquation : [a, b, c, d] as 'ax+by+cz=d'

WargodHernandez:
Same equation used in Neighbourhood and DistanceComputationTools, also the same data type. 

Calculated in the buildUp function.

Simplifies using the DistanceComputationTools.

No changes to ccPlane Serialization


WargodHernandez:
I thought after those changes that updateRepresentation was called which then calls buildup, but yeah it looks like if plane->applyGLTransformation_recursive is called there isn't a call to buildup afterwords.

PR #982, opened 2019-11-05
Added computeCloud2SphereEquation

WargodHernandez:
computeCloud2SphereEquation takes sphere center and radius and sets all scalar values to distance.

Also removed Mainwindow from entity action and replaced with ccMainAppInterface

### DIFF
--- a/CC/include/DistanceComputationTools.h
+++ b/CC/include/DistanceComputationTools.h
@@ -271,7 +271,7 @@ public: //distance to simple entities (triangles, planes, etc.)
 		\param[out] rms will be set with the Root Mean Square (RMS) distance between a cloud and a plane (optional)
 		\return negative error code or a positive value in case of success
 	**/
-	static int computeCloud2SphereEquation(GenericIndexedCloudPersist *cloud, const CCVector3 SphereCenter, const PointCoordinateType SphereRadius, bool signedDistances = true, double* rms = nullptr);
+	static int computeCloud2SphereEquation(GenericIndexedCloudPersist *cloud, const CCVector3& SphereCenter, const PointCoordinateType SphereRadius, bool signedDistances = true, double* rms = nullptr);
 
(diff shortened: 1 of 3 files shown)


CC/include/DistanceComputationTools.h
```diff
@@ -263,6 +263,16 @@ class CC_CORE_LIB_API DistanceComputationTools : public CCToolbox
...
 
+	//! Computes the distance between each point in a cloud and a sphere
+	/** \param cloud a 3D point cloud
+		\param SphereCenter sphere 3d center point
+		\param SphereRadius sphere radius
+		\param signedDist whether to compute the signed or positive (absolute) distance (optional)
+		\param[out] rms will be set with the Root Mean Square (RMS) distance between a cloud and a plane (optional)
+		\return negative error code or a positive value in case of success
+	**/
+	static int computeCloud2SphereEquation(GenericIndexedCloudPersist *cloud, const CCVector3 SphereCenter, const PointCoordinateType SphereRadius, bool signedDistances = true, double* rms = nullptr);
```
dgirardeau: const CCVector3& SphereCenter (const ref)

CC/src/DistanceComputationTools.cpp
```diff
@@ -2150,6 +2150,38 @@ ScalarType DistanceComputationTools::computePoint2PlaneDistance(const CCVector3*
...
+		}
+		else
+		{
+			cloud->setPointScalarValue(i, static_cast<ScalarType>(std::abs(d)));
+		}
+		dSumSq += d * d;
+	}
+	if (rms)
+	{
+		*rms = dSumSq;
```
dgirardeau: don't you need to divide the square sum by count and then get the square root?

WargodHernandez: Yes! I missed that in the computeCloud2PlaneEquation also, I'll fix both locations

PR #987, opened 2019-11-05
Console stdout

### LINKED ISSUE #983: Console output
No output to console via printf when launching via command line. also built in console is just a spinner until the end of all parsing/processing.

**How to reproduce**

`cloudcompare.exe -silent -o cloud.bin -APPLY_TRANS CloudCompareMatrix.txt`
No output to stdout console will happen

**Expected behaviour**
should see all output via printf while -silent

**Your environment**
 - CC Version (or git hash/tag): #372e2fa5
 - OS & Version: Win10 pro
 - Graphics card: Nvidia Quadro M2200
 - Qt Version (if compiling):  5.11.2

WargodHernandez:
Fixes #983 on windows command line mode. (allows printf)

Also adds the console redirection to ccConsole which allows us to remove the specialty ccCommandLineParser extensions

```C++
ccCommandLineParser::print(const QString& message) const{}
ccCommandLineParser::warning(const QString& message) const{}
ccCommandLineParser::error(const QString& message) const{}
```
I left these in but removed the calls to printf so they are just forwarding directly to ccConsole.

### DIFF
--- a/qCC/ccCommandLineParser.cpp
+++ b/qCC/ccCommandLineParser.cpp
@@ -38,28 +38,18 @@ constexpr char COMMAND_SILENT_MODE[]	= "SILENT";
 void ccCommandLineParser::print(const QString& message) const
 {
 	ccConsole::Print(message);
-	if (m_silentMode)
-	{
-		printf("%s\n", qPrintable(message));
-	}
 }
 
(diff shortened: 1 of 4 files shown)


dgirardeau:
Thanks for the contribution!

PR #994, opened 2019-11-08
Restore Color and SF state after ICP

### LINKED ISSUE #991: point cloud color gets reset after fine registration (ICP)
**Describe the bug**
After using the "Finely register already roughly aligned entities" button to register a point cloud to a reference, the color of the moved cloud gets reset to 'Scalar Field' if it was originally RGB.  This becomes an issue when I'm slicing up a point cloud and registering it piecemeal, since I have to reset it for every piece.  I can't just reset it all at the end, either, since I'm using the colors to judge whether a piece is registered correctly.

**How to reproduce**
1. Have two point clouds, A and B, that contain scalar fields.
2. Change the color of B: edit -> colors -> set unique -> OK
3. select both point clouds and use ICP (tools->registration->fine registration)
4. B gets matched to A, but also B's color changes to 'use scalar field'.

**Expected behaviour**
Register without the change in color.

**Your environment**
 - CC Version (or git hash/tag): 2.10.2
 - OS & Version: Windows 10 64bit
 - Graphics card: Nvidia Titan XP

WargodHernandez:
Fixes #991 

And casts finalOverlapRatio to double to avoid compiler Warning C26451: Arithmetic overflow not that it would happen but it shuts up visual studios static analyzer :)

### DIFF
--- a/qCC/ccRegistrationTools.cpp
+++ b/qCC/ccRegistrationTools.cpp
@@ -62,8 +62,8 @@ bool ccRegistrationTools::ICP(	ccHObject* data,
 								int maxThreadCount/*=0*/,
 								QWidget* parent/*=0*/)
 {
-	bool RestoreColorState = false;
-	bool RestoreSFState = false;
+	bool restoreColorState = false;
+	bool restoreSFState = false;
 
 	//progress bar
 	QScopedPointer<ccProgressDialog> progressDlg;
@@ -112,8 +112,8 @@ bool ccRegistrationTools::ICP(	ccHObject* data,
 	if (data->isA(CC_TYPES::POINT_CLOUD))
 	{
 		ccPointCloud* pc = static_cast<ccPointCloud*>(data);
-		RestoreColorState = pc->colorsShown();
-		RestoreSFState = pc->sfShown();
+		restoreColorState = pc->colorsShown();
+		restoreSFState = pc->sfShown();
 		dataDisplayedSF = pc->getCurrentDisplayedScalarField();
 		oldDataSfIdx = pc->getCurrentInScalarFieldIndex();
 		dataSfIdx = pc->getScalarFieldIndexByName(REGISTRATION_DISTS_SF);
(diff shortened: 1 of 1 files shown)


qCC/ccRegistrationTools.cpp
```diff
@@ -62,6 +62,9 @@ bool ccRegistrationTools::ICP(	ccHObject* data,
 								int maxThreadCount/*=0*/,
 								QWidget* parent/*=0*/)
 {
+	bool RestoreColorState = false;
```
dgirardeau: lower case first?

qCC/ccRegistrationTools.cpp
```diff
@@ -194,7 +199,7 @@ bool ccRegistrationTools::ICP(	ccHObject* data,
 			ParallelSort(distances.begin(), distances.end());
 			
 			//now look for the max value at 'finalOverlapRatio+margin' percent
-			maxSearchDist = distances[static_cast<unsigned>(std::max(1.0,count*(finalOverlapRatio+s_overlapMarginRatio)))-1];
+			maxSearchDist = distances[static_cast<unsigned>(std::max(1.0,count*(static_cast<double>(finalOverlapRatio)+s_overlapMarginRatio)))-1];
```
dgirardeau: Isn't finalOverlapRatio already a double value?

WargodHernandez: yes I was thrown off by the squigles going away for the static code analyzer it takes a moment after saving before showing any warnings, but the (Warning C26451: Arithmetic overflow) was actually for the static_cast<unsigned>
removing my first attempt, and changing to  

``` C++
maxSearchDist = distances[static_cast<size_t>(std::max(1.0,count*(finalOverlapRatio+s_overlapMarginRatio)))-1];
```
cleared it up. 
unsigned = 4 bytes, size_t = 8 bytes

Is there any reason to not use size_t in this case?

WargodHernandez: just started using VS2019 and I'm getting a ton of static analysis warnings that weren't present in VS2017 especially everything QT related 
image

there is an open issue for [VS2019](https://developercommunity.visualstudio.com/content/problem/520393/vs-2019-detects-many-false-warnings-while-vs2017-w.html) 

but outside of the QT noise it does pickup some seemingly legit risks
image

dgirardeau: Originally CC was 32bits only (hence the unsigned everywhere) and we also wanted to keep point indexes on 32bits so as to minimize the memory consumption (so unsigned became the default integer type I would say). 

Here I assume that count is an unsigned? At least the result of the calculus can't be greater than 'count' but the compiler/Visual can't be sure and raise the warning by default I guess.

dgirardeau: The next example is indeed interesting. Here it's more about the loss of numerical accuracy 'for nothing'. The float values resulting from the subtractions can definitely be casted to double.

PR #999, opened 2019-11-12
Qt5 signal and slot update

WargodHernandez:
Also in ccComparison moved the processEvents and compute distances into a singleshot timer lambda to remove them from the constructor

### DIFF
--- a/common/ccCameraParamEditDlg.cpp
+++ b/common/ccCameraParamEditDlg.cpp
@@ -241,12 +241,12 @@ void ccCameraParamEditDlg::pickPointAsPivot(bool state)
 		if (state)
 		{
 			m_associatedWin->setPickingMode(ccGLWindow::POINT_OR_TRIANGLE_PICKING);
-			connect(m_associatedWin, SIGNAL(itemPicked(ccHObject*, unsigned, int, int, const CCVector3&, const CCVector3d&)), this, SLOT(processPickedItem(ccHObject*, unsigned, int, int, const CCVector3&, const CCVector3d&)));
+			connect(m_associatedWin, &ccGLWindow::itemPicked, this, &ccCameraParamEditDlg::processPickedItem);
 		}
 		else
 		{
 			m_associatedWin->setPickingMode(ccGLWindow::DEFAULT_PICKING);
-			disconnect(m_associatedWin, SIGNAL(itemPicked(ccHObject*, unsigned, int, int, const CCVector3&, const CCVector3d&)), this, SLOT(processPickedItem(ccHObject*, unsigned, int, int, const CCVector3&, const CCVector3d&)));
+			connect(m_associatedWin, &ccGLWindow::itemPicked, this, &ccCameraParamEditDlg::processPickedItem);
 		}
 	}
 
--- a/common/ccCameraParamEditDlg.cpp
+++ b/common/ccCameraParamEditDlg.cpp
@@ -246,7 +246,7 @@ void ccCameraParamEditDlg::pickPointAsPivot(bool state)
 		else
 		{
 			m_associatedWin->setPickingMode(ccGLWindow::DEFAULT_PICKING);
-			connect(m_associatedWin, &ccGLWindow::itemPicked, this, &ccCameraParamEditDlg::processPickedItem);
+			disconnect(m_associatedWin, &ccGLWindow::itemPicked, this, &ccCameraParamEditDlg::processPickedItem);
 		}
 	}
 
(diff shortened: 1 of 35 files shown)


qCC/ccComparisonDlg.cpp
```diff
@@ -129,24 +129,20 @@ ccComparisonDlg::ccComparisonDlg(	ccHObject* compEntity,
...
 	//(otherwise the progress bars appear anywhere!)
-	if (!m_noDisplay)
-	{
-		show();
-		QCoreApplication::processEvents();
-	}
-
-	//compute approximate results and unlock GUI
-	computeApproxDistances();
+	QTimer::singleShot(0, this, [=]() { //singleshot 0 time will fire once after all events in the queue are processed
```
dgirardeau: Interesting solution

dgirardeau: Do we still need to call 'processEvents' then?

WargodHernandez: I'm not totally sure, I only tested removing show(); and removing the check for m_noDisplay (each area this is used calls show like this)

```C++
m_compDlg = new ccComparisonDlg(compCloud, refCloud, ccComparisonDlg::CLOUDCLOUD_DIST, this);
	connect(m_compDlg, &QDialog::finished, this, &MainWindow::deactivateComparisonMode);
	m_compDlg->show();
```
I found that after removing the show function I had to add the single shot to avoid the progress bars issue. I'll run a quick test to see if it makes any difference for me

WargodHernandez: I couldn't tell any difference. with or without 'processEvents' so I believe it is no longer required

common/ccCameraParamEditDlg.cpp
```diff
@@ -241,12 +241,12 @@ void ccCameraParamEditDlg::pickPointAsPivot(bool state)
...
 		{
 			m_associatedWin->setPickingMode(ccGLWindow::POINT_OR_TRIANGLE_PICKING);
-			connect(m_associatedWin, SIGNAL(itemPicked(ccHObject*, unsigned, int, int, const CCVector3&, const CCVector3d&)), this, SLOT(processPickedItem(ccHObject*, unsigned, int, int, const CCVector3&, const CCVector3d&)));
+			connect(m_associatedWin, &ccGLWindow::itemPicked, this, &ccCameraParamEditDlg::processPickedItem);
 		}
 		else
 		{
 			m_associatedWin->setPickingMode(ccGLWindow::DEFAULT_PICKING);
-			disconnect(m_associatedWin, SIGNAL(itemPicked(ccHObject*, unsigned, int, int, const CCVector3&, const CCVector3d&)), this, SLOT(processPickedItem(ccHObject*, unsigned, int, int, const CCVector3&, const CCVector3d&)));
+			connect(m_associatedWin, &ccGLWindow::itemPicked, this, &ccCameraParamEditDlg::processPickedItem);
```
asmaloney: This is a `disconnect`.

WargodHernandez: yep thank you 
I copy/pasted from the connect above Whoops

WargodHernandez:
Hey @dgirardeau & @asmaloney  this is all of the changes required for the main projects including ccViewer to use QT5 style commits the only remaining items are in the plugins. 
Unless there are any fixes or code review items required I will not be pushing anything further to this PR.

WargodHernandez:
The build check failing appears to be a timeout during a docker update before it even reached the before_script section, I'm not sure how to get travis to retry [failed travis log](https://travis-ci.org/CloudCompare/CloudCompare/jobs/611532296?utm_medium=notification&utm_source=github_status)

WargodHernandez:
> 
> 
> It seems that if you go the Travis build log page, and you sign in, you can restart the build. Let's see wha t happens.

Looks like success this time, Thanks

dgirardeau:
Ok I couldn't spot any particular issue, but maybe I'll wait for @asmaloney feedback as he has pretty sharp eyes for this!

dgirardeau:
Sorry for the delay. And thanks for the contribution ;)

asmaloney:
I apologize - I've been on other things and missed the ping from Daniel.

A couple of notes for future PRs:

- For third-party code like  qcustomplot there's not much point in modifying it because we will likely update it at some point. (`qcustomplot` should actually live in `contrib`, shouldn't it?)
- Smaller PRs are better. Anything that can be reviewed in a couple of minutes is much easier to deal with. Generally, once you're touching more than 4 or 5 files it's probably too large.
- In general, please don't mix in extra surprises like the casting changes. Each PR should only have one objective. (In this case, the casting is not part of "Qt5 signal and slot update".) This makes PRs easier to understand/review and makes the resulting commits "cherry-pickable" to other branches.

(Note that I use "in general" a lot there - obviously there are exceptions :-))

Thanks for all of that grunt work @WargodHernandez!

PR #1005, opened 2019-11-18
Cylinder primitive distance equation

WargodHernandez:
This adds cylinder to the list of primitive calculation types, this gives a different result than the cloud2mesh function with gives odd artifacts for points lying on the cylinders projected radius
image


WargodHernandez:
I started all three PRs today from master and I just noticed the conflict in DistanceComputationTools because of the earlier PR. sorry about that

WargodHernandez:
@dgirardeau 
Also in regards to the artifacts in Cloud2Mesh when applied to a cylinder, after testing it appears similar artifacts happen for cubes (points inline with edges),

image

dgirardeau:
Interesting! I should take a look at that (even though I'm not surprised there are limit cases du to numerical accuracy issues on 'borders'). Especially with very small coordinates?

(the PR still seems in conflict)

WargodHernandez:
@dgirardeau 
> (the PR still seems in conflict)

Yeah I wasn't sure the proper way to resolve that, if I was resolving conflicts on a pull visual studio lets me select which "change" to accept but the resolve conflicts button on github just shows a static page with the conflicts.

I can make another commit to this PR I just wasn't sure if that would be a weird commit history having the same change in 2 commits.

Let me know what you want and I'll do it.

Thanks for the support!

dgirardeau:
Can you perform a 'rebase' on 'master'?

WargodHernandez:
@dgirardeau 
> Interesting! I should take a look at that (even though I'm not surprised there are limit cases du to numerical accuracy issues on 'borders'). Especially with very small coordinates?
>
I'm not sure about the very small coordinates statement, the effect is easier to reproduce with larger coordinates, the Stanford bunny had a small bounding box ~x:0.15 y:0.15 z:0.12 and it took a few tries to get the artifacts.
where the sensor data I was working with was ~x:7.0 y:9.12 z:91.11

I was able to see similar artifacts when measuring against an STL which I know was manifold when exported from Catia (See the X block images below)

I suspect that in certain regions the distance is incorrectly set as a signed value rather than absolute, because if I use unsigned calculations I get reasonable results that don't have any discontinuities along those same projected regions.


image
image

image
image

dgirardeau:
I believe the artifacts appear generally when there is a "doubt" on which is the closest triangle. If the algorithm considers that the closest is the triangle next to the "true" one (because of very small numerical inaccuracies), then it choose the other triangle as reference, and the other triangle normal to sign the distance value. Here, it's particularly visible on sharp edges, where the normal has nothing in common with the neighbor triangle. And it gives a totally different sign...

PR #1007, opened 2019-11-23
computeCloud2ConeEquation and support in the CloudPrimitiveDist function

WargodHernandez:
Added support for cone Primitives, snout mode is not supported.

Also added a couple helper methods to ccCone to simplify grabbing cone axis end points and radii (the algorithm expects larger radii first and adding the helper method seemed a lot cleaner than adding it to the distance function)

again not surprising but here is a comparison between the cloud2Mesh and cloud2Primitive with only negative values shown
image
image

### DIFF
--- a/CC/include/DistanceComputationTools.h
+++ b/CC/include/DistanceComputationTools.h
@@ -263,6 +263,19 @@ public: //distance to simple entities (triangles, planes, etc.)
 	**/
 	static ScalarType computePoint2PlaneDistance(const CCVector3* P, const PointCoordinateType* planeEquation);
 
+	//! Computes the distance between each point in a cloud and a cone
+	/** \param cloud a 3D point cloud
+		\param coneP1 center point associated with the larger radii
+		\param coneP2 center point associated with the smaller radii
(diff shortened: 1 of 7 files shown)


dgirardeau:
Nice! Can you update the readme.md file as well?

WargodHernandez:
@dgirardeau Thanks for the reminder, I updated the change log as well as the UI dialog for both cones and cylinders. 

I know I want to add a couple more primitive calculations to future PRs. So hopefully these PRs haven't been too much work for you to review.

My list of what I want to add is currently 
1. Bounded plane
2. AA Cube
3. Cube
4. Infinite line
5. Line segment
6. Poly-line
7. Line swept sphere (Needs CC primitive support first)
8. Cone-sphere (Needs CC primitive support first)

dgirardeau:
No problem. If you support all the primitives in the future, then my only comment will be for the MainWindow::doActionCloudPrimitiveDist method: maybe use a switch / case statement instead of trying to convert the input entity to all possible types? But will see that during the next PRs ;)

PR #1015, opened 2019-12-20
Add cloud to rectangle and cloud to box primitive distance calculations

smortvedt:
Added the option to treat planes as bounded planes (rectangles) when performing the distance calculation, also added functionality to calculate the distance to boxes using the geometric definition. Updated the dialog ui form and the changelog.

PR #1020, opened 2020-01-14
Add cloud2Polyline primitive distance calculation

smortvedt:
Similar to my last pull request, this pull request adds the distance to a polyline primitive calculation.  Note: Chris is working on the changes requested from #1015 specifically the error code enum and the assignments in the if statement.

### DIFF
--- a/CC/src/DistanceComputationTools.cpp
+++ b/CC/src/DistanceComputationTools.cpp
@@ -2158,7 +2158,11 @@ ScalarType DistanceComputationTools::computePoint2LineSegmentDistSquared(const C
 	CCVector3 vec;
 	ScalarType distSq;
 	PointCoordinateType t = line.dot(*p - *start);
-	t /= line.norm2();
+	PointCoordinateType normSq = line.norm2();
+	if (normSq != 0) 
+	{
+		t /= normSq;
+	}
 	if (t < 0)
 	{
 		vec = *p - *start;
--- a/CC/src/DistanceComputationTools.cpp
+++ b/CC/src/DistanceComputationTools.cpp
@@ -2723,7 +2723,7 @@ int DistanceComputationTools::computeCloud2PolylineEquation(GenericIndexedCloudP
 	ScalarType dSumSq = 0;
 	for (unsigned i = 0; i < count; ++i)
 	{
-		ScalarType distSq = INFINITY;
+		ScalarType distSq = NAN_VALUE;
 		const CCVector3* p = cloud->getPoint(i);
 		for (unsigned j = 0; j < polyline->size() - 1; j++)
 		{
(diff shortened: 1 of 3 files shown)


CC/src/DistanceComputationTools.cpp
```diff
@@ -2677,6 +2703,66 @@ int DistanceComputationTools::computeCloud2BoxEquation(GenericIndexedCloudPersis
...
+	}
+	if (!cloud->enableScalarField())
+	{
+		return -3;
+	}
+	ScalarType d = 0;
+	ScalarType dSumSq = 0;
+	for (unsigned i = 0; i < count; ++i)
+	{
+		ScalarType distSq = INFINITY;
```
smortvedt: I wasn't sure what to use for infinity here

dgirardeau: We normally use 'NAN_VALUE' (and you can use std::isnan to test it)

WargodHernandez: I would argue that using NAN_VALUE is probably not the best choice here as the later use
```C++
distSq = std::min(distSq, computePoint2LineSegmentDistSquared(p, start, end));
```
Would become more complicated than required.

would it be acceptable to add the following to CCConst.h

```C++
//! INF as a ScalarType value
/** \warning: handle with care! **/
constexpr ScalarType INF_VALUE = std::numeric_limits<ScalarType>::infinity();

```
and then set any INF_VALUE at the end to NAN_VALUE if required outside of the inner loop?

if not then I would suggest we set the value to the first loop iteration value like:

```C++
ScalarType distSq = computePoint2LineSegmentDistSquared(p, 
							polyline->getPoint(0), 
							polyline->getPoint(1));
```

@dgirardeau what are your thoughts, is INF_VALUE a risk in other areas I'm not considering? Am I missing something that wouldn't require extra logic in the inner loop?

dgirardeau: Oh if it's just a temporary value, I definitely prefer to init the distance the first time with the right value. The infinity value wouldn't be properly handled by the other components of CloudCompare (especially at display time, etc.).

qCC/mainwindow.cpp
```diff
@@ -8997,6 +9003,17 @@ void MainWindow::doActionCloudPrimitiveDist()
 						ccConsole::Error(errString, "Box", returnCode);
 					break; 
 				}
+				case CC_TYPES::POLY_LINE:
+				{
+					signedDist = false;
```
WargodHernandez: Also need 
```C++
flippedNormals = false;
```

CC/src/DistanceComputationTools.cpp
```diff
@@ -2151,6 +2152,31 @@ ScalarType DistanceComputationTools::computePoint2PlaneDistance(const CCVector3*
 	return static_cast<ScalarType>((CCVector3::vdot(P->u, planeEquation) - planeEquation[3])/*/CCVector3::vnorm(planeEquation)*/); //norm == 1.0!
 }
 
+ScalarType DistanceComputationTools::computePoint2LineSegmentDistSquared(const CCVector3* p, const CCVector3* start, const CCVector3* end)
+{
+	CCVector3 line = *end - *start;
+	CCVector3 vec;
+	ScalarType distSq;
+	PointCoordinateType t = line.dot(*p - *start);
+	t /= line.norm2();
```
dgirardeau: maybe test the norm before using it as quotient? To avoid division by zero?

WargodHernandez:
@dgirardeau Is there anything left you want fixed on this?

dgirardeau:
Oh sorry, I was giving some times for others (well you or Andy ;) to make more comments.

dgirardeau:
And thanks for the contribution ;)

PR #1021, opened 2020-01-16
Minor distance related improvements

WargodHernandez:
To fix the clang warning about using abs, went through and set all uses of abs to std::abs

image

In computePoint2TriangleDistance I cast to double prior to subtraction, I found a slight improvement in the number of outliers when measuring with Cloud2Mesh after this change.

Finally I noticed the primitive distance function was being enabled when just a polyline was being selected. so I added parenthesis to the setEnabled check.


dgirardeau:
Nice catch!

PR #1022, opened 2020-01-20
Distance computation results enum

WargodHernandez:
As requested in #1015 added a distance computation results enum.
To not break any existing code, the functions still return int, but wherever possible the value of the int will be set to the enum.

Added a static int s_cellFunc_MT_results which is set to results enum, this supports using the results enum in the multi threaded functions.

Added a few missing error checks typically around some of the octree related "get" functions like getCellCodes, getPointsInCellByCellIndex, and getCellCodesAndIndexes

Cleaned up some code styling, missing spaces and braces primarily.


CC/src/DistanceComputationTools.cpp
```diff
@@ -2328,7 +2450,7 @@ int DistanceComputationTools::computeCloud2ConeEquation(GenericIndexedCloudPersi
 	{
 		*rms = sqrt(dSumSq / count);
 	}
-	return count;
+	return DISTANCE_COMPUTATION_RESULTS::SUCCESS;
```
asmaloney: Was this intended? Is anything expecting this to return a count?

This change occurs several times below.

CC/src/DistanceComputationTools.cpp
```diff
@@ -2814,16 +2979,24 @@ ScalarType DistanceComputationTools::ComputeCloud2PlaneRobustMax(	GenericCloud*
     assert(cloud && planeEquation);
 	assert(percent < 1.0f);
 
+	if (!cloud)
+	{
+		return 0;
```
asmaloney: Should these "`return 0`"s use the enum?

(This one and below.)

CC/src/DistanceComputationTools.cpp
WargodHernandez: No, I wrote the primitive distance functions originally and just intended to give a positive success, these are only used in the `doActionCloudPrimitiveDist` function in mainwindow.

CC/src/DistanceComputationTools.cpp
WargodHernandez: yes 0 was intentional the various plane distance/error functions which return ScalarType are used in various places and I didn't want any unintended consequences. This should happen in 
```
computeCloud2PlaneDistanceRMS
ComputeCloud2PlaneRobustMax
ComputeCloud2PlaneMaxDistance
ComputeCloud2PlaneDistance
```

WargodHernandez: Looking into some of where these get used I would support making changes to them in the future (to keep the PR reasonable).
here is one example of how these are used, this is from `fastMarchingForFacetExtraction::computeTCoefApprox`
```C++
ScalarType reprojError = CCLib::DistanceComputationTools::ComputeCloud2PlaneDistance(&Yk, theLSQPlaneEquation, m_errorMeasure);
			if (reprojError >= 0)
				return (1.0f - orientationConfidence) * static_cast<float>(reprojError); 
 
```
This looks to like a bug to me as both a perfect projection and an error will give the same result

CC/src/DistanceComputationTools.cpp
WargodHernandez: by No I mean: No nothing is expecting this to return a count.
Yes this was intentional.

PR #1023, opened 2020-01-22
Scripting plugin WIP

WargodHernandez:
Plugin does run and successfully evaluates script.
Long road to go.

See [ChaiScript](http://chaiscript.com/index.html) for a language overview. Version 5.8.6 is the latest version supporting C++11.

This is a work in Progress and no where near ready for... well anything, but it is a start.
I wanted to have this as a WIP item on the main repo so the main CC team can give any feed back as I'm working through it.


PowerHydrant:
I applaud the effort as a heavy user of command line mode through .bat files and unix shell scripts. 

1. Please include the equivalent of silent mode and the selective capturing of console output to callbacks or other mechanism for consumption by another process.
2. Please allow (as appropriate to ChaiScript) for scripting and script calling from another process
3. It would be nice CC to be accessible as a "server" via scripts interacting with a "Client" process in both "silent" and "Non-silent" mode
4. We find high value in CC as an integrated cohesive system 

Thanks very much

WargodHernandez:
My intent is to add command line mode, my thought was two options, 1 take in a script as a file path, 2 take in a script directly on the command line. 
The first option should be pretty easy in comparison to the 2nd.

As far as integration of CloudCompare functionality cohesively, this was specifically why I chose ChaiScript. ChaiScript was designed for native integration with C++, it has support for call backs, creating functions, calling functions (with overloading resolution), creating variables, creating new types and derived types. It has c++ exception support (you can use try/catch blocks in the script and catch exceptions thrown in c++ functions called by the script)

As far as integrating with external callbacks outside of both script or c++ should be possible. As the scripts can call any function that c++ can, so passing an external function pointer into the script in theory could work.

This morning I've started on the code editor, my plan is to have project files with single/multiple code files integrated. Along with syntax highlighting (there is already an existing QT highlighter for ChaiScript) and line numbers. I'll be basing the editor on QT code editor example and the preexisting syntax highlighter, so there is a solid framework that I should be able to integrate it pretty quickly.

WargodHernandez:
Here is an example script showing the C++/ChaiScript interaction
image
image


And here I'm showing the chaiScript syntax, paired character, and currentline highlighting. Also the line numbers are displayed.
image

WargodHernandez:
I have begun adding in the CCLib::CCGeom types and testing as I go, I can create CCVector(both 2d and 3d, using templates so all the same types available in c++ work) in script and pass them to CC functions expecting those types. 

image

Hey @dgirardeau is this plugin something that after its actually useful will be accepted into CC? and if so do you want to comment at all on any issues with the way I'm going about this. It'll be easier to make the changes sooner rather than later :)

dgirardeau:
Yes of course. That sounds interesting (and a lot of people are interested by scripting).

For the review, I'll try, but I'll be in the US this coming week so I might not have a lot of time ;)

WargodHernandez:
Thanks @dgirardeau I am mostly at this point interested in the teams thoughts on the organization of the plugin, folder structure and class hierarchy. It is still far too early to expect a detailed review (at least in my opinion). I have scheduled the next 20 workdays for this plugin AKA the next 4 weeks, so there will be time for in depth reviews later.

I also see a few places where some changes to the main code base and ccMainAppInterface could be useful, and shouldn't hurt anything, so I will create those in stand-alone pull requests before relying on the functionality in the plugin.

One last area that I could see wanting to expand is the plugin interface, I'd like to expand it to include a method for querying the plugin for classes/function/variables that the plugin wants to expose to the scripting engine, it would be an optional portion of the interface so the plugin authors can determine whether they want to include support or not (perhaps something similar to how the command line interface is up to the plugin author). I just need to figure out if it is possible to do that without including the scripting engine headers into the plugin interface itself (or whether or not that would be acceptable)

WargodHernandez:
image

The plugin is almost somewhat useful, at least it can do something that there currently is no method in the main CC for a user to do.

WargodHernandez:
Hey @dgirardeau have you had any chance to look at this at all?


Here is todays new status.

The plugin can now execute chai scripts from the command line, and even add new commands to the command line that will be available to the cloud compare command line parser

in this example I create a new commandline command in the "chaicommandlineext.chai" and then let cloud compare execute it with -Chai_command the functions inside and the name and keyword are freely assignable in script.

CloudCompare.exe -silent -CHAI "chaicommandlineext.chai" **-Chai_command**

image


image

dgirardeau:
Yes I looked at your previous commit and I was already very interested in the ability to load simple structures (points, polylines, planes, etc.) from custom text files.

And the fact that you can even use it via the command line makes it even more interesting.

And now the big question: how to explain properly how this works to users, and provide meaningful examples, store useful scripts, etc.?

I can of course give you a wiki account so that you can create a dedicated page. We can also of course have a dedicated section on the forum, etc.

I'll definitely advertise this new feature during the Users workshop in Paris (Palaiseau) next month!

WargodHernandez:
Great, I'm pretty happy with how it's shaping up, I am reasonably confident in how the code that exposes the CC C++ to ChaiScript is done, it is essentially the same style as QT5 connects.

I am less confident on some of the code in ChaiScriptingPlugin.h and .cpp, I have to do a few clunky things to try to make sure the ChaiScript engine is destroyed prior to the application closing. (On windows if the app terminates prior to destroying ChaiScript there will be an unhandled exception due to windows destroying thread local storage prior to ChaiScript). 
I can disable multithread support for ChaiScript and the issue goes away, but I would like to maintain MT in case it makes sense to run scripts in separate threads. maybe if we can expose a signal in CC for pre termination, something just to give any system or plugin a chance to properly release resources.
 
I thought I had it taken care of prior to implementing the command line, in normal circumstances (or if a script error happens) it works fine and takes care of itself, but if the command line parser runs into an unsupported command or another command returns false, there will be a crash because I don't have a way to know to kill ChaiScript.

asmaloney:
There's a [QCoreApplication::aboutToQuit()](https://doc.qt.io/qt-5/qcoreapplication.html#aboutToQuit) signal.

Would that working timing-wise or is that too late?

WargodHernandez:
> There's a QCoreApplication::aboutToQuit() signal.
> Would that working timing-wise or is that too late?

I didn't know about that signal... I will try it out. 
I had just noticed the plugin release in main.cpp line 267-271
I moved that outside of the if statement and added try catch per plugin 
```C++
//release the plugins
	for ( ccPluginInterface *plugin : ccPluginManager::get().pluginList())
	{
		try
		{
			plugin->stop(); //just in case
		}
		catch (const std::exception & e)
		{
			QMessageBox::warning(0, QString("%1 - crashed!").arg(plugin->getName()), QString("%1 - crashed while attempting to stop\n%2").arg(plugin->getName()).arg(e.what()));
		}
		catch (...)
		{
			QMessageBox::warning(0, QString("%1 - crashed!").arg(plugin->getName()), QString("%1 - crashed while attempting to stop\n").arg(plugin->getName()));
		}
	}
```

I'll attempt the QCoreApplication::aboutToQuit() signal first though I think that would probably be the cleanest.

WargodHernandez:
@asmaloney the QCoreApplication::aboutToQuit() signal appears like it is fired too late, I still could cause crashes when using it.

but this worked for me
```C++
qAddPostRoutine([]() {ChaiScriptingPlugin::TheInstance()->destroyChai(); });

```

WargodHernandez:
@asmaloney @rjanvier 

Hey BTW since ChaiScript is cross platform, I figured I should follow the various platforms for folder search paths for scripts and modules. 

Modules are just shared libraries with a known entry point for loading ChaiScript modules.

While the scripts location should be where the user stores whatever scripts they write and want to include in other scripts.
 
Do either of you have recommendations for default paths for those two locations?

asmaloney:
The "correct" place would probably be folders in [QStandardPaths::AppDataLocation](https://doc.qt.io/qt-5/qstandardpaths.html), but it's not necessarily the most convenient place for user-generated scripts...

WargodHernandez:
I was thinking I would add a configuration menu that let the user set additional search paths and save them to the QSettings for this plugin so they should be able to make it convenient for themselves... but I am trying to at least start with normal platform expectations

WargodHernandez:
@dgirardeau @asmaloney 
What is the groups opinion on how to structure ChaiScript modules (the modules are just shared libraries which implement the following function to load some C++ into ChaiScript:
```C++
CHAISCRIPT_MODULE_EXPORT  chaiscript::ModulePtr create_chaiscript_module_test_module()
{
chaiscript::ModulePtr m(new chaiscript::Module());

  m->add(chaiscript::fun([](){return "hello world";}), "hello_world");
return m;
}
```

I've been testing with a subfolder inside of the plugins directory, this is working and doesn't cause any issues with the CC plugin manager as it isn't searching recursively for plugins.

I'm struggling with the decision because these are essentially plugins to a plugin... It feels like they shouldn't be off by themselves.

Also for an example of a use case, I am planning on making a module to expose QTNetworking to ChaiScript. so a person could create a script like this

```JS
load_module("QTNetwork_module")
var dlm = DownloadManager()
dlm.downloadFinished = fun(x)
   {
   //Have CC do something with whatever downloaded
   // x stores the network reply in this example
   )
dlm.doDownload("someURLhere");
```

This way there is no baked in networking unless requested by a user script.

asmaloney:
> It feels like they shouldn't be off by themselves.

I agree with this. In my opinion, if these are modules that you are shipping as part of the plugin's default installation, they should be included where the plugin is installed in a subfolder.

The tricky bit is third party (fourth party?) modules - if you want users to add their own. I don't believe users should be putting anything in any of the CC installation locations, so I think these belong somewhere in QStandardPaths::AppDataLocation as mentioned above. Having the user customize this path is also an option, but this adds more complexity and may increase the support burden.

WargodHernandez:
> The tricky bit is third party (fourth party?) modules - if you want users to add their own. I don't believe users should be putting anything in any of the CC installation locations, so I think these belong somewhere in QStandardPaths::AppDataLocation as mentioned above. Having the user customize this path is also an option, but this adds more complexity and may increase the support burden.

Well user created Modules can load from anywhere, I provide a std::vector<std::string> folder locations to search for modules, and another for scripts. so the user will always be able to add another directory they if they want, but I feel like some a number of useful modules could be provided by default, also I am thinking of moving all of the CloudCompare types to their own module(s). so the base plugin is just about providing the interface and the script editor and commandline execution. 
Should make compilation when just changing the base plugin faster rather than the C++/ChaiScript interface which does add around a minute of extra compilation for me (chaiscript compilation has lots of templating going on).

PowerHydrant:
Thanks for the work done in this domain. I hope I am not duplicating anyone else's input and I know @WargodHernandez made reference to command line mode in earlier comments. I hope/presume that command line mode with -SILENT invocation and chaiscript sscripting will be supported. Thanks

WargodHernandez:
Well I will be respecting -SILENT but I can't stop some code called by a script from creating dialogs. But the scripting environment will be aware of whether it is in command line mode, and whether or not -SILENT is active

plugins/3rdParty/qChaiScripting/ChaiScriptingPlugin.cpp
```diff
@@ -0,0 +1,409 @@
...
+}
+
+void ChaiScriptingPlugin::onNewSelection(const ccHObject::Container& selectedEntities)
+{
+	if (chai_onNewSelection)
+	{
+		std::vector<std::shared_ptr<ccHObject>> sel;
+		try
+		{
+			for each (auto objectName in selectedEntities)
```
tmontaigu: This doesn't look like standard C++ 😧

plugins/3rdParty/qChaiScripting/codeEditor/codeEditor.cpp
```diff
@@ -0,0 +1,546 @@
...
+//#                     COPYRIGHT: Chris S Brown                           #
+//#                                                                        #
+//##########################################################################
+
+
+#include <QtWidgets>
+#include <QPlainTextEdit>
+
+
+#include "codeeditor.h"
```
tmontaigu: Linux is case sensitive so it must be `#include "codeEditor.h"`

plugins/3rdParty/qChaiScripting/ChaiScriptingPlugin.cpp
WargodHernandez: yeah I'm living in MSVC land :( I'll get this patched up

plugins/3rdParty/qChaiScripting/codeEditor/codeEditor.cpp
WargodHernandez: yeah I'm living in MSVC land :( I'll get this patched up

dgirardeau:
Hi @WargodHernandez, long time no hear about you! Hope you are doing well.

What do you think we should do with this PR?

WargodHernandez:
> Hi @WargodHernandez, long time no hear about you! Hope you are doing well.
> 
> What do you think we should do with this PR?

Hey @dgirardeau yeah it’s been a long time, I started over at SpaceX and have barely had time to keep my sanity let alone up to date with CC, although I do still read most of the GitHub PRs :) 
 
 I sadly believe this should probably just be closed, it obviously hasn’t kept up with 2.13 or anything over this past year. Plus with @tmontaigu python integration I think more people would be served by that pairing vs having to learn chaiscript.

The biggest problem with this plugin is how tightly its coupled to C++, having to expose so much and keep all the signatures up to date. It’s a long term maintenance nightmare :(

dgirardeau:
Thanks Chris. I hope you have fun at SpaceX!

PR #1028, opened 2020-02-03
Classic snap confinement instead of strict

balazsdukai:
Proposal to use classic confinement instead of strict. If the data source directory resides outside of the user's /home, its not possible to access it when using strict confinement.


dgirardeau:
Seems reasonable ;)

PR #1030, opened 2020-02-04
PLY: Use an error callback to give more feed back on errors

### LINKED ISSUE #276: Issue more detailed error messages when reading a PLY file (= rPly hack)
If  PLY file conatins a float property that is too large, it gives the error:
[11:23:13] An error occurred while loading 'SurfacePoints_test': reading error (no access right?)

Couldn't attach the file for some reason, so here is a pastebin: http://pastebin.com/1JgWwCej (the last value on the first line makes it break) ( -9.25596e+61 )

tmontaigu:
This gives a bit more feedback when something went wrong.
Fixes #276


dgirardeau:
Nice!

PR #1031, opened 2020-02-06
Translate along and rotate around any arbitrary vector

### LINKED ISSUE #895: Translate/rotate along an arbitrary vector
Adding an option to the graphical translate/rotate tool for translating along an arbitrary vector instead of only along cardinal axis (x, y, z). I frequently find it useful to translate along the normal vector of a plane. I've tested modifying the dialog source and didn't have any trouble adding the functionality by overwriting one of the other axis, but I have no experience with QT to make the dialog modifications.

Maybe an advanced mode to the dialog with more options? An xyz triplet showing the vector along with a button to pick a vector from the db? 

Currently I copy a vector into XL and have a scalar to translate along it and a text output of a 4x4 matrix to apply to the cloud I want to move.

smortvedt:
This PR fixes #895 by adding the following functions:
- Rotation using an arbitrary vector as both the axis and center of rotation.
- Rotation using an arbitrary vector as the axis but using the object's center as the center of rotation.
- Rotation as above but using an arbitrary plane's normal as the vector.
- Translation along an arbitrary plane in the plane's x, y, and z coordinates.
- Translation along an arbitrary vector using the vector as the z direction.

The Rotation/Translation UI was modified to have dropdown menus to select the plane/line references if they exist in the db root. Only single segment polylines are allowed to be selected. There is a button to toggle between the advanced mode and the regular mode as well. Any comments or suggestions would be appreciated! Thanks for your time to review.

### DIFF
--- a/qCC/ccGraphicalTransformationTool.cpp
+++ b/qCC/ccGraphicalTransformationTool.cpp
@@ -531,8 +531,12 @@ void ccGraphicalTransformationTool::advRotateRefUpdate(int index)
 					mainWindow->db()->unselectEntity(m_advRotateRef);
 				}
 				m_advRotateRef = m_planesAndLineSegments[i];
-				m_advRotateRefIsChild = entityInTransformList(m_advRotateRef);
+				if (m_advTranslateRef != m_advRotateRef) // already selected
+				{
+					mainWindow->db()->selectEntity(m_advRotateRef, true);
+				}
 				refAxisRadio->setEnabled(true);
+				m_advRotateRefIsChild = entityInTransformList(m_advRotateRef);
 				if (m_advRotateRefIsChild)
 				{
 					refAxisRadio->setChecked(true);
@@ -541,10 +545,6 @@ void ccGraphicalTransformationTool::advRotateRefUpdate(int index)
 				else
 				{
 					objCenterRadio->setEnabled(true);
-				}
-				if (m_advTranslateRef != m_advRotateRef) // already selected
-				{
-					mainWindow->db()->selectEntity(m_advRotateRef, true);
 				}
 			}
 			if (!setAdvRotationAxis(m_advRotateRef))
@@ -584,6 +584,21 @@ void ccGraphicalTransformationTool::clear()
 	m_rotation.toIdentity();
 	m_translation = CCVector3d(0,0,0);
 	m_rotationCenter = CCVector3d(0,0,0);
+}
+
+void ccGraphicalTransformationTool::clearAdvModeEntities()
+{
+	MainWindow* mainWindow = MainWindow::TheInstance();
+	if (mainWindow && m_advTranslateRef != nullptr && m_associatedWin)
+	{
+		mainWindow->db()->unselectEntity(m_advTranslateRef);
+		m_advTranslateRef = nullptr;
+	}
+	if (mainWindow && m_advRotateRef != nullptr && m_associatedWin)
+	{
+		mainWindow->db()->unselectEntity(m_advRotateRef);
+		m_advRotateRef = nullptr;
+	}
 }
 
 bool ccGraphicalTransformationTool::addEntity(ccHObject* entity)
@@ -834,7 +849,15 @@ void ccGraphicalTransformationTool::updateAllGLTransformations()
 	if (m_advTranslateRefIsChild && m_advTranslateRef != nullptr) 
 	{
 		//update m_advTranslationTransform if the ref object is translated/rotated by virtue of being in m_toTransform
-		m_advTranslationTransform = m_position * ccGLMatrixd(m_advTranslateRef->getGLTransformationHistory().data());
+		if (m_advTranslateRef->isA(CC_TYPES::PLANE))
+		{
+			m_advTranslationTransform = m_position * ccGLMatrixd(m_advTranslateRef->getGLTransformationHistory().data());
+		}
+		else if (m_advTranslateRef->isA(CC_TYPES::POLY_LINE))
+		{
+			ccPolyline* line = static_cast<ccPolyline*>(m_advTranslateRef);
+			m_advTranslationTransform = m_position * arbitraryVectorTranslation(*line->getPoint(1) - *line->getPoint(0));
+		}
 	}
 	
 	MainWindow::RefreshAllGLWindow(false);
@@ -905,6 +928,8 @@ void ccGraphicalTransformationTool::apply()
 		}
 	}
 
+	clearAdvModeEntities();
+
 	stop(true);
 
 	clear();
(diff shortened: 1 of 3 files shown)


qCC/ccGraphicalTransformationTool.cpp
```diff
@@ -225,12 +225,91 @@ void ccGraphicalTransformationTool::stop(bool state)
...
 void ccGraphicalTransformationTool::glTranslate(const CCVector3d& realT)
 {
 	CCVector3d t(	realT.x * (TxCheckBox->isChecked() ? 1 : 0),
 					realT.y * (TyCheckBox->isChecked() ? 1 : 0),
 					realT.z * (TzCheckBox->isChecked() ? 1 : 0));
 
+	bool arbitraryVectorTranslation = true;
+	if (arbitraryVectorTranslation) 
+	{
+		CCVector3d vec(-.20, -.20, -.20);
```
dgirardeau: Is there a rationale behind this -0.2 quantity? Or is it really arbitrary?

qCC/ccGraphicalTransformationTool.cpp
```diff
@@ -210,6 +226,68 @@ bool ccGraphicalTransformationTool::setAdvancedTranslationTransform(ccHObject* t
...
+			ccLog::Print(QString("%1,%2,%3").arg(m_advRotationRefObjCenter[0]).arg(m_advRotationRefObjCenter[1]).arg(m_advRotationRefObjCenter[2]));
+			CCVector3d newCenter = (m_advRotationRefObjCenter - m_rotation.inverse() * m_position.getTranslationAsVec3D());
+			setRotationCenter(newCenter);
+		}
+		else
+		{
+			CCVector3d newCenter = CCVector3d::fromArray(m_toTransform.getBB_recursive().getCenter().u);
+			setRotationCenter(newCenter);
+		}
+		rotComboBox->setCurrentIndex(3); //Z
```
dgirardeau: Maybe use an enum here to be clearer? (and at the similar places as well)

qCC/ccGraphicalTransformationTool.cpp
```diff
@@ -218,7 +296,7 @@ void ccGraphicalTransformationTool::advTranslateRefChanged(int index)
 	}
 	int id = translateComboBox->itemData(index).toInt();
 	int selectedObjectIndex = -1;
-	for (int i; i < m_planesAndLineSegments.size(); i++)
+	for (int i = 0; i < m_planesAndLineSegments.size(); i++)
```
dgirardeau: I believe it should be size_t instead of int (if m_planesAndLineSegments is a vector, or a ccHObject::Container = vector)

dgirardeau: Same comment for other places

qCC/ccGraphicalTransformationTool.cpp
smortvedt: I believe you're looking at outdated code, this was test code from my first commit. Before I updated the UI to import a vector I manually changed the test vector.

qCC/ccGraphicalTransformationTool.cpp
smortvedt: Sure, I can add an enum throughout

qCC/ccGraphicalTransformationTool.cpp
smortvedt: Sure, ill change that

dgirardeau:
Ok, I tested on my side.

First thing: I believe you should increase the 'overlay UI' width, otherwise it's hard to read the options and the entities names:
image
(weel, that's probably because I have a 4K screen now :( )

dgirardeau:
Second:
Depending on the screen rotation, the translation along the plane was a bit weird (inversion of up/down compared to left/right --> I mean the translation along left/right was following the mouse motion, but the translation along up/down direction was inverted).

And if you change the virtual camera position (press the space bar of click on the "pause" button, move the camera, and then resume), the translation doesn't seem to stick to the plane directions anymore.

smortvedt:
Sorry about the translate weirdness, I think you were seeing the mouse moves still in reference to the origin frame and not the plane frame. I added a transform to the mouse moves so they should match up with the plane reference frame if using that option. 

Here's what the UI looks like on my screen at width 200:
image

I bumped it up to 250, it looks like maybe that would do it for yours as well, but I am kinda just guessing, is there a better way to size the width? Percentage maybe? Here's what my screen looks like a 250 

image

image

image

dgirardeau:
I'll try to update the dialog size so that it works better on my side (there seems to be something else happening, as the radio buttons are still displayed in a weird way even though the dialog is much larger):
image

dgirardeau:
And the 'translate along' option doesn't seem to work anymore on my side!

WargodHernandez:
@smortvedt 
I believe you are missing a QHBoxLayout around the radio buttons, I haven't checked it but that is a common side effect of not having that

smortvedt:
Just updated the layout. The groupbox widget was causing the layout spacing to be off, so I took that out and added a slot to effectively group the two radios.

As for the translation not working on your side, is the object not moving at all now? Or is it translating but still along the origin frame? I can add print statements to help track down the issue, on my side its working though :/

smortvedt:
@WargodHernandez just found a bug, when the reference object is a child of the object being rotated, the reference object is rotated as well and the resulting translations afterwards are not as expected. I'll be working on this bug, maybe this is what you are seeing as well!

dgirardeau:
On my side it is more that I can freely translate the object 'as usual', and the constraint of the plane doesn't seem to be taken into account (the planes were not children of the cloud)

dgirardeau:
I can try to debug it on my side but I'll need some time.

smortvedt:
Just committed the bug fix, I tested the all combinations of use that I could think of (rotation plane as child, rotation line as non-child, translation line as child, etc. etc.) as well as the different operations in series and it all passed my checks. 
The tests I ran boil down to rotating and translating along planes/lines when the are:
- a child of the selected object(s) (moving with the object)
- selected with the object(s) to translate/rotate (moving with the object)
- separate from the object(s) to translate/rotate (not moving with the object)

Chris is going to pull the code and test it later today. If its still not working for you I can add print statements in spots that may highlight what's the issue is.

dgirardeau:
Maybe I'm missing something.

I use this test dataset: 
[maquette_with_planes.zip](https://github.com/CloudCompare/CloudCompare/files/4201021/maquette_with_planes.zip)

I select the cloud, start the 'Interactive Transformation' tool, and then select the 'Advanced' mode and select one of the plane in the 'Translate along' field:
image

I was expecting to be constrained along this plane when translating the cloud. But it's as if nothing changed and as if the tool is just behaving 'normally'.

WargodHernandez:
> I was expecting to be constrained along this plane when translating the cloud. But it's as if nothing changed and as if the tool is just behaving 'normally'.

it appears you have all the translation active 
image
That would make it similar to the default, freedom on all axis.
if you restrict your selection to the x &/or y axis then the cloud will slide along the plane
if you select z you will slide along its normal.

or did you try that and I missed something?

dgirardeau:
Ah! Sorry, I didn't understood this option was relying on the above Tx, Ty and Tz checkboxes. (maybe because the first time I tested the tool it was already doing something even in these checkboxes were all checked?).

Anyway, now I get it,and it works ;)

dgirardeau:
Thanks for the contribution again!

PR #1033, opened 2020-02-07
WIP: Select correct LAS version_minor depending on the point format used

### LINKED ISSUE #1012: Bug with 2.11 alpha saving latest .las files
**Describe the bug**
In CloudCompare 2.11 alpha, when saving a cloud as a .las format 1.3 or 1.4 it creates a faulty file. file has internally 1.2 version but uses point format 7. File cannot be used anywhere. Not even in CloudCompare itself.

**How to reproduce**

1. Open a file (any point cloud will do)
2. Start saving it as .las/.laz file
3. Observe it asking for either 1.0 - 1.2 version or 1.3 - 1.4 version
4. Select the later, 1.3 - 1.4 format and save it
5. File created
6. File unusable everywhere and error about point format 7 pops up.

**Expected behaviour**
Should be able to save non-corrupted .las files as latest (1.4 las version) format. 

**Additional context**
Las-files 1.0 - 1.2 are saved as 1.2 in the las-file with point format 2. Las-files 1.3 - 1.4 are saved as format 1.2 with point format 7 (hex offset position 0x68 is 7 in the offending file signaling the point format)

Las format 1.2 is incompatible with point format 7 as per specifications and reader.prepare will fail with this corrupted file.

change las-file header to 1.4 format when saving 1.4 las-files if using point format 7

**Your environment**
 - CC Version (or git hash/tag): 2.11 alpha
 - OS & Version: Windows 10
 - Graphics card: Not relevant
 - Qt Version (if compiling): Not relevant

tmontaigu:
It should fix  #1012 
Also corrects the save dialog ui path which was not changed when the standard LASFilter was moved to a IO Plugin

I was not able to test it yet as I have a problem linking the qLASFWIO plugin to LASlib

### DIFF
--- a/plugins/core/IO/qLASFWFIO/CMakeLists.txt
+++ b/plugins/core/IO/qLASFWFIO/CMakeLists.txt
@@ -16,7 +16,7 @@ if (WIN32)
 		set( CC_PLUGIN_CUSTOM_SOURCE_LIST Filter/LASFWFFilter.cpp )
 			
 		#we need the saveLASFileDlg.ui file
-		set( CC_PLUGIN_CUSTOM_UI_LIST ${QCC_IO_LIB_SOURCE_DIR}/ui_templates/saveLASFileDlg.ui )
+		set( CC_PLUGIN_CUSTOM_UI_LIST ../qPDALIO/ui/saveLASFileDlg.ui )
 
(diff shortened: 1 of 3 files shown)


dgirardeau:
Thanks!

PR #1035, opened 2020-02-13
Fix polyline special case

smortvedt:
Sorry about this! I just was going through the code and saw this special case that didn't look right and sure enough it wasn't. This bug only affects certain polylines depending on the direction they were made in, so before most of them translated correctly... but I just tested all line cases (2 directions per octant) and they all translated as expected now.


WargodHernandez:
I just noticed that the change log didn't get updated either. Sorry missed that

Issue #403, opened 2016-06-21
How to access command line mode and call plugins?


OH8HO:
This is a two part question:
1. I've been trying to figure out how to use the command line mode with little luck thus far. I'm using windows and mac. Do we access the command line mode through CC or do we simply called it from terminal/CMD? 

2. Assuming question #1 is a breeze and I can now ask CC to do something in command line mode, is it possible to use plugins like M3C2 (M3C2 is windows only)?

Thanks!

cloudcompare:
Indeed you have to call the 'command line' mode from a (system) console (e.g. 'cmd' on Windows, not sure how to start this on a Mac).

Then to call the M3C2 plugin, I made changes recently so that the plugins are now loaded even in command line mode. But then one needs to modify the code to be able to call this plugin, input parameters, etc.

OH8HO:
Thanks I've got CC running from CMD. Which code would I need to modify to call the plugin? QM3C2_PUGIN_DLL.dll or something else?

cloudcompare:
To work properly, we should create a generic mechanism to issue commands to a plugin.

A specific syntax like '-plugin:{plugin_name}'. Then all the commands after this one would be passed to the plugin.

On the plugin side (ccStdPluginInterface), we can add a generic method (e.g. 'parseCommand') that would take the remaining commands as argument. This way the plugin will be able to process its own commands and all the specific code will be on the plugin side.

Eventually you would be able to implement the specific command processing for qM3C2 easily.

Do you want me to setup this mechanism? Will you be able then to implement the specific command line process for qM3C2?

OH8HO:
Just to clarify. The idea being that we could write:

CloudCompare -O file_1.las -O file_2.las -plugin:qM3C2 arg1:arg2:arg3 

That would be incredibly helpful. M3C2 has quiet a few arguments, would it be too difficult to have access to most of them?

cloudcompare:
Yes, or to keep the syntax the same as for the other tools, it may be something like:
`CloudCompare -O file_1.las -O file_2.las -plugin:qM3C2 -option1 arg1.1 arg1.2 -option2 arg2 etc.`

We would have to figure out how to specify to M3C2 which cloud should be used. And indeed there's a lot of options. This is why I wondered if you were ready to implement the command line process yourself for M3C2 ;).

dh81:
Would be really nice to have that by the command line. I unfortunately can not do it by myself.

OH8HO:
[...]

PR #1036, opened 2020-02-16
Ransac plugin modification

WargodHernandez:
I'm working through modifying the Ransac plugin to add command line & ChaiScript support. This first effort is just splitting the  doAction() into doAction() handles the dialog and adding the resulting container to the app DB.
and 
`ccHObject* qRansacSD::executeRANSAC(ccPointCloud* ccPC, RansacParams& params, bool silent)`

all of the 
m_app->dispToConsole commands have been replaced with ccLog::Warning or ccLog::Error to support the command line


dgirardeau:
Are the messages still routed to the graphical console when calling ccLog from a plugin.

WargodHernandez:
@dgirardeau 
> Are the messages still routed to the graphical console when calling ccLog from a plugin.

Yes, it does since the graphical console is the attached instance to the static ccLog
Honestly I think this change should be made everywhere to make it universal. Maybe add a generic ccLog::Message(QString, MSGLevel) function to be equivalent to the dispToConsole function. The only place where it is different at all is when using the command line version and sending an error the function has a return of false. 
But otherwise all cmd.Print, cmd.Warning, and dispToConsole are all just calling ccLog static functions in essentially the same way

dgirardeau:
I remember that at one point in time, it was not working because the static instance was not the same in the main application, and in the (plugin) DLL part... This is why the dispToConsole method was added. Have you tried on your side?

WargodHernandez:
Yes, I got the result on the console when I used ccLog::Error at least, I'll have to check for the warning and print, but I would expect them to work since I got the message on error

dgirardeau:
Interesting! Then in this case we can definitely remove the dispToConsole mechanism.

WargodHernandez:
Oh actually I guess I have checked the print, warning, and error. On the chaiscript plugin, I have switched over to just using ccLog and the scripts were tested with all three types

asmaloney:
Accessing application globals directly from a plugin is a bad idea in my opinion.

1) It increases the coupling of plugins & application which is the opposite of what we should be aiming for: a small, clean, well-defined contract between plugins and application. It would be the API + whatever globals are used in the plugins + whatever classes are used that are not explicitly in the API. Better not change those globals ever again!

2) It may introduce lifetime/linking issues. We deal with multiple compilers, linkers, and OSes. Getting them all to play well with globals, ODR, dynamic libraries, and lifetime issues is not trivial. So while it might work now for you on your compiler, even something as simple as changing the order of variable declarations might break it in the future.

WargodHernandez:
I can agree with that logic, I think it points out a missing interface for plugins, just a generic ccLog interface not tied to m_app or cmd , so we can cut down on duplication and double checks of which interface is valid in the deeper parts of the plugins which are shared between command line and CC app

WargodHernandez:
Another area that could benefit from a common interface would be progress indication, the interface would spawn a QProgressDialog for CC and command line, and just print start message and indicator dots on the console for silent command line.

WargodHernandez:
@dgirardeau 
I have the Command Line mode now functioning here is an example of it running with default settings with the maquette_with_planes.bin model you posted recently
image
with the following command
```
@set PATH=.
CloudCompare.exe -O "C:\Users\af836c\test\maquette_with_planes.bin" -RANSAC
```

I need to look into how to support the save all at once feature, and also need to work on the automatic naming convention a bit. I believe right now without the added date/time the planes would have all saved over each other.

Here is a list of the subcommands that are evaluated

1. "EPSILON_ABSOLUTE"
2. "EPSILON_PERCENTAGE_OF_SCALE"
3. "BITMAP_EPSILON_PERCENTAGE_OF_SCALE"
4. "BITMAP_EPSILON_ABSOLUTE"
5. "SUPPORT_POINTS"
6. "MAX_NORMAL_DEV"
7. "PROBABILITY"
8. "ENABLE_PRIMITIVE"
a. "PLANE" 
b. "SPHERE" 
c. "CYLINDER" 
d. "CONE" 
e. "TORUS"

	
The defaults for these when unspecified are as follows
```C++
        epsilon = 0.005f * scale;
	bitmapEpsilon = 0.01f * scale;
	SUPPORT_POINTS = 500;
	maxNormalDev_deg = 25;
	primEnabled[RPT_PLANE] = true;
	primEnabled[RPT_SPHERE] = false;
	primEnabled[RPT_CYLINDER] = false;
	primEnabled[RPT_CONE] = false;
	primEnabled[RPT_TORUS] = false;
```

this will work for as many clouds as are loaded when the command is executed, and when you use more than one cloud it is probably a good idea to use "EPSILON_PERCENTAGE_OF_SCALE" and "EPSILON_PERCENTAGE_OF_SCALE", scale is calculated for each cloud individually prior to evaluation.

WargodHernandez:
@dgirardeau @asmaloney 
One other item to note is this does not store the new sub clouds with each primitive, I'm not sure how to do that cleanly since the command line seperates the clouds and meshes. I saw the CLGroupDesc but there is no corresponding groups access in ccCommandLineInterface like there is for clouds and meshes.
I wasn't sure if there was a proper way to attach the cloud sub to the mesh and have the pair save together.

WargodHernandez:
Alright I got everything sorted out, I believe.
The plugin still access ccLog currently but I believe creating another interface should be a stand-alone pull request.
I've tested with and without -silent and multiple input clouds, and specifying all the various output and directory options.

I'm going to switch it over to ready for review

plugins/core/Standard/qRANSAC_SD/qRANSAC_SD_Commands.h
```diff
@@ -0,0 +1,514 @@
...
+			return cmd.error(QObject::tr("No point cloud to attempt RANSAC on (be sure to open one with \"-O [cloud filename]\" before \"-%2\")").arg(COMMAND_RANSAC));
+		}
+		qRansacSD::RansacParams params;
+		QStringList paramNames = QStringList() << EPSILON_ABSOLUTE << EPSILON_PERCENTAGE_OF_SCALE <<
+			BITMAP_EPSILON_PERCENTAGE_OF_SCALE << BITMAP_EPSILON_ABSOLUTE <<
+			SUPPORT_POINTS << MAX_NORMAL_DEV << PROBABILITY << ENABLE_PRIMITIVE <<
+			OUT_CLOUD_DIR << OUT_MESH_DIR << OUT_GROUP_DIR << OUT_PAIR_DIR << OUTPUT_INDIVIDUAL_PRIMITIVES <<
+			OUTPUT_INDIVIDUAL_SUBCLOUDS << OUTPUT_GROUPED << OUTPUT_INDIVIDUAL_PAIRED_CLOUD_PRIMITIVE;
+		QStringList primitiveNames = QStringList() << PRIM_PLANE << PRIM_SPHERE << PRIM_CYLINDER << PRIM_CONE << PRIM_TORUS;
+		QString outputCloudsDir = "";
```
dgirardeau: Not sure about this initialization to "" (QStrings could be left to their default 'null' value?)

plugins/core/Standard/qRANSAC_SD/qRANSAC_SD_Commands.h
```diff
@@ -0,0 +1,514 @@
...
+			}
+			if (!paramNames.contains(param))
+			{
+				cmd.arguments().push_front(param);
+			}
+		}
+		unsigned char primCount = 0;
+		for (unsigned char k = 0; k < 5; ++k)
+		{
+			primCount += (unsigned)params.primEnabled[k];
```
dgirardeau: static_cast?

plugins/core/Standard/qRANSAC_SD/qRANSAC_SD_Commands.h
```diff
@@ -0,0 +1,514 @@
...
+			cmd.print(QObject::tr("\tDefault Shape Search == %1").arg(PRIM_PLANE));
+			params.primEnabled[qRansacSD::RPT_PLANE] = true;
+		}
+		if (!outputIndividualClouds && !outputIndividualPrimitives && !outputGrouped && !outputIndividualPairs)
+		{
+			cmd.print(QObject::tr("\tDefault output == %1").arg(OUTPUT_GROUPED));
+			outputGrouped = true;
+		}
+
+		for (CLCloudDesc clCloud : cmd.clouds())
```
dgirardeau: const CLCloudDesc& ?

plugins/core/Standard/qRANSAC_SD/qRANSAC_SD_Commands.h
WargodHernandez: yeah, the folly of copy/paste from the existing code :)
I'll change both locations

plugins/core/Standard/qRANSAC_SD/qRANSAC_SD_Commands.h
WargodHernandez: The cloud inside clCloud gets possibly modified when passed to executeRANSAC (normal can be added), also there is no overload const overload for ccCommandLineParser::exportEntity.

Plus I tried with CLCloudDesc& and experienced crashing...

plugins/core/Standard/qRANSAC_SD/qRANSAC_SD_Commands.h
WargodHernandez: yeah it worked without the initialization to "", I guess I'm just used to always initializing to something

WargodHernandez:
I think this resolves at least one open issue #403 once merged

PR #1040, opened 2020-02-21
Moving default color for points (and vertices) to RGBA

dgirardeau:
To fix a big bug with ATI cards (they don't accept VBOs that are not 32bits aligned :| ), I propose to manage point clouds and meshes with RGBA colors. It might also allow later some nice rendering with transparent points!


WargodHernandez:
So currently if the alpha value is set to anything other than ccColor::MAX does the point render differently? Or will we need to instrument something the rendering code?

dgirardeau:
Yes. But for now it might be a little ugly since you have to handle how the order in which points are displayed (from back to front) if you want a decent rendering / transparency effect... The LOD mechanism might come in handy for that.

WargodHernandez:
> Yes. But for now it might be a little ugly since you have to handle how the order in which points are displayed (from back to front) if you want a decent rendering / transparency effect... The LOD mechanism might come in handy for that.

Got it, I am interested in trying out the transparency with the projects I work with which tend to create hundreds or thousands of smaller clouds from a few larger source clouds. The clouds are essentially surfaces and I know which clouds are nearer to the camera, so I am thinking I can work out cloud draw order pretty easily (instead of per point draw order)

libs/qCC_db/ccColorTypes.h
```diff
@@ -215,21 +215,30 @@ namespace ccColor
...
 	{
 		return Rgb( static_cast<ColorCompType>(color.r * MAX),
 					static_cast<ColorCompType>(color.g * MAX),
 					static_cast<ColorCompType>(color.b * MAX) );
 	}
 	
-	//! Conversion from Rgbaf
-	inline Rgb FromRgbf(const Rgbaf& color)
+	//! Conversion from Rgbaf to Rgb
+	inline Rgb FromRgbafToRgb(const Rgbaf& color)
```
WargodHernandez: I was going to suggest a change to this function name when I saw it in the ChaiScript side of things since it hadn't followed the surrounding naming convention, I had renamed it to FromRgbaf

WargodHernandez:
This branch works on my two main Windows systems, both of them are nvidia cards though.

Has anyone tried this out on ATI/Linux/Mac configs yet?

dgirardeau:
Yes I have an ATI (Radeon Pro) that was lent to me by a company. And this is how I discovered the VBO 'bug' of ATI cards, and this was the initial motivation for this updated.

PR #1044, opened 2020-02-24
Fix large unique IDs bug

dgirardeau:
Avoid increasing unique IDs with the highest loaded ID to avoid picking issues (due to too high IDs)


dgirardeau:
Not my branch?!

PR #1045, opened 2020-02-24
Use Ubunutu Bionic for the CI & add qpdal plugin to compilation

tmontaigu:
Changes the travis ci to build the qPDALIO plugin, this makes travis use Ubuntu Bionic (18.04LTS) as  Xenial (16.04LTS) does not have pdal in the repos.

I don't know if upgrading from Xenial to Bionic will have some unwanted effects and if there any objections to doing it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 sudo: required
-dist: xenial
+dist: bionic
 cache: ccache
 language: cpp
 compiler:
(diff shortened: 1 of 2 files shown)


WargodHernandez:
if we still want to support Xenial can we add Bionic as an additional build type? 

I don't need Xenial support BTW, just curious.

tmontaigu:
If we would still need Xenial support we could see if travis allows us to build on different ubuntu versions I know Github actions allow that so I guess travis also.

PR #1047, opened 2020-02-24
Fix picking issues due to too high unique IDs

dgirardeau:
Avoid increasing unique IDs with the highest loaded ID to avoid picking issues (due to too high IDs)


WargodHernandez:
I wonder if it would make sense to have a reserved range of UID for the User Interface to use (like all dialog created/temporary interactors would use 1 range and everything else would be a 2nd range)?

Also maybe we should add a ccLog::Warning when the UID is near max.

dgirardeau:
Interesting idea (even though I doubt you created almost 4 billion entities?! I guess there's more probably a glitch somewhere). And I need to find what is the exact limit for OpenGL (since apparently 32bit integers are supported).

WargodHernandez:
> Interesting idea (even though I doubt you created almost 4 billion entities?! 

I like to create things, apparently billions of things.  :-)  

I think its mostly because I can create millions of things when I run my process, and if I save any of those any of those items into a file which then I mix in with other things than I will keep advancing that UID counter without actually using all those UIDs. and when I was developing the automation I ran the process sometimes hundreds of times looking for memory leaks, or tweaking settings, I would delete most/all of the end items before each run, but that wouldn't turn the counter back.

dgirardeau:
Ok, then at least with the new system, all IDs are reset at each run... Not sure how to properly handle fixed/reserved IDs yet.

WargodHernandez:
Very rough idea could be changing

```C++
//! Resets the unique ID
void reset() { m_lastUniqueID = 0; }
```

to 

```C++
constexpr unsigned FIRST_UNRESERVED = 100;
//! Resets the unique ID
void reset() { m_lastUniqueID = FIRST_UNRESERVED; }
```

and then with dialog/UI based interactors these really need to work 100% of the time for the app to feel polished. 
`setUniqueID(someConstID_LT_100);`

qCC/ccClippingBoxTool.cpp
```diff
@@ -22,6 +22,7 @@
 #include "ccClippingBoxRepeatDlg.h"
 #include "ccCropTool.h"
 #include "ccGLWindow.h"
+#include "ccReservedDs.h"
```
WargodHernandez: was ccReservedDs.h intentional or should it be ccReservedIDs.h ?

dgirardeau: Oops! Nice catch!

PR #1054, opened 2020-03-06
Avoid possible buffer overflow in BinFilter.

JoseGoncalves:


dgirardeau:
Ah! That's very old code (hopefully not used anymore) but thanks for the fix!

PR #1060, opened 2020-03-18
Load/save GLTF mesh using fx-gltf

### LINKED ISSUE #851: glTF/glb format support (export/import)
This feature has been requested originally on the [CC forum](http://www.danielgm.net/cc/forum/viewtopic.php?t=2845). I currently use an [online converter](http://modelconverter.com) to convert from OBJ to glTF 2.0 but an integrated solution would eliminate this extra step. glTF has the ambition to become the new common publishing format for 3D content.

The specification is available on the [KhronosGroup GitHub page](https://github.com/KhronosGroup/glTF/blob/master/specification/2.0).

EpicWink:
Will fix #851


asmaloney:
Thank you for the WIP.

I don't know that we want to support a core plugin for this. That's why I made the [MeshIO](https://github.com/asmaloney/MeshIO) a 3rd party plugin to handle COLLADA, glTF, and IFC-SPF..

(We've also not yet moved to requiring C++14 & it seems that fx-gltf requires it?)

dgirardeau:
Oh! I myself forgot about your MeshIO plugin!

Andy, shouldn't we include it as a submodule in the 'plugins/3rdParty' folder to avoid such confusion?

EpicWink:
I'll close this in a day if we're happy with using MeshIO instead

asmaloney:
What we're talking about is adding MeshIO as a submodule in 3rd party.

Then if you want to put this in a separate repo, we can link a submodule in there as well.

They would not be built as part of the main releases, but people can build them themselves. At least this way there's some visibility. We probably need a page somewhere to explain it and list the ones in there.

This would mean a PR whenever a new version is available.

What do you think?

asmaloney:
I'm going to propose that you separate this into its own repo and we can add it as a 3rd party plugin as well.

glTF might work better/differently with the fx-gltf library then what I'm using in MeshIO.

We may need some changes to allow multiple plugins to support the same extensions (a way to set a default handler when opening with "All (*.*)" for example).

dgirardeau:
Interesting. We already have this issue on Windows with the LAS filters (Q_PDAL_IO and Q_LAS_FWF_IO). They both load LAS/LAZ files. The second one doesn't declare any default type, and users have to open the files via the Open Dialog and manually select the right filter.

(Sadly this is necessary as the two 3rd party libraries don't support the same features :| ).

asmaloney:
Along the same lines, one thing I did with MeshIO was to add the plugin name to the filter, so at least if there are multiple plugins working with the same extensions, you can pick the one you want.

Screen Shot 2020-03-19 at 10 45 53 AM

(Although, now that I look at it, maybe the format should come first, followed by the plugin name.)

As this list gets quite big, we may need to look at sorting and/or a different solution. A dropdown with 25+ items is... not very user friendly :-)

WargodHernandez:
> As this list gets quite big, we may need to look at sorting and/or a different solution. A dropdown with 25+ items is... not very user friendly :-)

How about a file settings dialog where we could let users to adjust the priority for each IO filter, or a always use this filter for these file types (with save/load handled individually), or always prompt the user except when -silent in command line.

We could keep track of whether or not a user has made the decision before and ask them when multiple filters support a file type, and allow them to "remember this decision" for follow on file load/save events.

This would require adding a section to the permanent settings, and adjusting the filter priorities prior to any load/save operations, but I don't think it would be an excessive project.

EpicWink:
I think the real user unfriendliness is having multiple options for the same file-type in the first place. In theory, each filter should produce equivalent output for the same file type anyway. I think it would make more sense to choose which filter is used for a given file-type a build time

dgirardeau:
The issue with LAS at least, is that the two 3rd party libraries have very different support levels:
- PDAL doesn't fully manage the LAS 1.3+ and 1.4 formats, especially with full waveform data (FWF). But it has a good support, it is cross-platform, etc.
- LASlib is a one-man project with an old-fashioned coding style and it only compiles on Visual Studio... but it supports all LAS formats and FWF 😅

tmontaigu:
>     * LASlib is a one-man project with an old-fashioned coding style and it only compiles on Visual Studio... but it supports all LAS formats and FWF

LASlib also compiles on linuxes and should also compile on macos

dgirardeau:
Ah! But do we really want to rely only on LASlib? 👿

EpicWink:
I'm closing this as there's no interest. We can re-open if that's not the case. Hopefully this is searchable if users request this feature in the future

PR #1061, opened 2020-03-19
MeshIO plugin

EpicWink:
Add [MeshIO](https://github.com/asmaloney/MeshIO) as a plugin for loading GLTF/GLB/COLLADA mesh files

This replaces #1060 

Note: the cmake variable is `PLUGIN-3rdParty_MESH_IO`, when perhaps it should become `PLUGIN_MESH_IO`

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "plugins/core/IO/qE57IO/extern/libE57Format"]
 path = plugins/core/IO/qE57IO/extern/libE57Format
 url = https://github.com/asmaloney/libE57Format
+[submodule "plugins/3rdParty/MeshIO"]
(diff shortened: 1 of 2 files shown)


EpicWink:
Ahh, I think there are some minor issues still @dgirardeau:
* The commit in question is broken for x64 Linux using GCC 9.2: see asmaloney/MeshIO#3
* It requires `assimp`, which takes a long time to clone
* The plugin is not included in Travis CI
* The Cmake variable has a funny name

EpicWink:
I'm closing this as there's no interest. We can re-open if that's not the case. Hopefully this is searchable if users request this feature in the future.

In addition, this plugin should probably be easier to find (table in CloudCompare's README? I wiki page?)

PR #1064, opened 2020-03-20
Format: Convert spaces to tabs

asmaloney:
Fixes 3 files in CCLib

### DIFF
--- a/CC/src/DgmOctree.cpp
+++ b/CC/src/DgmOctree.cpp
@@ -1,19 +1,19 @@
 //##########################################################################
-//#                                                                        #
-//#                               CCLIB                                    #
-//#                                                                        #
-//#  This program is free software; you can redistribute it and/or modify  #
-//#  it under the terms of the GNU Library General Public License as       #
-//#  published by the Free Software Foundation; version 2 or later of the  #
-//#  License.                                                              #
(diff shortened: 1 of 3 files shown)


CC/src/DgmOctree.cpp
```diff
@@ -1,19 +1,19 @@
...
-//#  License.                                                              #
-//#                                                                        #
-//#  This program is distributed in the hope that it will be useful,       #
-//#  but WITHOUT ANY WARRANTY; without even the implied warranty of        #
-//#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the          #
-//#  GNU General Public License for more details.                          #
-//#                                                                        #
-//#          COPYRIGHT: EDF R&D / TELECOM ParisTech (ENST-TSI)             #
-//#                                                                        #
+//#																		   #
```
dgirardeau: Maybe not the headers?

asmaloney: They actually line up properly in editors - just not on GitHub.

(Also - it would be cleaner to get rid of the trailing '#' anyways :-) )

asmaloney: I wish there were a GitHub setting for tab display size...

PR #1072, opened 2020-03-23
Add CI build on osx & windows using github actions

tmontaigu:
It uses conda as the package manager to get dependencies

QPCL does not work on mac OS because clang does not have openmp ?!

QRANSAC_SD does not work on mac OS  because aligned_alloc which should be available since C11 standard in the stdlib.h header,  is not available on clang.
(Apparently C++17 added aligned_alloc in the standard, and clang supports that but we do not build with C++17 enabled)

We should probably re add the old custom aligned_alloc that was implemented here for OSX targets
https://github.com/CloudCompare/CloudCompare/commit/5d4fb8cf6ef2bef845913a1378a2eca5a5a8ed26

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,113 @@
+name: Build
+
+on: push
+
+jobs:
+  build:
+    name: ${{ matrix.config.name }} SCALAR_DOUBLE=${{ matrix.scalar_double }}
+    runs-on: ${{ matrix.config.os }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        scalar_double: ["ON", "OFF"]
+        config:
+          - {
+              name: "Windows Latest MSVC",
+              os: windows-latest,
+              generator: "NMake Makefiles",
+              conda_library_dir: "Library",
+              compile_qpcl: "ON",
+              compile_qransac: "ON",
+              zlib_name: z.lib,
+              xercesclib_name: xerces-c_3.lib,
+           }
+          - {
+              name: "macOS Latest Clang",
+              os: macos-latest,
+              generator: "Ninja",
+              conda_library_dir: ".",
+              compile_qpcl: "OFF",
+              compile_qransac: "OFF",
+              zlib_name: libz.a,
+              xercesclib_name: libxerces-c-3.2.dylib
+            }
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+        with:
+          submodules: true
+
+      - uses: goanpeca/setup-miniconda@v1
+        with:
+          miniconda-version: 'latest'
+          activate-environment: ccenv
+          auto-update-conda: true
+
+      - name: Install dependencies
+        shell: pwsh
+        run: conda install -y -c conda-forge `
+          qt=5.12.* `
+          pdal=2.0.* `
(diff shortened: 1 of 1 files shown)


asmaloney:
> clang does not have openmp

Actually it does - we need to modify the cmake to link with it properly, but that requires using the "modern cmake" approach with targets. We're not quite set up for that properly .

[This](https://stackoverflow.com/a/56207996) is the correct way.

> QRANSAC_SD does not work on mac OS 

FYI - I've never included the RANSAC plugin in the "official" release because I've never had it build cleanly.

WargodHernandez:
@tmontaigu 
> We should probably re add the old custom aligned_alloc that was implemented here for OSX targets
5d4fb8c

well the original allocator looked like this
```
#include <memory>
#ifndef __APPLE__
#include <malloc.h>
#else
#include <stdlib.h>
#endif
#include <xmmintrin.h>
```
This should have provided _mm_malloc and _mm_free from either malloc.h or xmmintrin.h
but on platforms where that wasn't supported (think ARM) the xmmintrin.h wouldn't be available, and on systems without malloc.h (and even some with it) _mm_malloc would be undefined. 

all of this was so that when compiled with an Intel compiler there is a speedup with the intrinsics
From xmmintrin.h
```
#ifdef __ICL
extern void* __cdecl _mm_malloc(size_t _Siz, size_t _Al);
extern void __cdecl _mm_free(void *_P);
#endif  /* __ICL */ 
```
the __ICL define is for the Intel compiler. without ICL on windows the _mm_malloc comes from malloc.h like this
```
#define _mm_free(a)      _aligned_free(a)
#define _mm_malloc(a, b) _aligned_malloc(a, b)
```


When I changed the defines to the following
```
#if defined(__x86_64__) || defined(__i386__) || defined(_M_IX86) || defined(_M_X64)
#include <xmmintrin.h>
#endif

#ifdef _mm_malloc
#ifndef a_malloc
#define a_malloc(align,sz) _mm_malloc((align),(sz))
#endif // !a_malloc
#endif // !_mm_malloc
#ifdef _mm_free
#ifndef a_free
#define a_free(ptr)  _mm_free((ptr))
#endif // !a_free
#endif // !_mm_free

#ifndef a_free  
#define a_free(a)      free(a) 
#endif // !_mm_free
#ifndef a_malloc
#define a_malloc(a, b) aligned_alloc(a, b)
#endif // !_mm_malloc
```

The intent was to use aligned_alloc to be compatible with Windows, Linux and Mac and allow the speed up when compiled with the Intel compiler. The only reason to use these custom vector classes is the performance gained when compiled with an Intel compiler.

My thought is... how many people actually use the Intel compiler on CC? 

If we still want the benefit for Intel compilers we can add another define to use whatever aligned malloc is available to Mac when compiled with C++11/14

dgirardeau:
I haven't heard about anyone compiling CC with the Intel compiler... Considering the plugin is already performing quite well on the other compilers, I would personally favor the option where we remove all that (it improves the maintainability and readability 😉).

WargodHernandez:
Should we merge this in, the actual content of the PR looks like its working and isn't really related to the other talk (OpenMP & alloc) in this thread
image

dgirardeau:
Thanks Thomas ;)

PR #1074, opened 2020-03-24
[Feature] unroll action unroll mesh as well

rjanvier:


rjanvier:
@dgirardeau I though the feature already exists.

qCC/mainwindow.cpp
```diff
@@ -5589,14 +5589,21 @@ void MainWindow::doActionUnroll()
...
 
 		if (pc->getParent())
 		{
 			pc->getParent()->addChild(output);
 		}
+		
+		//if it's a mesh, unroll it by swapping vertices
+		if(m_selectedEntities[0]->isA(CC_TYPES::MESH)){
+			ccMesh* mesh = ccHObjectCaster::ToMesh(m_selectedEntities[0]);
+					mesh->setAssociatedCloud(output);
```
dgirardeau: I think if you want to do this, you would have first to delete the old vertices? But to do that properly, you have to remove it from the DB tree.

It would look like:
```
mesh->removeChild(pc);
mesh->setAssociatedCloud(output);
delete pc;
pc = nullptr;
```
(and make sure pc is not used afterwards)

And also you should actually call mesh->addChild(output); in this case (instead of adding it to pc's parent).

rjanvier: I wanted to keep both because actually we keep both so I thought it feels more natural this way (but it could be awkward if not documented, I agree). What is your preference?
agree for the `mesh->addChild` thing

dgirardeau: Well you would keep only the old vertices then? That's a little bit strange since you can't restore the vertices of the mesh with the old cloud or anything)

In this case wouldn't it be better to clone the mesh and replace its vertices?
```
		ccMesh* outputMesh = new ccMesh(mesh, output);
		outputMesh->addChild(output);
		addToDB(outputMesh, true, true, false, true);
```

dgirardeau: On my side I have:
```
	if (output)
	{
		pc->setEnabled(false);
		ccConsole::Warning("[Unroll] Original cloud has been automatically hidden");

		if (m_selectedEntities[0]->isA(CC_TYPES::MESH))
		{
			ccMesh* mesh = ccHObjectCaster::ToMesh(m_selectedEntities[0]);
			mesh->setAssociatedCloud(output);		ccMesh* outputMesh = new ccMesh(mesh, output);
			outputMesh->addChild(output);
			addToDB(outputMesh, true, true, false, true);
		}
		else
		{
			if (pc->getParent())
			{
				pc->getParent()->addChild(output);
			}
			addToDB(output, true, true, false, true);
		}

		updateUI();
	}
```

rjanvier: Before this change we kept both clouds (original and unrolled). I thought keeping the original vertices could be convenient (for documentation purposes) even if we couldn't swap them anymore... 
Maybe your proposal the full copy of the mesh is a compromise but in this case I think we should make it optional in because I afraid a copy is not always wanted / possible (if your mesh is too big for example).
let's call this for example  "unroll mesh", it could be a checkbox in the current dialog box
- unchecked -> the current behavior (before this PR)
- checked -> the mesh is copied and the copy is unrolled
Does it seems reasonable?

dgirardeau: Oh then in this case I would just add a checkbox "keep original entity"?

rjanvier: would it apply both for cloud and mesh? Or only for mesh?

dgirardeau:
Oh no for both then. So as to get a consistent behavior.

qCC/mainwindow.cpp
dgirardeau: Oh no for both then. So as to get a consistent behavior.

rjanvier: at last I kept your initial proposal because we will clone anyway (so I let the user do the clean up if he want to).
I use `cloneMesh()` method instead of copy constructor because In my specific case I want to keep material and textures and I feel it could be useful for the vast majority of the users.

PR #1084, opened 2020-03-28
{cmake} Move macOS-specific stuff out of the main cmake files

asmaloney:
The extra fix_libs script seems to no longer be required as CGAL is header-only


qCC/Mac/CMakeLists.txt
```diff
@@ -1,12 +1,21 @@
...
+	set_target_properties( ${PROJECT_NAME} PROPERTIES
+        INSTALL_RPATH "@executable_path/../Frameworks"
+        MACOSX_BUNDLE_INFO_PLIST ${CMAKE_CURRENT_SOURCE_DIR}/CloudCompare.plist )
+
+    set( MACOSX_BUNDLE_ICON_FILE cc_icon.icns ) 
+    set( MACOSX_BUNDLE_SHORT_VERSION_STRING "${PROJECT_VERSION}" )
+    set( MACOSX_BUNDLE_LONG_VERSION_STRING "${PROJECT_VERSION}" )
+    set( MACOSX_BUNDLE_BUNDLE_VERSION "${PROJECT_VERSION}" )
+
+	set( CLOUDCOMPARE_MAC_BASE_DIR ${CMAKE_INSTALL_PREFIX}/${CLOUDCOMPARE_DEST_FOLDER}/CloudCompare.app  CACHE INTERNAL "CloudCompare bundle dir")
```
rjanvier: indent?

ccViewer/Mac/CMakeLists.txt
```diff
@@ -1,12 +1,24 @@
...
 
-   # install icons
-   install( FILES ../images/icon/cc_viewer_icon.icns DESTINATION ${CCVIEWER_MAC_BASE_DIR}/Contents/Resources COMPONENT Runtime )
+	set( MACOSX_BUNDLE_ICON_FILE cc_viewer_icon.icns )
+	set( MACOSX_BUNDLE_SHORT_VERSION_STRING "${PROJECT_VERSION}" )
+	set( MACOSX_BUNDLE_LONG_VERSION_STRING "${PROJECT_VERSION}" )
+	set( MACOSX_BUNDLE_BUNDLE_VERSION "${PROJECT_VERSION}" )
 
-   # remove any dev-machine specific lib paths
-   install( CODE "execute_process( COMMAND \"${CMAKE_SOURCE_DIR}/scripts/mac/fix_libs.sh\"
```
rjanvier: CGAL is optionally header only starting from 4.9 and it's the default since 5.0. Maybe you have to check the version of CGAL and force/check if header only is enabled. What do you think?

asmaloney: This was actually a hack-workaround for homebrew that was used for building releases. (And the script had a hard-coded path & version numbers in it.)

I don't think the complexity of checking versions and keeping this is worth it.

PR #1088, opened 2020-03-29
Write correct value for MTL Transparency

rjanvier:
This is a follow up of #1059. The same "Tr" inversion occurs at save time.


dgirardeau:
Good catch :D

PR #1096, opened 2020-04-01
Github Actions Push and Pull

WargodHernandez:
This will cause two checks to run for those of us opening pull requests from this repo, but will cause the build to happen when the PR is from another fork.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,6 @@
 name: Build
 
-on: push
+on: [push, pull_request]
 
 jobs:
   build:


WargodHernandez:
Notice https://github.com/CloudCompare/CloudCompare/pull/1094 when I submitted a PR from my fork, only Travis executed for the check.

tmontaigu:
but now checks are done twice when there is a pull request

WargodHernandez:
> 
> 
> but now checks are done twice when there is a pull request

Yeah I know, I wish there was an OR selection. 
The solution listed [here](https://github.community/t5/GitHub-Actions/How-to-trigger-an-action-on-push-or-pull-request-but-not-both/td-p/35805) looks like it needs configuration on a per branch basis.

WargodHernandez:
@tmontaigu 
It looks like Travis currently runs checks twice currently for pull requests originating from this repo
image

asmaloney:
Looks like we need to revisit this :-)

The general consensus seems to be that running twice is by design - once for the current branch and once for the (potential) merge.

tmontaigu:
Ok, it makes sense i didn't knew travis was doing that

asmaloney:
Should we give this a try? Can't hurt.

PR #1099, opened 2020-04-01
Attempt to build on Push or Pull

WargodHernandez:
Attempt to build on Push or Pull rather than both


WargodHernandez:
Still executed both push and pull :(

PR #1104, opened 2020-04-06
Use Ui class as member instead of inheriting

asmaloney:


qCC/db_tree/ccPropertiesTreeDelegate.cpp
```diff
@@ -60,6 +60,7 @@
 #include <QCheckBox>
 #include <QComboBox>
 #include <QHBoxLayout>
+#include <QLineEdit>
```
WargodHernandez, line 63: Why was this needed for the inheritance change?

asmaloney: Ah :-)

Because these UI header files include half of Qt (which is one of the reasons to move to the member instead of inheritance), QLineEdit was included indirectly in this file.

So when the UI header is moved out of the .h, there may be some fixes like this.

Issue #1106, opened 2020-04-06
color from scalar fields crash


WargodHernandez:
color from scalar fields crash

 - CC Version (or git hash/tag): v2.11.x
 - Qt Version (if compiling): 5.11.2

If you open the color from scalar fields without having existing colors a crash will occur. (needs to resize the RGB table)

Call stack shown below
(Starting in the OnApply within ccColorFromScalarDlg.cpp)
![image](https://user-images.githubusercontent.com/7495752/78614027-3f090b80-7822-11ea-8db2-40cca0980213.png)

![image](https://user-images.githubusercontent.com/7495752/78613964-11bc5d80-7822-11ea-8a36-10e5bb43d950.png)

![image](https://user-images.githubusercontent.com/7495752/78613818-aa9ea900-7821-11ea-80b5-66ec053d2ba3.png)

WargodHernandez:
Merged fix #1109

PR #1109, opened 2020-04-06
Fixes https://github.com/CloudCompare/CloudCompare/issues/1106

WargodHernandez:
Fixes https://github.com/CloudCompare/CloudCompare/issues/1106

### DIFF
--- a/qCC/ccColorFromScalarDlg.cpp
+++ b/qCC/ccColorFromScalarDlg.cpp
@@ -420,6 +420,10 @@ void ccColorFromScalarDlg::onApply()
 					col[i] = 255 - col[i];
 				}
 			}
+			if (!m_cloud->hasColors())
+			{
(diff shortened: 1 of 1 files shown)


qCC/ccColorFromScalarDlg.cpp
```diff
@@ -420,6 +420,10 @@ void ccColorFromScalarDlg::onApply()
 					col[i] = 255 - col[i];
 				}
 			}
+			if (!m_cloud->hasColors())
```
dgirardeau, line 423: Why not calling it at the beginning of the "onApply" method? Since the other case below ('else' case) has the same issue.

WargodHernandez: I totally missed that, thanks!
https://github.com/CloudCompare/CloudCompare/pull/1114

PR #1111, opened 2020-04-07
Header cleanup

asmaloney:
- use Ui class as member instead of inheriting
- remove unused includes
- use forward declaration


asmaloney:
Only 89 more Ui classes to move out of headers :-)

Issue #892, opened 2019-03-30
Update QCustomPlot to 2.x


asmaloney:
Our included QCustomPlot is version 1.3.2 from 2015 which isn't getting any fixes/updates anymore. We should update to 2.0.1.

This (quite likely) means changes on our side.

In addition, instead of including the code directly, we should consider adding it as a git submodule from here:

https://gitlab.com/DerManu/QCustomPlot

dgirardeau:
Oh yes indeed!

I did that a long time ago, and never really used it that much so I simply forgot about it in a way :D

PR #1118, opened 2020-04-07
Update QCustomPlot to 2.0.1

asmaloney:
The API changed between QCustomPlot 1.x and 2.x. 

This PR requires careful review - there are a couple of changes I'm unsure about.

This is the type of PR you might want to pull and test. The things I was able to test seem to work fine...

(Looks much cleaner on high DPI screens!)


qCC/ccQCustomPlot.h
```diff
@@ -126,14 +125,15 @@ class QCPColoredBars : public QCPBars
 			if (colors.size() > i)
 				newData.color = colors[i];
 			m_coloredData.insertMulti(newData.key, newData);
-			mData->insertMulti(newData.key, newData);
+//			data()->insertMulti(newData.key, newData);
+			QCPBars::addData(newData.key,newData.value);
```
asmaloney, line 129: This is the main thing I wasn't 100% sure of. `insertMulti` doesn't exist, and I'm now only adding the double, not the struct - and I'm doing it through the QCPBars class, not the data().

qCC/db_tree/sfEditDlg.cpp
```diff
@@ -44,19 +44,15 @@ sfEditDlg::sfEditDlg(QWidget* parent/*=0*/)
...
+		m_associatedSFHisto = new ccHistogramWindow;
 		QHBoxLayout* hboxLayout = new QHBoxLayout(m_ui->histoFrame);
 		hboxLayout->addWidget(m_associatedSFHisto);
 		hboxLayout->setContentsMargins(0, 0, 0, 0);
 		m_associatedSFHisto->enableSFInteractionMode(true);
 		m_associatedSFHisto->xAxis->setTickLabels(false);
-		//m_associatedSFHisto->xAxis->setAutoSubTicks(false);
-		//m_associatedSFHisto->xAxis->setSubTickCount(0);
-		m_associatedSFHisto->xAxis->setAutoTickCount(6);
+		m_associatedSFHisto->xAxis->ticker()->setTickCount(6);
```
asmaloney, line 53: I was also not 100% on these conversions from `setAutoTickCount`. They moved from setting it directly to a ticker class that doesn't quite have the same things (as far as I can tell).

asmaloney:
When complete, this will (mostly) address #892.

It required one fix to QCustomPlot, so moving to their repo won't work at the moment. The fix has been submitted to them.

qCC/db_tree/sfEditDlg.cpp
dgirardeau: Same thing, it seems to work fine isn't it?

PR #1119, opened 2020-04-07
Add finer-grained control over the SF interaction mode in ccHistogramWindow

asmaloney:
Sam wanted a way to turn off the tools for adjusting the min/max display on the histogram, but allowing interaction with the saturation tools.

This adds flags to let you enable one, both, or none of the interactions.

It also turns off the min/max display for ccColorFromScalarDlg as requested.

### DIFF
--- a/qCC/ccColorFromScalarDlg.cpp
+++ b/qCC/ccColorFromScalarDlg.cpp
@@ -298,7 +298,7 @@ void ccColorFromScalarDlg::updateHistogram(int n)
 
 		//and make histogram grey
 		m_histograms[n]->clear();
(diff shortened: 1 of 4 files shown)


WargodHernandez:
I apparently should have checked Github before doing the same thing, like hours ago lol. Whichever way we want to go is fine, I can modify the PR I just submitted if we want to take this syntax. I left the original intact and added an additional function when you want something other than the original interactors.

I also had a couple small changes which helped a lot with flashing while resizing windows. (I store the min/max & satMin/satMax and then apply them during refresh).

Issue #1117, opened 2020-04-07
ccColorDialog crash if none is current displayed scalar field


WargodHernandez:
**Describe the bug**
System crash if the currently displayed scalar field is set to none.

**How to reproduce**
set the currently displayed scalar field to none from the properties window.

`getCurrentDisplayedScalarField()` will be null in this case which will cause a null de-reference in the ccColorFromScalarDlg constructor
![image](https://user-images.githubusercontent.com/7495752/78698320-06674180-78b7-11ea-9ae8-dfc8b0553de1.png)

Issue #1116, opened 2020-04-07
ccColorFromScalarDlg resize window changes all colours


asmaloney:
**Describe the bug**
If I have R, G, and B showing and in colour, then adjust the window size, the colours all change to the same colour.

Seems to be the last colour that was adjusted.

The same happens for HSV.

**How to reproduce**

I'm using the **pumpACartesian.e57** file from the test data.

Bring up the ccColorFromScalarDlg and adjust the histograms so they all show up in colour:

<img width="614" alt="Screen Shot 2020-04-07 at 11 39 28 AM" src="https://user-images.githubusercontent.com/391371/78689915-d757cc80-78c4-11ea-9eef-925ab886274a.png">

Resize the window and they all turn one colour:

<img width="614" alt="Screen Shot 2020-04-07 at 11 39 48 AM" src="https://user-images.githubusercontent.com/391371/78689930-db83ea00-78c4-11ea-9c4e-17f16e1586f0.png">

**Expected behaviour**

Colour should remain as it was for each histogram.

**Your environment**
 - CC Version (or git hash/tag): master
 - OS & Version: macOS
 - Qt Version (if compiling): 5.12.7

Issue #1110, opened 2020-04-07
ccColorFromScalarDlg changes the scalar field color scale 


WargodHernandez:
ccColorFromScalarDlg changes the scalar field color scale even when cancel is pressed.

Open the dialog and press cancel to see the following
![image](https://user-images.githubusercontent.com/7495752/78617124-3537d600-782b-11ea-9d05-4d9e22207cda.png)
Re-select the color scale to restore the previous colors
![image](https://user-images.githubusercontent.com/7495752/78617167-61ebed80-782b-11ea-976e-f42597c0ef4e.png)

The scalar field color scale should be restored when the dialog is closed either by applying or canceling.

PR #1121, opened 2020-04-08
ccColorFromScalarDlg modifications

WargodHernandez:
I took a swing at a few of the issues in the new ccColorFromScalarDlg tool.

This works for me, but I would appreciate other people giving it a tryout because I modified `ccHistogramWindow ` (it shouldn't change any existing code, I just added a way to hide specific SFInteraction tools, and to store the min, max, saturationMin, and saturationMax, and restore the values during refresh to cut down the flashing that can occur)

fixes https://github.com/CloudCompare/CloudCompare/issues/1110
fixes https://github.com/CloudCompare/CloudCompare/issues/1116
fixes https://github.com/CloudCompare/CloudCompare/issues/1117


qCC/ccColorFromScalarDlg.cpp
```diff
@@ -44,12 +44,21 @@ ccColorFromScalarDlg::ccColorFromScalarDlg(QWidget* parent, ccPointCloud* pointC
 {
 	m_ui->setupUi(this);
 	
+	
+	if (m_cloud->getCurrentDisplayedScalarFieldIndex() == -1)
```
dgirardeau, line 48: Should we test that there are any scalar field at all first? like cloud->hasScalarFields() or getScalarFieldCount() (can't remember)

qCC/ccColorFromScalarDlg.cpp
```diff
@@ -44,12 +44,21 @@ ccColorFromScalarDlg::ccColorFromScalarDlg(QWidget* parent, ccPointCloud* pointC
...
+	{
+		m_cloud->setCurrentDisplayedScalarField(0);
+	}
+	m_storedOrigColorScale = m_cloud->getCurrentDisplayedScalarField()->getColorScale();
```
dgirardeau, line 52: I would store 'm_cloud->getCurrentDisplayedScalarField()' in a temporary variable to shorten that block a little.

qCC/ccColorFromScalarDlg.cpp
```diff
@@ -142,19 +151,21 @@ ccColorFromScalarDlg::ccColorFromScalarDlg(QWidget* parent, ccPointCloud* pointC
...
-	updateChannel(0);
-	updateChannel(1);
-	updateChannel(2);
-	updateChannel(3);
-	m_cloud->getCurrentDisplayedScalarField()->setColorScale(m_colors[3]); //set grey colour ramp to start with
 }
 
 ccColorFromScalarDlg::~ccColorFromScalarDlg()
 {
+	m_cloud->getCurrentDisplayedScalarField()->setColorScale(m_storedOrigColorScale);
```
dgirardeau, line 161: ++ for a local variable with the current SF ;)

qCC/ccColorFromScalarDlg.cpp
```diff
@@ -142,19 +151,21 @@ ccColorFromScalarDlg::ccColorFromScalarDlg(QWidget* parent, ccPointCloud* pointC
...
+	m_cloud->getCurrentDisplayedScalarField()->setSaturationStart(m_storedOrigSatRange.min());
+	m_cloud->getCurrentDisplayedScalarField()->setSaturationStop(m_storedOrigSatRange.max());
+	m_cloud->getCurrentDisplayedScalarField()->setMinDisplayed(m_storedOrigDisplayRange.min());
+	m_cloud->getCurrentDisplayedScalarField()->setMaxDisplayed(m_storedOrigDisplayRange.max());
+	m_cloud->redrawDisplay();
```
dgirardeau, line 166: No need to call 'redrawDisplay' and 'prepareDisplayForRefresh + refreshDisplay'. The first one should be sufficient.

qCC/ccColorFromScalarDlg.cpp
```diff
@@ -332,29 +346,63 @@ void ccColorFromScalarDlg::updateHistogram(int n)
...
 }
 
-//mapping ranges changed
+void ccColorFromScalarDlg::setDefaultSatValuePerChannel(int n)
+{
+	//set default stretch (n.b. this is a cheap hack to avoid calculating percentiles [by assuming uniform data distribution])
+	m_scalars[n]->setColorScale(m_colors[n]);
+	m_minSat[n] = m_scalars[n]->getMin();
+	m_maxSat[n] = m_scalars[n]->getMax();
+	float range = m_maxSat[n] - m_minSat[n];
```
dgirardeau, line 375: Are they really 'float' or ScalarType?

qCC/ccColorFromScalarDlg.cpp
```diff
@@ -332,29 +346,63 @@ void ccColorFromScalarDlg::updateHistogram(int n)
...
+	//Ugly hack to force the plots saturation values to refresh
+	//properly after resize
+	m_scalars[n]->setColorScale(m_colors[n]);
+	double minSat = m_boxes_min[n]->value();
+	double maxSat = m_boxes_max[n]->value();
+	m_histograms[n]->setMinSatValue(minSat + 0.1);
+	m_histograms[n]->setMaxSatValue(maxSat - 0.1);
+	m_histograms[n]->setMinSatValue(minSat - 0.1);
+	m_histograms[n]->setMaxSatValue(maxSat + 0.1);
+	m_histograms[n]->replot();
```
dgirardeau, line 391: Is it necessary to call both replot and repaint? And can't we change the ccHistogram class so as to avoid needing to 'jiggle' these values ;) ?

qCC/ccHistogramWindow.cpp
```diff
@@ -60,10 +61,15 @@ ccHistogramWindow::ccHistogramWindow(QWidget* parent/*=0*/)
 	, m_verticalIndicatorPositionPercent(0)
 	, m_sfInteractionMode(false)
 	, m_selectedItem(NONE)
+	, m_showItems(BOTH_AREAS_AND_ARROWS)
 	, m_areaLeft(nullptr)
+	, m_areaLeftlastValue(std::nan(""))
```
dgirardeau, line 66: I prefer std::numeric_limits'<double'>::quiet_NaN() for the sake of cinsistency

qCC/ccHistogramWindow.cpp
```diff
@@ -91,6 +97,11 @@ ccHistogramWindow::~ccHistogramWindow()
 	clearInternal();
 }
 
+void ccHistogramWindow::setSelectableItemControls(SHOW_SELECTABLE_ITEMS showItems)
```
dgirardeau, line 100: So this is the part which is duplicate with Andy's PR? Not sure how you'll handle that

qCC/ccColorFromScalarDlg.cpp
samthiele: In theory (at least) the tool should not be able to be launched unless the cloud has at least one scalar field. Regardless, it can't hurt to check I suppose :)

qCC/ccColorFromScalarDlg.cpp
WargodHernandez: Honestly no idea, I called it an ugly hack because I tried a lot of different things that seemed like they should work, but didn't... and eventually settled on this. I think I spent most of yesterday messing around with this PR. 
I would prefer anything other than this 'jiggle' values hack, but I'm not sure what causes the issue in the first place considering during resize I would have expected QCustomPlot to take care of it. maybe the new version Andy setup will help with that?

asmaloney: It's possible. There are a lot of changes - the old one is almost 5 years old. And there were changes to the way drawing was done.

I made those changes on master because I thought they were too large for 2.11. I would suggest we (1) sort out the conflicting PRs/solutions, then (2) you try to apply your changes on master w/the updated custom plot.

If that works better than we can look at bringing those changes back to 2.11.

qCC/ccHistogramWindow.cpp
WargodHernandez: I am open going with his, I'm not overly attached to the way I proposed. I figured I'd wait for his response before modifying it again though :)

asmaloney: After discussion in Slack with Chris, I merged my changes and he will adapt his PR.

asmaloney:
Close in favour of #1129?

WargodHernandez:
closing see #1129

PR #1128, opened 2020-04-09
Use Ui class as member instead of inheriting

asmaloney:
This meant moving the dialog code out of the header into a new cpp file.

Also added nanoflann and Hough normals headers to the header list in CMake so they are included in dependencies.

### DIFF
--- a/plugins/core/Standard/qHoughNormals/src/qHoughNormalsDialog.cpp
+++ b/plugins/core/Standard/qHoughNormals/src/qHoughNormalsDialog.cpp
@@ -1,6 +1,6 @@
 //##########################################################################
 //#                                                                        #
-//#                       CLOUDCOMPARE PLUGIN: qHPR                        #
+//#                       CLOUDCOMPARE PLUGIN: qHoughNormals               #
 //#                                                                        #
 //#  This program is free software; you can redistribute it and/or modify  #
 //#  it under the terms of the GNU General Public License as published by  #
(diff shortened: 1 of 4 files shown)


plugins/core/Standard/qHoughNormals/src/qHoughNormalsDialog.cpp
```diff
@@ -0,0 +1,62 @@
+//##########################################################################
+//#                                                                        #
+//#                       CLOUDCOMPARE PLUGIN: qHPR                        #
```
dgirardeau, line 3: Not qHPR

plugins/core/Standard/qHoughNormals/src/qHoughNormalsDialog.cpp
```diff
@@ -0,0 +1,62 @@
...
+//#                                                                        #
+//#                  COPYRIGHT: Daniel Girardeau-Montaut                   #
+//#                                                                        #
+//##########################################################################
+
+#include "qHoughNormalsDialog.h"
+#include "ui_qHoughNormalsDlg.h"
+
+namespace {
+	#ifndef M_PI
```
dgirardeau, line 22: Can't we use the definitions in CCLib for Pi and DEG_TO_RAD / RAD_TO_DEG ?

plugins/core/Standard/qHoughNormals/src/qHoughNormalsDialog.cpp
asmaloney: :-) I copied that from the .h

plugins/core/Standard/qHoughNormals/src/qHoughNormalsDialog.cpp
asmaloney: We could - I was just factoring it out from where it was inline before.

Doing that means we bring in:
```
#include "CCTypes.h"

//system
#include <cfloat>
#include <cmath>
#include <limits>
```

Which doesn't seem worth it to me. This isn't something that never changes (I hope!).

PR #1129, opened 2020-04-10
Master ccColorFromScalarDlg modifications

WargodHernandez:
So here is the ccColorFromScalarDialog modifications similar to PR https://github.com/CloudCompare/CloudCompare/pull/1121

I was able to get rid of the jiggle this time 👍 and it is basically solid without flashing while resizing.

it recalculates and replots significantly less in this version

### DIFF
--- a/qCC/ccColorFromScalarDlg.cpp
+++ b/qCC/ccColorFromScalarDlg.cpp
@@ -41,37 +41,10 @@ ccColorFromScalarDlg::ccColorFromScalarDlg(QWidget* parent, ccPointCloud* pointC
 	: QDialog(parent, Qt::Tool)
 	, m_cloud(pointCloud)
 	, m_ui(new Ui::ColorFromScalarDialog)
+	, m_systemInvalid(false)
 {
 	m_ui->setupUi(this);
 	
-	if (m_cloud->getCurrentDisplayedScalarFieldIndex() == -1)
-	{
-		m_cloud->setCurrentDisplayedScalarField(0);
-	}
-	ccScalarField* sf = static_cast<ccScalarField*>(m_cloud->getCurrentDisplayedScalarField());
-	if (!sf) // I had this happen 1 time during testing but could never replicate
-	{
-		assert(false);
-		ccLog::Error("[ccColorFromScalarDlg] Get current scalar field failed!");
-		return; 
-	}
-	m_storedOrigColorScale = sf->getColorScale();
-	m_storedOrigSatRange = sf->saturationRange();
-	m_storedOrigDisplayRange = sf->displayRange();
-	//create histograms
-	QFrame* histoFrame[c_channelCount] = { m_ui->histoFrameR, m_ui->histoFrameG, m_ui->histoFrameB, m_ui->histoFrameA };
-	for (unsigned i = 0; i < c_channelCount; i++)
-	{
-		m_histograms[i] = new ccHistogramWindow(this);
-		m_histograms[i]->setRefreshAfterResize(false);
-		auto layout = new QHBoxLayout;
-		
-		layout->setContentsMargins( 0, 0, 0, 0 );
-		layout->addWidget( m_histograms[i] );
-		
-		histoFrame[i]->setLayout( layout );
-	}
-
 	//populate boxes
 	m_boxes_min[0] = m_ui->minInputSpinBoxR;
 	m_boxes_max[0] = m_ui->maxInputSpinBoxR;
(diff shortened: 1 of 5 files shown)


qCC/mainwindow.cpp
```diff
@@ -839,6 +839,7 @@ void MainWindow::doActionColorFromScalars()
 			ccColorFromScalarDlg* cfsDlg = new ccColorFromScalarDlg(this, cloud);
 			cfsDlg->setAttribute(Qt::WA_DeleteOnClose, true);
 			cfsDlg->show();
+			cfsDlg->setupDisplay(); //required otherwise the initial histograms are grey
```
asmaloney, line 842: Kind of nitpicky, but it feels like this is in the wrong place. If the dialog doesn't work properly without this being called, then it should be called somewhere in the dialog class.

Does it work if `setupDisplay` is called at the end of the constructor?

If not, we should override show() and call it there the first time the dialog is shown.

Now - we only use this dialog here, so if you think it's overkill to fix properly I don't feel too strongly - about a 5.75 on this one. :-)

(In fact that `setAttribute` should be done in the constructor as well.)

WargodHernandez: Easy fix. I had fixed the original issue that formerly required this. I was able to just remove it.

I also moved the setAttribute to the constructor

qCC/ccColorFromScalarDlg.cpp
```diff
@@ -88,18 +102,19 @@ ccColorFromScalarDlg::ccColorFromScalarDlg(QWidget* parent, ccPointCloud* pointC
...
 			m_combos[i]->addItem(m_cloud->getScalarFieldName(s));
 		}
 		m_combos[i]->setCurrentIndex( m_cloud->getCurrentDisplayedScalarFieldIndex() );
 	}
-
-	m_ui->fixA->setChecked(true); //set alpha to checked
-
+	
+	m_ui->fixA->setChecked(true); //set alpha fixed to checked
+	m_prevFixed[3] = false;
```
dgirardeau, line 117: is 3 in fact 'c_channelCount-1'?

qCC/ccColorFromScalarDlg.cpp
```diff
@@ -43,13 +43,27 @@ ccColorFromScalarDlg::ccColorFromScalarDlg(QWidget* parent, ccPointCloud* pointC
 	, m_ui(new Ui::ColorFromScalarDialog)
 {
 	m_ui->setupUi(this);
-	
+	this->setAttribute(Qt::WA_DeleteOnClose, true);
+	if (m_cloud->getCurrentDisplayedScalarFieldIndex() == -1)
```
dgirardeau, line 47: I'm still expecting a sanity check on the fact that the cloud has at least one scalar field (I'm not comfortable with making too much assumptions on what has been done by the caller)

qCC/ccColorFromScalarDlg.cpp
```diff
@@ -139,29 +154,39 @@ ccColorFromScalarDlg::ccColorFromScalarDlg(QWidget* parent, ccPointCloud* pointC
...
-	m_cloud->getCurrentDisplayedScalarField()->setColorScale(m_colors[3]); //set grey colour ramp to start with
+	for (unsigned i = 0; i < c_channelCount; i++)
+	{
+		if (sf)
+		{
+			m_scalars[i] = sf;
+			updateChannel(i);
+		}
+	}
+	sf->setColorScale(m_colors[3]); //set grey colour ramp to start with
```
dgirardeau, line 170: Is it once again c_channelCount-1?

qCC/ccColorFromScalarDlg.cpp
```diff
@@ -43,13 +43,27 @@ ccColorFromScalarDlg::ccColorFromScalarDlg(QWidget* parent, ccPointCloud* pointC
 	, m_ui(new Ui::ColorFromScalarDialog)
 {
 	m_ui->setupUi(this);
-	
+	this->setAttribute(Qt::WA_DeleteOnClose, true);
```
asmaloney, line 46: I made a poor suggestion here - my apologies.

Doing it this way means that you can't create one of these dialogs on the stack. (You can, but it will crash when you close it :-) ) It's safer leaving this in mainwindow.

qCC/ccColorFromScalarDlg.cpp
WargodHernandez: I'm open to suggestion on what to do in the case of a failure, the test is easy enough, but on any failure the only way to handle it will be for the caller not to show the dialog at all after construction, which relies on assuming the caller does the error check on the dialog.

qCC/ccColorFromScalarDlg.cpp
WargodHernandez: No problem, easy fix

qCC/ccColorFromScalarDlg.cpp
asmaloney: Maybe put an error at the top of the dialog explaining the problem and disable everything except the Close button?

dgirardeau: Oh it's very defensive, and only to prevent a crash. I would simply return right away (the dialog won't work, that's perfect).

asmaloney: So an assert() for debug mode, and an error to the console so there's some feedback in a release build? (like below)

WargodHernandez: I went through and added error checks and on unrecoverable errors the dialog has all elements other than the close button, I don't display the warning on the dialog itself but ccLog::Error creates a popup and a console message.

I found another bug while going through this, fixed wouldn't work on initial dialog display when the SF Min and Max prevented the default fixed value.

WargodHernandez: I'd say another look at this is required because I was poking around all over the place and changed some order of operations to support the error handling, and I spent a bunch of time trying to track down that fixed value bug, so yeah I'd like another look from you guys :)

dgirardeau: I guess it's ok (even if you've pushed the logic quite far ;). In the case there's no SF, I would have simply not connected the signals / slots so that nothing happens and you don't have to test everywhere that the system is valid... But at least now you have "belt and suspenders" (don't know if that translates ;)

WargodHernandez: That for sure translates, I get accused of that pretty regularly 😀

PR #1135, opened 2020-04-13
Colour from Scalar bugfixes

samthiele:
Fixed the following bugs:

(1) Histograms not refreshing/recalculating when scalar field was changed,
(2) Histograms being recalculated more than needed.
(3) Slider positions resetting every time settings (e.g. reversed) changed on any histogram.

### DIFF
--- a/qCC/ccColorFromScalarDlg.cpp
+++ b/qCC/ccColorFromScalarDlg.cpp
@@ -305,7 +305,9 @@ void ccColorFromScalarDlg::toggleColors(int state)
 	{
 		//update colourmaps
 		updateColormaps();
-		setupDisplay();
+
+		//refresh histograms
+		refreshDisplay();
 	}
 }
 
@@ -315,21 +317,23 @@ void ccColorFromScalarDlg::toggleColorMode(bool state)
 	{
 		//update colourmaps
 		updateColormaps();
-		setupDisplay();
+		refreshDisplay();
 	}
 }
 
-void ccColorFromScalarDlg::setupDisplay()
+// update/redraw histograms but don't reset postion of UI elements/sliders etc. 
+void ccColorFromScalarDlg::refreshDisplay()
 {
 	if (!m_systemInvalid)
 	{
-		//update channels and histograms
+		// refresh histograms
 		for (int i = 0; i < c_channelCount; i++)
 		{
-			updateChannel(i);
+			updateHistogram(i);
 		}
 	}
 }
+
 
 void ccColorFromScalarDlg::updateHistogram(int n)
 {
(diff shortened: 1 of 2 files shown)


samthiele:
Also - I noticed that line 585: 
`m_cloud->setPointColor(p, ccColor::FromQColor(QColor(col[0], col[1], col[2], col[3])));`

doesn't seem to set/change the alpha channel (at least on my computer), while lines 617-618:

```
ccColor::Rgb rgb = ccColor::Convert::hsv2rgb(col[0] * 360.0f, col[1], col[2]);
				m_cloud->setPointColor(p, ccColor::Rgba(rgb, static_cast<int>(col[3] * ccColor::MAX)));
```

do seem to...

Is this an issue with ccColorFromScalarDlg, or something wrong with how the point cloud is handling/setting colours?

samthiele:
Also - I'm finding the tool takes a long time to load for large point clouds. This is presumably because it (1) calculates the minimum/maximum scalar value and (2) generates a histogram 4 times for the currently active scalar ( == 8 complete iterations of the cloud? ) when it creates the red, green, blue and alpha histograms. Is there a way we can perhaps calculate a single histogram and then clone/duplicate it? 

And is there a way to check if we need to recalculate the min/max value of the scalar field (rather than simply doing it every time)?

dgirardeau:
For the issue with QColor, it's because FromQColor returns a 'Rgb' color, and not a 'Rgba' instance. You have to use FromQColora to get a Rgba one ;)

dgirardeau:
Normally the min and max values of scalar fiels should always be up-to-date (no need to compute it each time).

And you can construct an histogram from a 'bin' array (fromBinArray). You cloud compute the first histogram, and the init the others with the bin array from the first histogram.

WargodHernandez:
Because each channel can be set to a different Scalar, copying the first histogram to the remaining channels will be wrong if they aren't all on the same scalar. Or am I mistaken??

asmaloney:
I think - since it's happening in the constructor- they are all the same SF at the start aren't they?

WargodHernandez:
in the constructor that's fine but `m_histograms[n]->fromSF(m_scalars[n], 255, false, true);` was removed from the updateHistogram() function.

We would need to tie the scalar field combo box changed event to a function to run fromSF again to still retain that functionality I believe

WargodHernandez:
also will the interactors still display when using fromBinArray ? I had to modify ccHistogramWindow to see those when using the fromBinArray, but I didn't keep the changes

asmaloney:
Isn't that done in (new) line 459 in _updateChannel_ right before the call to _updateHistogram_? Maybe I'm missing the order of things.

samthiele:
Yes - @asmaloney is more or less correct - I've re-ordered a little to avoid recalculating the histograms all the time. 

In the constructor, the first histogram is created and then its bin array is copied/used to initialise all the other (which use the same default scalar field). The interactors seem to work fine for me. 

Then, if a different scalar is chosen for any field (red, blue, green ..) using the combo box, updateChannel is called, and that's where the new histogram is calculated. 

Does this still work for you @WargodHernandez, or does it break some of your changes?

WargodHernandez:
This all works great, I just got a chance to actually check it out and yeah it is much better! 👍

WargodHernandez:
@samthiele 
> The interactors seem to work fine for me.

This is because when `m_combos[i]->setCurrentIndex(m_cloud->getCurrentDisplayedScalarFieldIndex());` on line 167 in the constructor is called all of the histograms are initialized fromSF in the 
```
	void onChannelChangedR(int) { updateChannel(0); }
	void onChannelChangedG(int) { updateChannel(1); }
	void onChannelChangedB(int) { updateChannel(2); }
	void onChannelChangedA(int) { updateChannel(3); }
```
events, and then overridden by lines 174-180

I just tried commenting out some of those lines and the system still functioned the same. so we should be able to make the following modifications in the constructor to be even more efficient.
 
```C++
//updateChannel(0); //init first histogram
for (unsigned i = 0; i < c_channelCount; i++) //copy data from this histogram into the next ones
{
	//m_scalars[i] = sf;
	setDefaultSatValuePerChannel(i);
	//m_histograms[i]->fromBinArray(m_histograms[0]->histoValues(), m_histograms[0]->minVal(), m_histograms[0]->maxVal());
	updateHistogram(i);
}	
```

asmaloney:
This is precisely where the assert is happening I mentioned in https://github.com/CloudCompare/CloudCompare/issues/1136 and it's because of those combo boxes. 

Let me work on the fix.

PR #1143, opened 2020-04-16
Replace internal CC with CCCoreLib as a submodule

asmaloney:
CCCoreLib lives here: https://github.com/CloudCompare/CCCoreLib


CMakeLists.txt
```diff
@@ -77,14 +77,7 @@ endif()
...
 
-add_subdirectory( CC )
-
-# Propagate the choice from CCLib to use doubles for ScalarType
-if ( ${USE_SCALAR_TYPE_DOUBLE} )
-	add_definitions( -DSCALAR_TYPE_DOUBLE )
-else()
-	add_definitions( -DSCALAR_TYPE_FLOAT )
-endif()
+add_subdirectory( extern/CCCoreLib )
```
dgirardeau, line 80: extern 😢

asmaloney: For cmake projects, this is the recommended place for external repo and library dependencies.

What we have - _contrib_ - is semantically backwards to me. They aren't contributing to our project, we're using their work as an external dependency. It's "external" to the project.

Here's "Modern Cmake"'s take on the structure:

https://cliutils.gitlab.io/modern-cmake/chapters/basics/structure.html

I've also seen 3rdparty/thirdparty, but _extern_ makes the most sense in my opinion.

WargodHernandez: I support the extern concept as well. Third-party especially doesn't make sense with the CCCorelib coming from a first party, but it is coming from an external location.

dgirardeau: It felt weird to see this 'core' library (and from which everything began) being flagged as 'external' 😛

asmaloney: Ah! Now I see why you're crying :-)

It needs to fly on its own! 🦉

PR #1145, opened 2020-04-17
Plugin qAnimation Cubic interpolation

yschaeff:
WARNING: NOT READY TO BE MERGED

I'm in the progress of implementing cubic interpolation for the animation plugin. Currently it uses linear interpolation. I'm about 95% done. Here is an example in a video: https://schaeffer.tk/blob/cc_cubic.mp4 (68 MB)

It still needs 3 things:
- code cleanup / comment / align coding style with CC
- Fix some edge cases (perhaps)
- The View Matrix is still linearly interpolated as transformation matrices. This needs special threatment. I know why, and I think I know how. But I have done it yet.

The reason I'm sharing this is that I seek some feedback from the developers. If you look at my code it'll be fairly obvious I'm NOT a C++ coder. I basically just wrote C in your C++ codebase. And as a result I expect I made some eye hurting mistakes.

Will you consider mentoring me with this? I think it'll be a valuable addition to CC!

### DIFF
--- a/plugins/core/Standard/qAnimation/src/ViewInterpolate.cpp
+++ b/plugins/core/Standard/qAnimation/src/ViewInterpolate.cpp
@@ -21,47 +21,47 @@
 void
 poly_create(double x[4], double y[4], double result[4])
 {
-    double a1 = 2*(x[2]-x[0]);
-    double b1 = 2*(x[2]-x[1]);
-    double c1 = (6 / (double)(x[2]-x[1])) * (y[2] - y[1]) + (6 / (double)(x[1]-x[0])) * (y[0] - y[1]);
-    double a2 = 2*(x[2]-x[1]);
-    double b2 = 2*(x[3]-x[1]);
-    double c2 = (6 / (double)(x[3]-x[2])) * (y[3] - y[2]) + (6 / (double)(x[2]-x[1])) * (y[1] - y[2]);
+	double a1 = 2*(x[2]-x[0]);
+	double b1 = 2*(x[2]-x[1]);
+	double c1 = (6 / (double)(x[2]-x[1])) * (y[2] - y[1]) + (6 / (double)(x[1]-x[0])) * (y[0] - y[1]);
+	double a2 = 2*(x[2]-x[1]);
+	double b2 = 2*(x[3]-x[1]);
+	double c2 = (6 / (double)(x[3]-x[2])) * (y[3] - y[2]) + (6 / (double)(x[2]-x[1])) * (y[1] - y[2]);
 
-    double q = (a2*c1 - a1*c2) / (a2*b1 - a1*b2);
-    double p = (c2 - b2*q) / a2;
+	double q = (a2*c1 - a1*c2) / (a2*b1 - a1*b2);
+	double p = (c2 - b2*q) / a2;
 
-    double a = (p/(6*(x[2]-x[1])));
-    double b = (q/(6*(x[2]-x[1])));
-    double c = (y[1]/(x[2]-x[1]) - (p*(x[2]-x[1]))/6);
-    double d = (y[2]/(x[2]-x[1]) - (q*(x[2]-x[1]))/6);
+	double a = (p/(6*(x[2]-x[1])));
+	double b = (q/(6*(x[2]-x[1])));
+	double c = (y[1]/(x[2]-x[1]) - (p*(x[2]-x[1]))/6);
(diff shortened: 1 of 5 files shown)


WargodHernandez:
Not trying to ignore your pull request, I'm sorry I haven't had time to review this yet. 

Thank you for your effort, I will try to get a look this weekend.

dgirardeau:
Yes, same for me!

And by the way, I started a parallel branch recently where I changed the way the camera management is done to ease the smoothing of the trajectory (even if it's probably done in a different way). See https://github.com/CloudCompare/CloudCompare/tree/new_tools

yschaeff:
That's great. I have some changes here locally which I'll be pushing tonight. However I squashed and resuffled some commits. Maybe this PR will break when I --force it to github. If it does I will make a new PR ready to be reviewed.

yschaeff:
Okay github seems to accept my --force. I removed my local changes that where a hack to make it compile on my machine. So in *theory* this branch would be mergable.

There are some caveats and notes.

- For linear interpolation everything should work exactly as before. I changed some code to figure out the sequence of viewports. The state of current selected viewport in now internal to the qAnimationDlg object. This made the logic for looping much easier for the cubic interpolation case.
- Now, I haven't really found a satisfying solution to the ViewMat interpolation. The ccGLMatrixd::Interpolate() function does the right thing for the linear case but I'm not sure how to translate that to a 3nd order polynomial. So as a stopgap I'm extacting the Euler angles and interpolate those, then rebuild the matrix. In some cases this causes gimbal lock issues and dramatic transitions.
- I think a quaternion based approach might help here. But I haven't been able to tie that together with CC's existing quaternion code. There are multiple transformation matrix types in CC and not all of them support translating to & from quaternions. It didn't feel appropriate to roll my own.
- I fixed a bunch of issues with the dialog selection code where cloudcompare would hang or permanently disable the dialog. - it is a but more fool-proof.
- As said I have next to NULL c++ experience. So I hope I didn't butcher this too much.
- I haven't tested the render() function properly. As the version of CC I was able to compile is quite bare and without FFmpeg. (it is also slow on large pointclouds, so I guess I miss a bunch of harware optimasation stuff)

asmaloney:
Just to throw some more possibilities into the mix - several years ago I wrote something similar to control a camera in an animation. I used [QPropertyAnimation](https://doc.qt.io/qt-5/qpropertyanimation.html) and [QEasingCurve](https://doc.qt.io/qt-5/qeasingcurve.html) to do the interpolation.

This let me easily select any one of the [40 curves](https://doc.qt.io/qt-5/qeasingcurve.html#Type-enum) for interpolation.

(Caveat: I haven't looked at the implementation in this plugin, so I don't know how easy it would be.)

yschaeff:
@asmaloney Interesting. Though (admittetly I skimmed over the links) I do not think that is applicatble here. What you link to are predefined curves. In this PR I create my own curve by finding a 3rth order polynomial through 4 points. This gives not only a continues motion but also a continues (without abrupt changes) acceleration.

The problem is in the orientation of the camera (position is straight forward). If you only allow rotation around the X, Y, Z axis you'll run into gimbal lock issues.

WargodHernandez:
The qeasing curves could possibly be useful still to control speed along your interpolated path.
So rather than a fixed distance per iteration use an optional curve.

asmaloney:
Yes it might be a bit different. This is what I did:

- provided a way to position the camera and save a "view" (camera position, centre, up vector)
- I had an interface to add animation steps (views, pause, some custom stuff)
- each step had a duration, optional text, and a curve

Over the given "duration" it would move the camera using the curve for acceleration, interpolating the position/orientation between the two camera views.

You could set it up, play it, then save it out to a video (.mp4, .avi, or .mkv).

Screen Shot 2020-04-23 at 22 00 44 PM

(I also used ffmpeg to generate the video, but as an external command-line tool, not linked as a library.)

yschaeff:
> The qeasing curves could possibly be useful still to control speed along your interpolated path.
> So rather than a fixed distance per iteration use an optional curve.

Yes. But the point of my proposal is not to use fix distances or fixed speed. Nor are they linear. If two changes are far apart the camera will accelerate or decelerate as a result of followinf the polynomial.

yschaeff:
Earlier I made a python program implementing the cubic interpolation. https://github.com/yschaeff/interpolate if you look at the image you'll get an idea what is happening. It produces a curve segment that connects seamlessly (with the same 1st and 2nd derivative) with the next and previous segment.

PR #1161, opened 2020-04-22
{ci} Switch to ninja on Windows

asmaloney:


.github/workflows/build.yml
```diff
@@ -109,4 +118,5 @@ jobs:
             .
 
       - name: Build
-        run: cmake --build build --parallel --config Release
+        run: |
+          cmake --build build --config Release
```
tmontaigu, line 122: The `--config Release` can also be removed,
 it was needed for Visual Studio because with VS you specify the build type when building not when generating

PR #1163, opened 2020-04-23
Graphical Segmentation tool causes visibility issues

WargodHernandez:
Fixed points disappearing during graphical segmentation that was caused by passing RGB values to OpenGL functions expecting RGBA

See an example of the issue below
image

image

### DIFF
--- a/libs/qCC_db/include/ccColorTypes.h
+++ b/libs/qCC_db/include/ccColorTypes.h
@@ -247,7 +247,7 @@ namespace ccColor
 	//! Conversion from Rgb to Rgba
 	inline Rgba FromRgbToRgba(const Rgb& color)
 	{
-		return Rgba(color.r, color.g, color.b, MAX);
+		return Rgba(color, MAX);
 	}
 
 	//! Conversion from Rgbaf to Rgba
--- a/libs/qCC_db/include/ccColorTypes.h
+++ b/libs/qCC_db/include/ccColorTypes.h
@@ -244,6 +244,12 @@ namespace ccColor
 					static_cast<ColorCompType>(color.b * MAX) );
 	}
 	
+	//! Conversion from Rgb to Rgba
+	inline Rgba FromRgbToRgba(const Rgb& color)
+	{
+		return Rgba(color.r, color.g, color.b, MAX);
+	}
+
 	//! Conversion from Rgbaf to Rgba
 	inline Rgba FromRgbafToRgba(const Rgbaf& color)
 	{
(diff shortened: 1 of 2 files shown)


WargodHernandez:
*Note* This also affects v2.11

libs/qCC_db/src/ccPointCloud.cpp
```diff
@@ -2736,7 +2736,7 @@ void ccPointCloud::drawMeOnly(CC_DRAW_CONTEXT& context)
 							const ccColor::Rgb* col = m_currentDisplayedScalarField->getValueColor(pointIndex);
 							//we force display of points hidden because of their scalar field value
 							//to be sure that the user doesn't miss them (during manual segmentation for instance)
-							glFunc->glColor4ubv(col ? col->rgb : ccColor::lightGreyRGB.rgb);
+							glFunc->glColor4ubv(col ? ccColor::FromRgbToRgba(*col).rgba : ccColor::lightGrey.rgba);
```
dgirardeau, line 2739: wouldn't it be simpler to restore the previous call to 'glColor3ubv'? (since ccColor::lightGrey has an alpha channel equal to 255 anyway).

libs/qCC_db/include/ccColorTypes.h
```diff
@@ -244,6 +244,12 @@ namespace ccColor
 					static_cast<ColorCompType>(color.b * MAX) );
 	}
 	
+	//! Conversion from Rgb to Rgba
+	inline Rgba FromRgbToRgba(const Rgb& color)
```
dgirardeau, line 248: Mind that there's already a constructor of Rgba that takes a Rgb instance as first arugment, and an alpha value as second argument. --> return Rgba(color, MAX);

libs/qCC_db/src/ccPointCloud.cpp
WargodHernandez: I thought you needed to change to 4ubv for AMD support?

libs/qCC_db/include/ccColorTypes.h
WargodHernandez: Fair enough, I didn't notice it I was looking for a conversion function following the pattern of the others, I don't care much either way.

libs/qCC_db/src/ccPointCloud.cpp
dgirardeau: Oh no, that alignment constraints is for VBOs only. Here I just wanted to propagate the support for transparency, but this none of the inputs are 'transparent' here, it was my mistake to convert the call to glColor4ubv.

libs/qCC_db/include/ccColorTypes.h
dgirardeau: Well you can keep the more explicit conversion function. It's just to make the 'return' line more compact and efficient.

libs/qCC_db/src/ccPointCloud.cpp
dgirardeau: Yeah, I would definitely prefer the glColor3ubv option, since we are potentially calling this method millions of time, and here we are basically creating a new instance of Rgba for nothing each time.

PR #1169, opened 2020-04-25
Syntax fix after change of CCLib namespace to CCCoreLib

dgirardeau:


dgirardeau:
Ah that would be great.

asmaloney:
I can do the gamepad one - that's straightforward - but I don't know if I can do the other one. Don't know the libraries I need and if they are available through conda...

dgirardeau:
Hum, that's LASlib... Not sure it's available this way indeed.

PR #1182, opened 2020-04-29
Use 2 decimal places for Plane DB entity direction

EpicWink:


dgirardeau:
Isn't it too much? I'm not sure, since I'm not a geologist myself, but I kind of understood that geologists (who are the primary users of dip and dip direction I believe) only considers integer values.

What is your use case?

EpicWink:
A mathematical tool shouldn't be making decisions on the precision of its results: that's up to the user. Ideally, we would also report the uncertainty of these values, but that's some fairly complex propagation.

My use case is estimating the orientation of box-shaped equipment on a vertical structure. We have quite accurate point-cloud data (from LiDAR), and can get low RMS when fitting the plane. Currently, we resort to using trigonometry on the reported normal

dgirardeau:
My issue is more about the usability, and the fact that we change the way and the purpose this tool was made for originally.

Maybe we could just add the version with several digits next to the integer value (between parentheses)?

EpicWink:
I've made the 2-digit precision direction display another row in the properties viewer

dgirardeau:
Ok

Issue #1174, opened 2020-04-26
CMake Modernization: Plugins


asmaloney:
In my relentless march to modernize cmake usage in the repo, I'm starting in on the biggest section: plugins. This is a short writeup to let you know what I'm thinking. Please add your thoughts if you feel strongly about any of these things.

Right now - like the rest of the repo - they do things kind of backwards: set up a bunch of stuff, declare the target. This needs to be inverted so the target is set up first, then everything hangs off the target. 

My next steps are (roughly):

- [x] Make the plugin API files (currently in `plugins`) into a static library (CCPluginAPI) and move it to the libs directory (partially done locally)
- [x] Once that works, change the new library into a dynamic lib (and figure out installation)
- [x] Work through plugin-by-plugin and switch from GLOB to proper targets, update filesystem layouts, use AUTOUIC, etc.
- [ ] Move logic from CMakePluginTemplate.cmake into an install cmake file that loops over plugins and installs them based on type

(For the next steps in this CMake project - see #1218.)

PR #1183, opened 2020-04-29
{cmake} Make CCPluginAPI a shared lib (step 3)

asmaloney:
This also factors out the logic for installing shared libs (but not plugins yet).

Part of #1174


asmaloney:
I think I have the Windows linking correct now.

Please test this when you get a chance.

PR #1184, opened 2020-04-30
Use new CCCoreLib::radiansToDegrees()

asmaloney:
Part of CloudCompare/CCCoreLib#20

### DIFF
--- a/common/ccCameraParamEditDlg.cpp
+++ b/common/ccCameraParamEditDlg.cpp
@@ -31,7 +31,7 @@
 #include <ccGLWidget.h>
 
 //CCCoreLib
-#include <CCConst.h>
+#include <CCMath.h>
 #include <GenericTriangle.h>
 
 //Qt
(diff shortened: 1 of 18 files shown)


asmaloney:
This one (and the subsequent PRs like it) touch a lot of math. The way I'm doing it with `[[deprecated]]` should prevent me from mixing anything up, but it would be useful to have someone (or more than one person) verify it.

There will be three PRs like this (including this one).

asmaloney:
I'm not sure what you mean by "kind of static" here? They aren't _static_.

[Here's](https://stackoverflow.com/a/22103109) a pretty good translation to English of the standard concerning `static` and `inline`.

Aside: I'm curious where the "capitalize static methods" naming convention came from? I don't think I've seen it before. Do you find it provides additional useful info when reading code? I've probably broken it several times in the past! 😇

dgirardeau:
I mean it's not tied to a class/object instance and you can call it anywhere. It's really useful to be able to spot them (with the capital letter at the beginning) so that you don't need to instantiate an object to call it.

asmaloney:
OK - I still don't see it ("spot them" when reading the code? reading the header? when writing? in IDE completion?), but if you feel strongly about it and want me to change it I will. (My feelings on it are about a 6.7/10 😜)

You should probably update the conventions doc to explain it as well because right now it just refers to "static methods".

dgirardeau:
I would personally prefer it indeed ;) (and yes it's mostly when relying on code completion).

PR #1189, opened 2020-05-03
ShapeFilter: syntax fix + more explicit error messages

dgirardeau:
I got a SHP file from a user with a really misleading error (about missing memory while it was actually a corrupted file). I fixed that, and the the syntax by the same way (to be more compliant with CC's code base ;).

### DIFF
--- a/libs/qCC_io/ShpFilter.cpp
+++ b/libs/qCC_io/ShpFilter.cpp
@@ -712,6 +712,12 @@ static CC_FILE_ERROR ReadMeasures(QDataStream& shpStream, int32_t numPoints, std
 	double mMax;
 	shpStream >> mMin >> mMax;
 
+	if (IsESRINoData(mMin) && IsESRINoData(mMax))
+	{
+		shpStream.skipRawData(8 * numPoints);
+		return CC_FERR_NO_ERROR;
+	}
+
 	try
 	{
 		scalarValues.resize(numPoints, NAN_VALUE);
@@ -721,19 +727,12 @@ static CC_FILE_ERROR ReadMeasures(QDataStream& shpStream, int32_t numPoints, std
 		return CC_FERR_NOT_ENOUGH_MEMORY;
 	}
 
-	if (IsESRINoData(mMin) || IsESRINoData(mMax))
-	{
-		ccLog::Warning("[SHP] Pontentially invalid measurements");
-		//return CC_FERR_MALFORMED_FILE; //apparently we shouldn't treat this as an error
-		return CC_FERR_NO_ERROR;
-	}
-
 	//M values (an array of length NumPoints)
 	for (int32_t i = 0; i < numPoints; ++i)
 	{
 		double m;
 		shpStream >> m;
-		if (!IsESRINoData(m))
+		if (!IsESRINoData(m) && std::isfinite(m))
 		{
 			scalarValues[i] = static_cast<ScalarType>(m);
 		}
--- a/libs/qCC_io/ShpFilter.cpp
+++ b/libs/qCC_io/ShpFilter.cpp
@@ -43,7 +43,6 @@
 //System
 #include <array>
 
-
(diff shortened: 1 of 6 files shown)


libs/qCC_io/ShpFilter.cpp
```diff
@@ -721,68 +701,74 @@ static std::vector<CCVector3> readPoints(QDataStream& shpStream, int32_t numPoin
...
 {
 	int32_t vertCount = lastIndex - firstIndex + 1;
+	if (vertCount <= 0)
+	{
+		assert(false);
+		return nullptr;
+	}
 
-	auto *vertices = new ccPointCloud("vertices");
+	ccPointCloud* vertices = new ccPointCloud("vertices");
```
asmaloney, line 753: I thought when we discussed `auto` way back in the C++11 change, this was acceptable?

Generally this is an acceptable "new C++ style" that [clang static analysis](https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-auto.html) can warn about.

It avoids repeating the type that's already there and makes for easier-to-read code.

```cpp
auto foo = 5; //bad;
auto bar = new blat( "ding" ); //good
```

libs/qCC_io/ShpFilter.cpp
```diff
@@ -794,30 +780,31 @@ ccMesh *createMesh(
...
+			areAllValuesOfPartsNaNs = std::all_of(scalarValues.begin() + firstIndex,
 			                                      scalarValues.begin() + lastIndex,
 			                                      [](ScalarType val)
 			                                      { return std::isnan(val); });
 		}
-		if (!areAllValuesOfPartsNans)
+		if (!areAllValuesOfPartsNaNs)
 		{
-			auto *sf = new ccScalarField("Measures");
+			ccScalarField* sf = new ccScalarField("Measures");
```
asmaloney, line 801: Same as above

libs/qCC_io/ShpFilter.cpp
```diff
@@ -836,9 +823,10 @@ ccMesh *createMesh(
 }
 
 //! Builds the patches that where read from a MultiPatch record of a Shapefile
-//! patches are build as meshes, each part is its own mesh
-//! The only supported Patches/Part types are Triangle_Fan & Triangle_Strip
-CC_FILE_ERROR buildPatches(
+/** Patches are buit as meshes, each part is its own mesh.
+	The only supported Patches/Part types are Triangle_Fan & Triangle_Strip
+**/
```
asmaloney, line 828: Why switch to C-style comments?

Mixing them like this makes it more difficult to read.

libs/qCC_io/ShpFilter.cpp
dgirardeau: I personally think it's less explicit (you have to look at the code on the right to understand the type. And it's particularly 'lazy' when it's not for very long templated types.

libs/qCC_io/ShpFilter.cpp
dgirardeau: Same as above :D

libs/qCC_io/ShpFilter.cpp
dgirardeau: I thought '//!' was for the brief description but apparently it's not... The issue is that it's like that everywhere in the code...

libs/qCC_io/ShpFilter.cpp
asmaloney: You're never going to _only_ read `auto *vertices` or `ccPointCloud* vertices` and then stop. You're going to read the whole line anyway, so it's just extra noise.

dgirardeau: Not always. Depends on the circumstances. Really, for me auto is a shortcut that should only be used when necessary.

dgirardeau: I mean, if you only want to know the type, you'll get the information faster.

PR #1197, opened 2020-05-07
{cmake} Update FBXIO plugin

asmaloney:
I didn't change any of the linking and includes in the main CMakeLists file since I can't compile and test it.

If the SDK provides a proper cmake config file, you should be able to use _find_package_ and then most of the additional FBX stuff in CMakeLists should not be required.


asmaloney:
@dgirardeau - I think you're the only one who can build this plugin.

As I mentioned in the description I can't really clean this up as much as I'd like because I don't have the FBX cmake files to see what they do.

Please let me know if it compiles alright!

dgirardeau:
Ok I'll try that now

dgirardeau:
Ok it seems to work fine!

PR #1199, opened 2020-05-07
Cherry pick from v2.11.x

dgirardeau:
- ShapeFilter syntax fix + more explicit error messages
- qRDBIO plugin update
- syntax updates after change of CCLib namespace to CCCoreLib


plugins/3rdParty/qRDBIO/CMakeLists.txt
```diff
@@ -4,8 +4,11 @@ option( PLUGIN_IO_QRDB "Check to install RDB 2 I/O plugin" OFF )
 
 # CloudCompare RDB 2 I/O plugin
 if (PLUGIN_IO_QRDB)
+
 	project( QRDB_IO_PLUGIN )
 
+	AddPlugin( NAME ${PROJECT_NAME} )
```
asmaloney, line 10: This will need to identify the plugin type which is used for installation of the plugin.

```cmake
AddPlugin( NAME ${PROJECT_NAME} TYPE io )
```

PR #1214, opened 2020-05-12
Use new CCCoreLib::LessThanEpsilon()/CCCoreLib::GreaterThanEpsilon()

asmaloney:
Part of CloudCompare/CCCoreLib#20

### DIFF
--- a/plugins/core/IO/qPDALIO/src/LASFilter.cpp
+++ b/plugins/core/IO/qPDALIO/src/LASFilter.cpp
@@ -362,9 +362,9 @@ CC_FILE_ERROR LASFilter::saveToFile(ccHObject* entity, const QString& filename, 
 
 	//optimal scale (for accuracy) --> 1e-8 because the maximum integer is roughly +/-2e+9
 	CCVector3d diag = bbMax - lasOffset;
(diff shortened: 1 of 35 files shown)


plugins/core/IO/qPDALIO/src/LASFilter.cpp
```diff
@@ -362,9 +362,9 @@ CC_FILE_ERROR LASFilter::saveToFile(ccHObject* entity, const QString& filename,
 
 	//optimal scale (for accuracy) --> 1e-8 because the maximum integer is roughly +/-2e+9
```
WargodHernandez, line 363: Should this comment be updated? I know it's not relevant to your change, just bringing it up

asmaloney: Sorry - I missed this.

I don't know :-) Maybe @dgirardeau has thoughts on it?

I think I'll go ahead with the commit since it's been pending for so long and if something needs to be done with this comment, it can be done separately.

dgirardeau: Ah ah yes, it's probably 1e-9 instead of 1e-8, but that's not a big deal

Issue #1218, opened 2020-05-13
CMake Modernization: Next steps


asmaloney:
Following #1174, I see the next steps looking like this:

- [ ] Move files that are used by both plugins and apps into _CCPluginAPI_ (these are the ones that the Compass, Facets, and SRA plugins use from _Common_ and _qCC_ *see below)
- [ ] Convert _Common_ into _libs/CCAppCommon_ as a shared library (this is where all things shared between the apps, but not used by plugins live)
- [ ] Move _ccViewer_ and _qCC_ into a new _apps_ directory and rename qCC to _CloudCompare_
- [ ] Reorganize the ccViewer and CloudCompare directories (may require discussion!)
- [ ] Rework the cmake installation (use windeployqt and macdeployqt properly, move and clean up cmake installation code)

Files to move into _CCPluginAPI_ (I'm not yet sure how tricky this will be...):
- common/ccOverlayDialog.[h,cpp]
- common/ccPickingListener.h
- common/ccPickingHub.[h,cpp]
- common/ccQtHelpers.h
- {CC source}/ccColorScaleSelector.[h,cpp]
- {CC source}/ccColorScaleEditorDlg.[h,cpp]
- {CC source}/ccColorScaleEditorWidget.[h,cpp]
- {CC source}/ccRenderToFileDlg.[h,cpp]
- {CC source}/ui_templates/renderToFileDialog.ui
- {CC source}/ui_templates/colorScaleEditorDlg.ui

Please add your thoughts if you feel strongly about any of these things.

PR #1223, opened 2020-05-16
Move some files from "qCC" into the CCPluginAPI lib

asmaloney:
- {CC source}/ccColorScaleSelector.[h,cpp]
- {CC source}/ccColorScaleEditorDlg.[h,cpp]
- {CC source}/ccColorScaleEditorWidget.[h,cpp]
- {CC source}/ui_templates/colorScaleEditorDlg.ui

...and because it's needed by ccColorScaleEditorDlg: ccPersistentSettings.h

Also changed ccColorScaleEditorDlg from inherited Ui to member variable so we don't need the ui header file exposed.

This is part of #1218


dgirardeau:
Hum, actually I wanted to create a big library below 'qCC' with all the dialogs, semi-algorithms, etc.

But you can go on with this PR. We'll move the dialog when this big library will be created.

asmaloney:
At the end of the day we may be approaching the same problem from different ends.

What I'm trying to do is separate out the things that plugins can use from the things that are application-specific. In my view, the plugins shouldn't reach into the CC source for anything.

When I'm done this series of PRs (1 more after this I think), the plugins will depend only on the libraries in "libs". (This bundle of libs would effectively be a large-ish plugin SDK...)

What I see is the following:

- CCAppCommon: common code for applications (old "common" - plugins do not use this)
- CCPluginAPI: code that applications & plugins use (added to over time)

I agree that the CloudCompare stuff needs to reorganize (see step in #1218).

I have thoughts on a couple of different ways to reorganize what's currently in qCC - maybe that's best discussed in #1218.

dgirardeau:
Since ccColorScaleSelector will also be used by CC, isn't it a bit weird to have CloudCompare rely on CCPluginAPI?

For me this big 'dialogs' lib will indeed go under 'libs'. And plugins (or any other project) will be of course able to reuse any of them.

dgirardeau:
And I'd like all dialogs to be at the same place if possible (otherwise it's a bit strange for me to have dialogs in various places)

asmaloney:
> Since ccColorScaleSelector will also be used by CC, isn't it a bit weird to have CloudCompare rely on CCPluginAPI?

Maybe the name needs to change. It's more of an SDK that both applications and plugins rely on. That's where dialogs common to CC, ccViewer, and plugins would live.

Any dialogs that are only used by CC actions, for example, would stay in CC. Again - I have ideas on how that might be organized (and I know @WargodHernandez does as well) so that's best discussed in the issue.

Issue #1224, opened 2020-05-17
Architecture problem: Q_OBJECT in static libs


asmaloney:
We have two libraries - CCFbo and qCC_glWindow - that are created as static libs which also use Q_OBJECT.

Each thing (application/library) they are linked to ends up getting their own qobject metadata. One implication is that _qobject_cast_ may fail when crossing dynamic library boundaries.

I'm kind of surprised we haven't run into this before (or maybe we have and not recognized it!).

**Solution:** CCFbo and qCC_glWindow should be changed to dynamic libs.

PR #1225, opened 2020-05-17
Changes CCFbo to a dynamic library

asmaloney:
Half of #1224


libs/CCFbo/CMakeLists.txt
```diff
@@ -1,8 +1,10 @@
...
 
-add_library( ${PROJECT_NAME} STATIC )
+add_library( ${PROJECT_NAME} SHARED )
 
 add_subdirectory( include )
 add_subdirectory( src )
 
+target_compile_definitions( ${PROJECT_NAME} PRIVATE CCFBO_LIBRARY_BUILD )
+
 target_link_libraries( ${PROJECT_NAME} Qt5::Core Qt5::OpenGL Qt5::OpenGLExtensions )
```
dgirardeau, line 10: Shouldn't you call InstallSharedLibrary as well?

asmaloney: Absolutely I should!

Thanks.

PR #1232, opened 2020-05-20
Add logging

dsame:
The key changes:

https://github.com/dsame/CloudCompare/blob/f19658228050c83f193a45e4af22b89e015b1332/.github/workflows/build.yml#L79 - Ensure the execution is not aborted on the first stderr write

https://github.com/dsame/CloudCompare/blob/f19658228050c83f193a45e4af22b89e015b1332/.github/workflows/build.yml#L122 - `2>&1` - don't interrupt normal execution even if something is written to stderr

https://github.com/dsame/CloudCompare/blob/f19658228050c83f193a45e4af22b89e015b1332/.github/workflows/build.yml#L125 - print ErrorRecords in readable format

Also there's `catch` block that is not expected to be run normally, but just in case there's a code to  show the cmake logs

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -124,9 +124,9 @@ jobs:
             $i = 0
             $Output | % {
               if ($_.GetType().Name -eq 'ErrorRecord') {
-                Write-Host "${i}: $_.Exception.Message"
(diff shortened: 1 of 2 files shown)


asmaloney:
Thanks @dsame!

Superseded by #1234 to switch to bash.

PR #1233, opened 2020-05-21
Use bash for CMake configure step

dsame:


asmaloney:
Thanks a lot @dsame!

Superseded by #1234 to fix merge conflicts and handle env vars properly. We'll see how it goes...

PR #1245, opened 2020-05-22
Various fixes

dgirardeau:


plugins/3rdParty/qRDBIO/CMakeLists.txt
```diff
@@ -16,8 +16,17 @@ if (PLUGIN_IO_QRDB)
 	# cmake variable rdb_DIR
 	find_package(rdb REQUIRED)
 
-	# Set to include this I/O plugin in ccViewer
-	set( CC_IS_IO_PLUGIN 1 )
+	file( GLOB PLUGIN_HDR_LIST ${CMAKE_CURRENT_SOURCE_DIR}/*.h )
```
asmaloney, line 19: GLOB shouldn't be used anymore (except for very specific circumstances). It doesn't allow dependencies to work properly.

You should explicitly list the sources as in the example plugin.

plugins/core/Standard/qPCL/PclUtils/filters/BaseFilter.h
```diff
@@ -184,7 +184,7 @@ class BaseFilter: public QObject
 	//! Called when action is triggered
 	/** \note performAction calls start() in base class
 	**/
-	int performAction();
+	Q_SLOT int performAction();
```
asmaloney, line 187: As discussed in Slack, the correct solution is to convert the connects to Qt5-style connects.

plugins/3rdParty/qRDBIO/CMakeLists.txt
dgirardeau: That's a pain, especially for plugins with a lot of files (not this one, but I have others). I didn't understood that we lost this capacity of the previous plugin script.

plugins/core/Standard/qPCL/PclUtils/filters/BaseFilter.h
dgirardeau: Has it been done? If not, I just want to make this plugin compile for now...

plugins/3rdParty/qRDBIO/CMakeLists.txt
asmaloney: The CMake devs explicit recommend not to use GLOB for sources because it causes dependency problems.

What I did to make lists with lots of files is just select the files in the Finder and copy. That copies the file names as a list. Then paste that in and copy-paste the prefix for the line (or use an editoe with a tool to "prefix lines").

Can Windows do something similar? Or use the command line to `ls` (or equivalent) and use that list?

plugins/core/Standard/qPCL/PclUtils/filters/BaseFilter.h
asmaloney: Can't you just change the connects instead?

plugins/3rdParty/qRDBIO/CMakeLists.txt
dgirardeau: Nope :|. Well of course you can always use a console and copy the list indeed.

But it means also we need to update this file anytime we change a name, or add a file. And that's super not efficient... Especially since it was working fine.

plugins/core/Standard/qPCL/PclUtils/filters/BaseFilter.h
dgirardeau: Don't really have the time for that (I'm struggling like hell to rebase my most important branch right now with all the recent changes)

plugins/3rdParty/qRDBIO/CMakeLists.txt
asmaloney: > it means also we need to update this file anytime we change a name, or add a file

That's the whole point. Doing so maintains the dependencies properly.

> Especially since it was working fine.

For you on your setup (maybe).

At the end of the day you can do whatever you want with your plugins :-) This one (RDBIO) shouldn't even be in this repo.

dgirardeau: What wasn't working with the plugins before? Apart that it was not 'clean'?

qRDBIO is useful to the community, and could be moved to the 'core' plugins. It was put in '3rdParty' because it was originally developed by a '3rd party'.

dgirardeau: It is not "my" plugin :|

asmaloney: > What wasn't working with the plugins before?

Many times I'd make a change and cmake would not re-run itself because the dependency didn't exist. So I'd get strange compile errors. Done properly, cmake is aware of the dependencies and can keep itself up to date.

> qRDBIO

Does qRDBIO not fit with qRiegl_IO, qFARO_IO, qDotProduct_IO which are not in the repo? They use proprietary libraries and are all Windows-only (I think)?

> It is not "my" plugin

;-) I was referring to the other ones you mentioned with lots of files.

plugins/core/Standard/qPCL/PclUtils/filters/BaseFilter.h
asmaloney: I would do it, but I can't compile them to make sure I get it right.

My IDE has an option that just converts old-style to new style - does that exist in [QtVisualStudioTools2019](https://marketplace.visualstudio.com/items?itemName=TheQtCompany.QtVisualStudioTools2019)?

dgirardeau: I don't use the Qt Visual addin anymore (because it was kind of 'conflicting' with CMake and causing strange issues). Not very keen to go back to it. And I clearly don't have the time for that now.

asmaloney: > it was kind of 'conflicting' with CMake 

Ah - well Qt6 is going to be CMake-based, and they've already made a lot of changes to support cmake with Qt Creator, so they're moving everything that way!

WargodHernandez:
Should this PR be canceled, I believe these fixes are no longer required because of #1246 and #1247

dgirardeau:
Yes indeed

PR #1250, opened 2020-05-23
Update mainWindow.ui

soble:
Add shortcuts
T - segmenT
X - eXtract sections, X-sections
E - Export cloud info


soble:
Replacement patch for #1248 as I don't know my way around git well enough to edit it cleanly

dgirardeau:
Thanks for the contribution. I'll also add it to the 2.11.x branch so that it is active in the soon to be released 2.11 version ;)

PR #1255, opened 2020-05-25
{cmake} Use Qt's (mac|win)deployqt

asmaloney:
Also fixes CCCoreLib dependency installation on Windows

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,7 @@ list( APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/" )
 
 include( CMakePolicies )
 include( CMakeSetCompilerOptions )
+include( DeployQt )
 
 # CCViewer
 option( OPTION_BUILD_CCVIEWER "Check to compile CCViewer project" ON )
@@ -35,10 +36,6 @@ endif()
 
 # Default debug suffix for libraries.
 set( CMAKE_DEBUG_POSTFIX "d" )
-
-# Default install folders
-# (on Windows - msvc generator - the '_debug' suffix is automatically added for debug configurations)
-set( INSTALL_DESTINATIONS CloudCompare )
 
 # RPATH Linux/Unix: (dynamic) libs are put in a subdir of prefix/lib,
 # since they are only used by qCC/ccViewer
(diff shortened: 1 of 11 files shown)


WargodHernandez:
@asmaloney 
I just tested this on Windows and this fixed a problem I previously had with the release mode QT dlls being installed on a debug install, but did introduce a new issue, when in release mode.

From a fresh clone of master after setting the following variables
```JSON
"variables": [
        {
          "name": "QT5_ROOT_PATH",
          "value": "C:/Qt/5.11.2/msvc2017_64/",
          "type": "PATH"
        },
        {
          "name": "Qt5_DIR",
          "value": "C:/Qt/5.11.2/msvc2017_64/lib/cmake/Qt5",
          "type": "PATH"
        }
```
 with a CMAKE_BUILD_TYPE = Debug
Installing would append "d" to all of the libs but would copy the release QT dlls. which would then fail because the app was looking for the d suffix.
image

**After cloning this branch from scratch and following the same steps above the debug install works properly.**

but at this point switching to release mode shows that QTConcurrent wasn't installed

image

This DLL isn't installed during debug either but that is because of the guards against including concurrent in debug mode
```C++
#ifdef CC_CORE_LIB_USES_QT_CONCURRENT
#ifndef CC_DEBUG
//enables multi-threading handling
```

WargodHernandez:
@asmaloney 
It took me a while to find out that windeployqt has some bugs and limitations, even though concurrent was definitely a dependency, windeployqt just never recognizes it as a dependency even though dependencyWalker does show it... following this post https://stackoverflow.com/questions/41579958/windeployqt-does-not-find-all-the-dependencies
According to [QT Windows deployment](https://doc.qt.io/qt-5/windows-deployment.html)
> The application may require additional 3rd-party libraries (for example, database libraries), which are not taken into account by windeployqt.

I modified the post build command in DeployQT.cmake to 
```CMake
add_custom_command(
			TARGET ${DEPLOY_QT_TARGET}
			POST_BUILD
			COMMAND ${CMAKE_COMMAND} -E remove_directory "${temp_dir}"
			COMMAND ${CMAKE_COMMAND} -E make_directory "${temp_dir}"
			COMMAND ${CMAKE_COMMAND} -E copy ${app_path} ${temp_app_path}
			COMMAND "${win_deploy_qt}"
				${temp_app_path}
				--concurrent
				--no-angle
				--no-opengl-sw
				--no-quick-import				
				--verbose=1
			VERBATIM
		)
```

And everything worked correctly.

PR #1261, opened 2020-06-02
fix incorrect openGL color function 2.11.x

WargodHernandez:
This is essentially the same as https://github.com/CloudCompare/CloudCompare/pull/1163/commits/2112ac3c3a9083b1c987f5b22ca74babb0f79afb but I had issues Cherry Picking that commit so I recreated it here.

This fixes http://cloudcompare.org/forum/viewtopic.php?f=10&t=4578 on v2.11.x, already fixed on master.


WargodHernandez:
The 2.11.x branch still has the faulty CI

PR #1274, opened 2020-06-13
RANSAC First cut at making some improvements.

WargodHernandez:
Now you can select certain maximum values for some primitive types with the GUI.
Still needs Torus control and command line support.

Posting this as a draft so I can get more feedback, but in my use cases, this works significantly better than the previous.


plugins/core/Standard/qRANSAC_SD/src/qRANSAC_SD.cpp
```diff
@@ -166,10 +181,28 @@ void qRansacSD::doAction()
...
 
 	if (!rsdDlg.exec())
 		return;
 
+	s_maxSphereRadiusEnabled = rsdDlg.maxSphereRadiuscheckBox->isChecked();
+	s_maxCylinderRadiusEnabled = rsdDlg.maxCylinderRadiuscheckBox->isChecked();
+	s_maxConeRadiusEnabled = rsdDlg.maxConeRadiuscheckBox->isChecked();
+	s_maxConeLengthEnabled = rsdDlg.maxConeLengthcheckBox->isChecked();
+	s_maxCylinderLengthEnabled = rsdDlg.maxCylinderLengthcheckBox->isChecked();
+	s_maxConeAngleEnabled = rsdDlg.maxConeAnglecheckBox->isChecked();
```
rjanvier, line 205: Maybe I would add an option to toggle least square fitting on inliers set, what do you think? I won't have a specific use case in mind for not using the final round of levmar tought.

WargodHernandez: I actually have considered a final fitting option. Specifically the sphere fit by CC after being found by RANSAC have been much better. (I'm finding spheres the are known metrologic references, and the RANSAC returns were typically a few percentage larger than real, where CCs were within the known characteristic limits for the sphere and my measurement device. 
Only sphere fitting in CC is much slower

WargodHernandez: image
in this image the spheres point clouds and the first sphere primitives are RANSAC returns, and the 2nd sphere primitives are using CC fit sphere functions on the sphere point clouds from RANSAC.
The spheres should be 0.5" and 0.25" +/- 0.005" from the manufacturer

rjanvier: Ouch, sphere_0004 fitting by Schnabel's algo is really bad. I though final refinement by least square fitting was enabled by default in the plugin.

WargodHernandez: That is after final refinement, I haven't tried it without it, but I did turn on precision levmar mode to get it that decent. Prior to about a week ago I hadn't ever managed to get anything useful for sphere/cylinder finding with this plugin.

WargodHernandez:
Current state of the user interface looks like this
image

dgirardeau:
Wow, much more options! Nice.

WargodHernandez:
The extra options made a large difference for my use cases, I couldn't really use sphere, cylinder, cone searches before this because they would tend to "overpower" the search for planes

WargodHernandez:
By the way this modified the ransac library to limit the results early on, rather than just throwing away results outside the range at the end. This lets it find the best shapes possible without having to run the entire process multiple times like I was doing before.

Also the save leftover points had a few interesting uses for me like finding all the points which didn't belong a primitive group, I found that it could almost act as an edge detector with the right settings

WargodHernandez:
I don't remember what was left to do on this when I stopped because of getting busy... I'm going to check into it and see if it's worth merging.

WargodHernandez:
So after going through this yesterday, I can see that I hadn't gotten the cylinder length limits, as well as the cone lone/radii limits to work.

I believe I have tracked down why they don't work and but I'll still need some time to decide on a proper solution.

essentially the bitmap primitive shape isn't being cloned at the same time as the associated 3d shape and for cylinder and cones the height is calculated on the bounding box of the bitmap primitive

WargodHernandez:
builds failing due to the version of github actions, see https://github.com/CloudCompare/CloudCompare/commit/3138b1da052f9a84f9ec7105b5bba9877859d5b8

dgirardeau:
Does it mean that we should merge anyway?

WargodHernandez:
well the failed builds are failing on (push) but succeeding on (pull_request).
So the builds on push are using the old github actions, but the pull_request uses the current actions.

So yeah it should be fine.

dgirardeau:
And is everything ready on your side? No more commit ahead?

WargodHernandez:
I wasn't able to find a reasonable way to make cylinder lengths, or any advanced parameters for cones to work properly.
I can possibly support them later so instead of completely removing the related code I hid the UI for options which aren't yet supported. Testing on my side, all the options shown work as expected.

image

dgirardeau:
Have you updated the changelog?

WargodHernandez:
I'll get the change log updated 🤦

PR #1314, opened 2020-08-24
Cherry picks

dgirardeau:
From v2.11

### DIFF
--- a/plugins/core/IO/qAdditionalIO/src/PVFilter.cpp
+++ b/plugins/core/IO/qAdditionalIO/src/PVFilter.cpp
@@ -171,7 +171,7 @@ CC_FILE_ERROR PVFilter::loadFile(const QString& filename, ccHObject& container, 
 	CCCoreLib::NormalizedProgress nprogress(pDlg.data(), numberOfPoints);
 
 	ccPointCloud* loadedCloud = nullptr;
-	CCLib::ScalarField* sf = nullptr;
+	CCCoreLib::ScalarField* sf = nullptr;
 	//if the file is too big, it will be chuncked in multiple parts
 	unsigned chunkIndex = 0;
 	unsigned fileChunkPos = 0;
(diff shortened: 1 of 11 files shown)


plugins/core/IO/qAdditionalIO/src/PVFilter.cpp
```diff
@@ -181,41 +182,40 @@ CC_FILE_ERROR PVFilter::loadFile(const QString& filename, ccHObject& container,
...
 				result = CC_FERR_NOT_ENOUGH_MEMORY;
 				delete loadedCloud;
 				loadedCloud = nullptr;
 				break;
 			}
+			sf = loadedCloud->getCurrentInScalarField();
+			assert(sf);
 		}
 
 		//we read the 3 coordinates of the point
```
WargodHernandez, line 213: I got distracted the last month or so and forgot to send up a PR for the PVfilter changes, I did mostly the same fix, I thought I'd post here so you can decide which you prefer, I think it simplifies things reading 4 values in at once so you don't have to do a NAN fill

```C++
//we read the 3 coordinates of the point and the scalar
		float rBuff[4];
		if (in.read((char*)rBuff, 4 * sizeof(float)) >= 0)
		{
			//conversion to CCVector3
			CCVector3 P((PointCoordinateType)rBuff[0],
						(PointCoordinateType)rBuff[1],
						(PointCoordinateType)rBuff[2]);
			loadedCloud->addPoint(P);
			loadedSF->addElement((ScalarType)rBuff[3]);
		}
		else
		{
			result = CC_FERR_READING;
			break;
		}
```

plugins/core/IO/qAdditionalIO/src/PVFilter.cpp
```diff
@@ -171,6 +171,7 @@ CC_FILE_ERROR PVFilter::loadFile(const QString& filename, ccHObject& container,
 	CCCoreLib::NormalizedProgress nprogress(pDlg.data(), numberOfPoints);
 
 	ccPointCloud* loadedCloud = nullptr;
+	CCLib::ScalarField* sf = nullptr;
```
WargodHernandez, line 174: CCCoreLib

plugins/core/IO/qAdditionalIO/src/PVFilter.cpp
dgirardeau: Yes, definitely simpler!

plugins/core/IO/qAdditionalIO/src/PVFilter.cpp
dgirardeau: (forgot to compile this plugin :D )

PR #1315, opened 2020-08-25
Crash Fix

WargodHernandez:
Noticed an occasional crash when rapidly changing point clouds or scalar fields with accelerator keys in the db tree window. I believe the messages were created while one point cloud or scalar field was selected but then applied to another. debugging lead me to two separate locations which could crash in this situation.

The fix on ccHistogramWindow has one adverse effect, if the situation happens that this code prevents a crash, it will not populate the ccHistogramWindow until you change scalar field or point cloud selection because the message to start it will not be resent. I'm not sure it's worth the effort to have a delay timer or some other mechanism to re-trigger.

### DIFF
--- a/libs/qCC_db/ccPointCloud.cpp
+++ b/libs/qCC_db/ccPointCloud.cpp
@@ -4923,7 +4923,7 @@ bool ccPointCloud::updateVBOs(const CC_DRAW_CONTEXT& context, const glDrawParams
 							for (int j = 0; j < chunkSize; j++, _sf++)
 							{
 								//we need to convert scalar value to color into a temporary structure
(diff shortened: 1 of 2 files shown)


WargodHernandez:
This will need to be Cherry-Picked to master also

dgirardeau:
I agree that a timer would be a little bit overkill (for something not that common ;)

WargodHernandez:
> 
> 
> I agree that a timer would be a little bit overkill (for something not that common ;)

Yeah not common unless like I sometimes do, use the mouse scroll wheel over the dbtree or the up/down keys for entity selection change. I actually had this crash on occasion for as long as I can remember but it was infrequent enough that It was difficult to track down or even explain steps to reproduce.

dgirardeau (approved):

Issue #1319, opened 2020-08-31
Keep comments when loading and saving plys 


eduardoiscar:
**Describe the feature you would like**
When loading a ply file that includes comment fields the comments get printed in the console. After doing some modifications to the file and saving it as a new .ply, the original comments are gone and replaced by:
 _comment Created by CloudCompare v2.12 alpha
 comment Created 31.08.20 11:16._ 

I think it would be very useful if the old comments could be kept, with the new ones simply being appended to the file after saving.

dgirardeau:
Hi,

Well, as CC handles a lot of different formats, once a cloud is loaded, not all the specifics of a particular format are kept. Especially if this information is not re-used by other formats.

eduardoiscar:
Yes, that makes sense. It would still be very helpful, as I have pointclouds with metadata saved in the plys comment section that have to be preserved after cleaning etc.
Would adding a new member variable to ccHObject (such as std::vector<std::string>) that would get populated from the PlyFilter load and save methods be an acceptable solution?

dgirardeau:
No, we have the concept of "meta-data" to associate a point cloud with any kind of data (you just need a unique key, and you can attach to it any type of data thanks to the QVariable interface). So you would have to same the comments as a QStringList and with a key such as "ply.comments". And then at saving time, check if the cloud has some meta-data with the 'ply.comments' key, and use them instead of the default CloudCompare one.

PR #1321, opened 2020-09-01
Issue #1319: Keep comments when loading and saving plys 

eduardoiscar:
Hi Daniel, 

I have implemented the ability to save the ply comments as metadata attached to the pointcloud during loading that gets written back to the ply file when saving it following your recommendations.

Thanks a lot for the support


libs/qCC_io/src/PlyFilter.cpp
```diff
@@ -379,8 +379,21 @@ CC_FILE_ERROR PlyFilter::saveToFile(ccHObject* entity, QString filename, e_ply_s
...
 	}
 	
-	ply_add_comment(ply, qPrintable(FileIO::createdBy()));
-	ply_add_comment(ply, qPrintable(FileIO::createdDateTime()));
+	if (vertices->hasMetaData("ply.comments")) 
+	{
+		QStringList comments = vertices->getMetaData("ply.comments").toStringList();
+		for(int i = 0; i<comments.size(); i++)
+		{
+			ply_add_comment(ply, comments[i].toStdString().c_str());
```
dgirardeau, line 387: I would use the qPrintable() shortcut instead

libs/qCC_io/src/PlyFilter.cpp
```diff
@@ -379,8 +379,21 @@ CC_FILE_ERROR PlyFilter::saveToFile(ccHObject* entity, QString filename, e_ply_s
 		else result = 0;
 	}
 	
-	ply_add_comment(ply, qPrintable(FileIO::createdBy()));
-	ply_add_comment(ply, qPrintable(FileIO::createdDateTime()));
+	if (vertices->hasMetaData("ply.comments")) 
+	{
+		QStringList comments = vertices->getMetaData("ply.comments").toStringList();
+		for(int i = 0; i<comments.size(); i++)
```
dgirardeau, line 385: Maybe use the modern 'for':
```
for (QString comment : comments)
{
    ply_add_comment(ply, qPrintable(comment));
}
```

libs/qCC_io/src/PlyFilter.cpp
```diff
@@ -917,13 +930,15 @@ CC_FILE_ERROR PlyFilter::loadFile(const QString& filename, const QString& inputT
 	/******************/
 	/***  Comments  ***/
 	/******************/
+	QStringList comments = QStringList();
```
dgirardeau, line 933: No need to initialize it:
```
QStringList comments;
```

dgirardeau (changes requested):
Thanks for the fast implementation. See my comments.

eduardoiscar:
Ok, made the suggested changes.

dgirardeau (approved):
Thanks!

Issue #1345, opened 2020-11-03
CloudCompare: error while loading shared libraries: libCCAppCommon.so: cannot open shared object file


theadib:
**Describe the bug**
after fresh make and install CloudCompare does not start. as of: 2020-11-03
```
make
sudo make install
CloudCompare
CloudCompare: error while loading shared libraries: libCCAppCommon.so: cannot open shared object file: No such file or directory
```

The distributed executable does not find the dynamic libraries:
```
adib@vega:~/Projekte/mycc/CloudCompare/build$ ldd /usr/local/bin/CloudCompare 
	linux-vdso.so.1 (0x00007ffc243f5000)
	libCCAppCommon.so => not found
	libQt5PrintSupport.so.5 => /lib/x86_64-linux-gnu/libQt5PrintSupport.so.5 (0x00007f168c94e000)
	libCCPluginAPI.so => not found
	libQCC_GL_LIB.so => not found
	libQCC_IO_LIB.so => not found
	libQCC_DB_LIB.so => not found
	libCCCoreLib.so => not found
```
The dynamic library was pushed to 
-- Installing: /usr/local/bin/libCCAppCommon.so
And the application is in:
/usr/local/bin/CloudCompare

There is no RPATH information in the binary.
https://stackoverflow.com/questions/58997230/cmake-project-fails-to-find-shared-library
```
readelf -d /usr/local/bin/CloudCompare

Dynamic section at offset 0x5d2670 contains 43 entries:
  Tag       Typ                          Name/Wert
 0x0000000000000001 (NEEDED)             Gemeinsame Bibliothek [libCCAppCommon.so]
 0x0000000000000001 (NEEDED)             Gemeinsame Bibliothek [libQt5PrintSupport.so.5]
 0x0000000000000001 (NEEDED)             Gemeinsame Bibliothek [libCCPluginAPI.so]
```

**Root of the problem**
the dynamic libraries MUST go to /usr/local/lib instead /usr/local/bin

**Workaround**
https://blog.andrewbeacock.com/2007/10/how-to-add-shared-libraries-to-linuxs.html
This can only be a temparary workaround until dynamic libraries goes to /usr/local/lib path
```
echo "/usr/local/bin" > cloudcompare.conf
sudo mv cloudcompare.conf /etc/ld.so.conf.d
sudo ldconfig
```
**Environment**
I am using Ubuntu 20.10

theadib:
[...]

Issue #1298, opened 2020-07-21
CloudCompare: error while loading shared libraries: libCCAppCommon.so


DmitryRedin:
Ubuntu 16.04

$git clone ....

$git submodule update --init --recursive

$make
Scanning dependencies of target shapelib
[  0%] Building C object libs/qCC_io/extern/shapelib/CMakeFiles/shapelib.dir/shptree.c.o
[  0%] Building C object libs/qCC_io/extern/shapelib/CMakeFiles/shapelib.dir/shpopen.c.o
[  0%] Building C object libs/qCC_io/extern/shapelib/CMakeFiles/shapelib.dir/safileio.c.o
[  1%] Building C object libs/qCC_io/extern/shapelib/CMakeFiles/shapelib.dir/dbfopen.c.o
[  1%] Linking C static library libshapelib.a
[  1%] Built target shapelib
Scanning dependencies of target CCCoreLib
[  1%] Building CXX object libs/qCC_db/extern/CCCoreLib/CMakeFiles/CCCoreLib.dir/src/AutoSegmentationTools.cpp.o
[  1%] Building CXX object libs/qCC_db/extern/CCCoreLib/CMakeFiles/CCCoreLib.dir/src/BoundingBox.cpp.o
[  1%] Building CXX object libs/qCC_db/extern/CCCoreLib/CMakeFiles/CCCoreLib.dir/src/CCMiscTools.cpp.o
[  2%] Building CXX object libs/qCC_db/extern/CCCoreLib/CMakeFiles/CCCoreLib.dir/src/CCShareable.cpp.o
[  2%] Building CXX object libs/qCC_db/extern/CCCoreLib/CMakeFiles/CCCoreLib.dir/src/ChamferDistanceTransform.cpp.o
[  2%] Building CXX object libs/qCC_db/extern/CCCoreLib/CMakeFiles/CCCoreLib.dir/src/CloudSamplingTools.cpp.o
[  2%] Building CXX object libs/qCC_db/extern/CCCoreLib/CMakeFiles/CCCoreLib.dir/src/Delaunay2dMesh.cpp.o
[  2%] Building CXX object libs/qCC_db/extern/CCCoreLib/CMakeFiles/CCCoreLib.dir/src/DgmOctree.cpp.o
[  3%] Building CXX object libs/qCC_db/extern/CCCoreLib/CMakeFiles/CCCoreLib.dir/src/DgmOctreeReferenceCloud.cpp.o
[...]

PR #1346, opened 2020-11-04
Linux install

luca-penasa:
Hack to solve install issues on linux,  see issues #1345 / #1298.
After much time not compiling cloudcompare on linux I hit this same problem.

I tried to fix the indent (tab vs spaces) but for some reasons, I was not able to do so. Anyway this is just to give an idea of the problem and would be probably better implemented by the original author of the cmake install functions.

Best, 
Luca

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,7 +65,7 @@ if( WIN32 )
         list( APPEND INSTALL_DESTINATIONS ${CCVIEWER_DEST_FOLDER} )
 	endif()
 elseif( UNIX AND NOT APPLE )
-	set( INSTALL_DESTINATIONS ${CMAKE_INSTALL_PREFIX}/bin )
+        set( INSTALL_DESTINATIONS ${CMAKE_INSTALL_PREFIX})
 endif()
 
 # Load advanced scripts
--- a/cmake/Install.cmake
+++ b/cmake/Install.cmake
@@ -19,12 +19,21 @@ function( InstallSharedLibrary )
 			${ARGN}
 	)
 
-	# For readability
-	set( shared_lib_target "${INSTALL_SHARED_LIB_TARGET}" )
-
+	# For readability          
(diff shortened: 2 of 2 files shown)


dgirardeau (approved):
I assume this works on your side. I fully trust you on Linux questions ;)

dgirardeau:
Maybe @rjanvier has some more insights?

luca-penasa:
> I assume this works on your side. I fully trust you on Linux questions ;)

thanks but I see the install system is now a little bit more complex, so I might be missing something of the overall picture... So probably its better to have it checked :-D

luca

dgirardeau:
Okay, let's wait for @rjanvier's opinion then (or maybe @asmaloney?)

rjanvier:
Thanks Luca, I'm not sure it's the way to go, but if it works it's fine for me. On my side, I have no clue of the why of this "new" cmake file and maybe the best for now is to see with its creator @asmaloney .

theadib:
Hi Luca, your solution works for me. Thanks.
Meanwhile I also "fixed" on my side Linux/Ubuntu 20.10/cmake 3.16 like this:
```
diff --git a/cmake/Install.cmake b/cmake/Install.cmake
index d9bfd9ea..b8ea7644 100644
--- a/cmake/Install.cmake
+++ b/cmake/Install.cmake
@@ -27,7 +27,7 @@ function( InstallSharedLibrary )
        foreach( destination ${INSTALL_DESTINATIONS} )                  
                _InstallSharedTarget(
                        TARGET ${shared_lib_target}
-                       DEST_PATH ${destination}
+                       DEST_PATH ${CMAKE_INSTALL_LIBDIR}
                )               
        endforeach()
 endfunction()
```
Adib

dgirardeau:
In the end, should I merge this PR? :D

theadib:
the PR solves for Linux deployments, where the issue has been reported. So it does not break Win, Mac.
+1

PR #1357, opened 2020-11-20
CI: Use conda-incubator/setup-miniconda@v2

tmontaigu:
And use newer thing to set environment variable

Fixes the CI that was failing due to recent changes on the ::set-env commands

https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/


dgirardeau:
Thanks ;)

PR #1361, opened 2020-11-25
bug fix orientation matrix for TLS/GBL sensor during PCD file load

WargodHernandez:
This works for me loading/saving multiple times while changing the translation and orientation arbitrarily, gives expected results.

I also confirmed after opening and saving a PCD the original and new file will be unchanged when viewed through pcl_viewer


dgirardeau (approved):

dgirardeau:
Thanks!

PR #1367, opened 2020-12-02
Local coordinate system primitive

### LINKED ISSUE #1365: Adding local coordinate system objects
See http://cloudcompare.org/forum/viewtopic.php?p=22356#p22356

Here is an example of the local system models I use, 3 planes & 3 polylines which I always move together so they act like a local coordinate system reference. 
This lets me move objects in their local reference system which makes it really easy to adjust multiple objects together, it also allows me to measure clouds easily in their own coordinate system. 
image

I didn't really want to add it to the primitive factory as they pollute the DB tree and you need to keep all 6 objects grouped together for them to make sense. I thought a new primitive would make more sense and could allow you to create the primitive as an automatic child to be turned on/off through the properties window. But still allow them to be select-able for the  Advanced mode translate/rotate form and the measuring modes.

WargodHernandez:
closes #1365

### DIFF
--- a/libs/qCC_db/include/ccCoordinateSystem.h
+++ b/libs/qCC_db/include/ccCoordinateSystem.h
@@ -20,6 +20,8 @@
 //Local
 #include "ccGenericPrimitive.h"
 
+class ccPlane;
+
 //! Coordinate System (primitive)
 /** Coordinate System primitive
 **/
@@ -32,7 +34,15 @@ public:
 		\param transMat optional 3D transformation (can be set afterwards with ccDrawableObject::setGLTransformation)
 		\param name name
 	**/
-	ccCoordinateSystem(PointCoordinateType displayScale, PointCoordinateType axisWidth, const ccGLMatrix* transMat = 0,
+	ccCoordinateSystem(PointCoordinateType displayScale, PointCoordinateType axisWidth, const ccGLMatrix* transMat = nullptr,
+		QString name = QString("CoordinateSystem"));
+
+	//! Default constructor
+	/** Coordinate System is essentially just a way to visualize a transform matrix.
+		\param transMat optional 3D transformation (can be set afterwards with ccDrawableObject::setGLTransformation)
+		\param name name
+	**/
+	ccCoordinateSystem(const ccGLMatrix* transMat,
 		QString name = QString("CoordinateSystem"));
 
 	//! Simplified constructor
@@ -46,6 +56,34 @@ public:
 	//inherited from ccGenericPrimitive
 	virtual QString getTypeName() const override { return "CoordinateSystem"; }
 	virtual ccGenericPrimitive* clone() const override;
+	
+	inline bool axisPlanesAreShown() const { return m_showAxisPlanes; }
+	void ShowAxisPlanes(bool show);
+	inline bool axisLinesAreShown() const { return m_showAxisLines; }
+	void ShowAxisLines(bool show);
+	
+	//ccPlane get2AxisPlane(int axisNum);
+	inline CCVector3 getOrigin() const { return m_transformation.getTranslationAsVec3D(); };
+	inline PointCoordinateType getAxisWidth() const { return m_width; };
+	void setAxisWidth(PointCoordinateType size);
+
+	inline PointCoordinateType getDisplayScale() const { return m_DisplayScale; };
+	void setDisplayScale(PointCoordinateType size);
+
+	ccPlane getXYplane() const;
+	ccPlane getYZplane() const;
+	ccPlane getZXplane() const;
+
+	//! Default Display scale
+	static constexpr PointCoordinateType DEFAULT_DISPLAY_SCALE = 1.0;
+	//! Minimum Display scale
+	static constexpr float MIN_DISPLAY_SCALE_F = 0.001f;
+	//! Default Axis line width
+	static constexpr PointCoordinateType AXIS_DEFAULT_WIDTH = 4.0;
+	//! Minimum Axis line width
+	static constexpr float MIN_AXIS_WIDTH_F = 1.0f;
+	//! Maximum Axis line width
+	static constexpr float MAX_AXIS_WIDTH_F = 16.0f;
 
 protected:
 
@@ -57,12 +95,16 @@ protected:
 	virtual bool fromFile_MeOnly(QFile& in, short dataVersion, int flags, LoadedIDMap& oldToNewIDMap) override;
 	virtual bool buildUp() override;
 
+	ccPlane createXYplane(const ccGLMatrix* transMat = nullptr) const;
+	ccPlane createYZplane(const ccGLMatrix* transMat = nullptr) const;
+	ccPlane createZXplane(const ccGLMatrix* transMat = nullptr) const;
+
 	//! CoordinateSystem options
 	PointCoordinateType m_DisplayScale;
 	PointCoordinateType m_width;
-
+	bool m_showAxisPlanes;
+	bool m_showAxisLines;
 	
-
 
 };
 
--- a/libs/qCC_db/include/ccCoordinateSystem.h
+++ b/libs/qCC_db/include/ccCoordinateSystem.h
@@ -1,0 +1,69 @@
+//##########################################################################
+//#                                                                        #
+//#                              CLOUDCOMPARE                              #
+//#                                                                        #
+//#  This program is free software; you can redistribute it and/or modify  #
+//#  it under the terms of the GNU General Public License as published by  #
+//#  the Free Software Foundation; version 2 or later of the License.      #
+//#                                                                        #
+//#  This program is distributed in the hope that it will be useful,       #
+//#  but WITHOUT ANY WARRANTY; without even the implied warranty of        #
+//#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the          #
+//#  GNU General Public License for more details.                          #
+//#                                                                        #
+//#          COPYRIGHT: Chris Brown                                        #
+//#                                                                        #
+//##########################################################################
+
+#pragma once
+
+//Local
+#include "ccGenericPrimitive.h"
+
+//! Coordinate System (primitive)
+/** Coordinate System primitive
+**/
+class QCC_DB_LIB_API ccCoordinateSystem : public ccGenericPrimitive
+{
+public:
+
+	//! Default constructor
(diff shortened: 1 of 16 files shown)


libs/qCC_db/include/ccCoordinateSystem.h
```diff
@@ -0,0 +1,111 @@
...
+
+#pragma once
+
+//Local
+#include "ccGenericPrimitive.h"
+
+class ccPlane;
+
+//! Coordinate System (primitive)
+/** Coordinate System primitive
```
dgirardeau, line 26: Maybe not necessary ;)

libs/qCC_db/src/ccCoordinateSystem.cpp
```diff
@@ -0,0 +1,262 @@
...
+#include "ccCoordinateSystem.h"
+
+//qCC_db
+#include "ccPlane.h"
+#include "ccPointCloud.h"
+
+
+ccCoordinateSystem::ccCoordinateSystem(PointCoordinateType displayScale, PointCoordinateType axisWidth, const ccGLMatrix* transMat/*= 0*/,
+	QString name/*=QString("CoordinateSystem")*/)
+	: ccGenericPrimitive(name, transMat), m_DisplayScale(displayScale), m_width(axisWidth), m_showAxisPlanes(true), m_showAxisLines(true)
```
dgirardeau, line 27: could you add one initialization per line? (with the comma in front of each one)

libs/qCC_db/src/ccCoordinateSystem.cpp
```diff
@@ -0,0 +1,262 @@
...
+{
+	updateRepresentation();
+	showColors(true);
+}
+
+ccCoordinateSystem::ccCoordinateSystem(const ccGLMatrix* transMat/*= 0*/,
+	QString name/*=QString("CoordinateSystem")*/)
+	: ccGenericPrimitive(name, transMat), m_DisplayScale(DEFAULT_DISPLAY_SCALE), m_width(AXIS_DEFAULT_WIDTH), m_showAxisPlanes(true), m_showAxisLines(true)
```
dgirardeau, line 35: same

libs/qCC_db/src/ccCoordinateSystem.cpp
```diff
@@ -0,0 +1,262 @@
...
+{
+	updateRepresentation();
+	showColors(true);
+}
+
+
+ccCoordinateSystem::ccCoordinateSystem(QString name/*=QString("CoordinateSystem")*/)
+	: ccGenericPrimitive(name), m_DisplayScale(DEFAULT_DISPLAY_SCALE), m_width(AXIS_DEFAULT_WIDTH), m_showAxisPlanes(true), m_showAxisLines(true)
```
dgirardeau, line 43: same

libs/qCC_db/src/ccCoordinateSystem.cpp
```diff
@@ -0,0 +1,262 @@
...
+}
+
+
+void ccCoordinateSystem::setAxisWidth(PointCoordinateType size)
+{
+	if (size >= MIN_AXIS_WIDTH_F && size <= MAX_AXIS_WIDTH_F)
+	{
+		m_width = size;
+	}
+	if (size == 0.0f)
```
dgirardeau, line 66: maybe put that at the top and return (to save some comparisons)

libs/qCC_db/src/ccGenericPrimitive.cpp
```diff
@@ -66,14 +66,33 @@ const ccGenericPrimitive& ccGenericPrimitive::operator += (const ccGenericPrimit
 	unsigned newFacesCount = facesCount + prim.size();
 	bool primHasVertNorms = prim.getAssociatedCloud()->hasNormals();
 	bool primHasFaceNorms = prim.hasTriNormals();
+	bool primHasColors = prim.getAssociatedCloud()->hasColors();
+	
+	if (primHasColors && !verts->hasColors())
+		{
+		if (verts->size() > 0 && !verts->setColor(ccColor::white))
```
dgirardeau, line 73: Is it necessary?

qCC/ccPrimitiveFactoryDlg.cpp
```diff
@@ -158,3 +175,46 @@ void ccPrimitiveFactoryDlg::setSpherePositionToOrigin()
...
+}
+
+void ccPrimitiveFactoryDlg::setCSMatrixToIdentity()
+{
+	csMatrixTextEdit->blockSignals(true);
+	csMatrixTextEdit->setPlainText("1.00000000 0.00000000 0.00000000 0.00000000\n0.00000000 1.00000000 0.00000000 0.00000000\n0.00000000 0.00000000 1.00000000 0.00000000\n0.00000000 0.00000000 0.00000000 1.00000000");
+	csMatrixTextEdit->blockSignals(false);
+}
+
+ccGLMatrix ccPrimitiveFactoryDlg::isCSMatrixValid(bool& valid)
```
dgirardeau, line 206: shouldn't it be called ```getCSMatrix(bool& valid)```?

qCC/db_tree/ccPropertiesTreeDelegate.cpp
```diff
@@ -1529,6 +1554,50 @@ QWidget* ccPropertiesTreeDelegate::createEditor(QWidget *parent,
...
 	break;
+	case OBJECT_COORDINATE_SYSTEM_AXES_WIDTH:
+	{
+		QComboBox* comboBox = new QComboBox(parent);
+
+		comboBox->addItem(tr(s_defaultPolyWidthSizeString)); //size = 0
+
+		for (int i = static_cast<int>(ccCoordinateSystem::MIN_AXIS_WIDTH_F); i <= static_cast<int>(ccCoordinateSystem::MAX_AXIS_WIDTH_F); ++i)
+		{
+			comboBox->addItem(QString::number(i));
```
dgirardeau, line 1565: Why not using a spinbox? (as for the other line/point widths)

qCC/db_tree/ccPropertiesTreeDelegate.cpp
```diff
@@ -1529,6 +1554,50 @@ QWidget* ccPropertiesTreeDelegate::createEditor(QWidget *parent,
...
+		}
+		if (m_currentObject && m_currentObject->isA(CC_TYPES::COORDINATESYSTEM))
```
dgirardeau, line 1567: useless considering the next lines

qCC/db_tree/ccPropertiesTreeDelegate.cpp
```diff
@@ -1529,6 +1554,50 @@ QWidget* ccPropertiesTreeDelegate::createEditor(QWidget *parent,
...
+		outputWidget = comboBox;
+	}
+	break;
+	case OBJECT_COORDINATE_SYSTEM_DISP_SCALE:
+	{
+		QDoubleSpinBox* spinBox = new QDoubleSpinBox(parent);
+		spinBox->setRange(1.0e-3, 1.0e6);
+		spinBox->setDecimals(3);
+		spinBox->setSingleStep(1.0e-1);
+		if (m_currentObject && m_currentObject->isA(CC_TYPES::COORDINATESYSTEM))
```
dgirardeau, line 1587: same

libs/qCC_db/include/ccCoordinateSystem.h
WargodHernandez: actually it is because I later return planes from this primitive so it needs the forward declaration

   ``` c++
    ccPlane getXYplane() const;
	ccPlane getYZplane() const;
	ccPlane getZXplane() const;
```

File: libs/qCC_db/src/ccGenericPrimitive.cpp (hunk 6 above)
Comment (WargodHernandez) [reply libs/qCC_db/src/ccGenericPrimitive.cpp:73]: sort of... you get a warning in the log about resizing the color table for an empty cloud

``` c++
bool ccPointCloud::reserveTheRGBTable()
{
	if (m_points.capacity() == 0)
	{
		ccLog::Warning("[ccPointCloud] Calling reserveTheRGBTable with an zero capacity cloud");
	}
```
This is called down the line when calling `setColor`
without the checks it pollutes the log each time you create a primitive which has color

qCC/ccPrimitiveFactoryDlg.cpp
WargodHernandez: yes, much better, I had originally just intended to check the matrix but it required duplicating most of the function to get it later so I changed it up to return the matrix rather than the bool, but didn't update the name

qCC/db_tree/ccPropertiesTreeDelegate.cpp
WargodHernandez: I was trying to be consistent with the other GL_Lines, polylines width adjustments.
Honestly I would rather use the drop down on the "display scale" if I could know a reasonable range ahead of time.
But I am willing to change it to the spinbox if you prefer it.

qCC/db_tree/ccPropertiesTreeDelegate.cpp
WargodHernandez: Thanks, got it

libs/qCC_db/src/ccCoordinateSystem.cpp
WargodHernandez: fixed

libs/qCC_db/src/ccCoordinateSystem.cpp
WargodHernandez: fixed

libs/qCC_db/src/ccCoordinateSystem.cpp
WargodHernandez: fixed

libs/qCC_db/src/ccGenericPrimitive.cpp
dgirardeau: Oh, I was more asking about the 'setColor' call, but now I realize it's in case we add a primitive without any color...

qCC/db_tree/ccPropertiesTreeDelegate.cpp
dgirardeau: Ah! I thought we were using spinboxes 😅 . Nevermind...

dgirardeau:
I think this PR can be merged. Thanks for the updates.

dgirardeau (approved):

PR #1371, opened 2020-12-07
RDBIO: support attributes with length > 1

NeroBurner:
Handle attributes with length > 1 correctly and make each entry selectable as
a scalar. For an attribute of length 3 this results in three selectable row
entries in the form:
- attribute_name[0]
- attribute_name[1]
- attribute_name[2]


dgirardeau (approved):

NeroBurner:
The number of rows was initialized by the magic number 4, as we have 4 known
non-scalar attributes we want to handle in a different way.

Replace this magic number to make the code more robust in the future.

dgirardeau:
Ok good. Tell me when you are ready to merge.

NeroBurner:
squashed and rebased on current master. I think it's ready to merge

Issue #1380, opened 2021-01-07
Incorrect Angle in Point Picking


neobis001:
**Describe the bug**
Angle C below should be 180 - 76 - 28 = ~76, not 104. Maybe it's adding A and B?

![angle-bug](https://user-images.githubusercontent.com/10540174/103941182-580d5880-50e3-11eb-8890-65e38a38661a.png)

**Your environment**
 - CC Version (or git hash/tag): 2.11.1 (Anoia)
 - OS & Version: Windows 10

WargodHernandez:
@neobis001 thanks for pointing out this error I've submitted a PR with a fix for this, surprising this went unnoticed for so long

PR #1381, opened 2021-01-07
fixes angle info in cc2DLabel

WargodHernandez:
Fixes Issue #1380


dgirardeau (commented):
I hope it doesn't depend on the order of the vertices 😅

WargodHernandez:
> 
> 
> I hope it doesn't depend on the order of the vertices 😅

I checked using the same input the original user had

image
image
image
image
image
image

I think that covers all the vertex pair orders *I hope*

dgirardeau:
Ok perfect! Many thanks for fixing this ;)

PR #1383, opened 2021-01-15
Translate/Rotate Advanced mode coordinate system support

WargodHernandez:
Added the ability to rotate and translate with reference to coordinate system primitive.

Also fixes a few minor bugs with the advanced mode.

### DIFF
--- a/qCC/ccGraphicalTransformationTool.cpp
+++ b/qCC/ccGraphicalTransformationTool.cpp
@@ -26,6 +26,7 @@
 #include <ccMesh.h>
 #include <ccPolyline.h>
 #include <ccPlane.h>
+#include <ccCoordinateSystem.h>
 #include <ccDBRoot.h>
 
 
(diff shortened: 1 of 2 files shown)


qCC/ccGraphicalTransformationTool.cpp
```diff
@@ -137,8 +141,8 @@ void ccGraphicalTransformationTool::advModeToggle(bool state)
 		this->setGeometry(this->x() , this->y(), 0, 0);
 		this->adjustSize(); //adjust size will minimize the display height with the dropdowns not visible
 		this->setGeometry(this->x() + (wPrev - 250), this->y(), 250, this->height());
-		advRotateRefUpdate(0); //index 0 is always the origin
```
WargodHernandez, line 140: I found that the state of the combo box was occasionally out of sync with the internal state because the state update was called, but nothing set the control to the new index, everywhere this happened has been replaced with setting the combo box index which will then trigger the same state update function

dgirardeau: Ah ok. I'm not the one who added this feature, therefore I'll trust your judgement ;)

dgirardeau (approved):

PR #1386, opened 2021-01-18
Add GBL/Camera sensor cloning support

WargodHernandez:
Recently while helping someone with camera sensor manipulations on the [forum](http://cloudcompare.org/forum/viewtopic.php?f=9&t=4979&p=22646#p22646) I needed multiple copies of a camera sensor, and needed to create it multiple times or to save it to a file and reload it multiple times.
This PR just adds Camera & GBL sensors to the list of entities which can be cloned via the clone button.


dgirardeau (approved):
Thanks ;)

Issue #1387, opened 2021-01-19
InitOpenGL does not take place with translations


WargodHernandez:
in main.cpp we determine if we are in command line mode and if not we setup openGL which must be done before initializing ccApllication 
```C++
bool commandLine = (argc > 1) && (argv[1][0] == '-');
#endif
   
	if ( !commandLine )
	{
		ccApplication::initOpenGL();

#ifdef CC_GAMEPAD_SUPPORT
		QGamepadManager::instance(); //potential workaround to bug https://bugreports.qt.io/browse/QTBUG-61553
#endif
	}

	ccApplication app(argc, argv, commandLine);

```

Later we check if the -LANG argument was passed and if so just set command line to false

```C++
//specific commands
	int lastArgumentIndex = 1;
	QTranslator translator;
	if (commandLine)
	{
		//translation file selection
		if (QString(argv[lastArgumentIndex]).toUpper() == "-LANG")
		{
			QString langFilename = QString::fromLocal8Bit(argv[2]);

			//Load translation file
			if (translator.load(langFilename, QCoreApplication::applicationDirPath()))
			{
				qApp->installTranslator(&translator);
			}
			else
			{
				QMessageBox::warning(nullptr, QObject::tr("Translation"), QObject::tr("Failed to load language file '%1'").arg(langFilename));
			}
			commandLine = false;
			lastArgumentIndex += 2;
		}
	}
```

In this case the `ccApplication::initOpenGL();` call is skipped completely, looking into that function it seems that macOSx will possibly have issues when using translations

```C++
void ccApplicationBase::initOpenGL()
{
	//See http://doc.qt.io/qt-5/qopenglwidget.html#opengl-function-calls-headers-and-qopenglfunctions
	/** Calling QSurfaceFormat::setDefaultFormat() before constructing the QApplication instance is mandatory
		on some platforms (for example, OS X) when an OpenGL core profile context is requested. This is to
		ensure that resource sharing between contexts stays functional as all internal contexts are created
		using the correct version and profile.
	**/
```

Not sure if anyone who can test this on OS X (maybe @asmaloney ?)

WargodHernandez:
It looks like full screen mode on windows would also be affected
```C++
// The 'AA_ShareOpenGLContexts' attribute must be defined BEFORE the creation of the Q(Gui)Application
	// DGM: this is mandatory to enable exclusive full screen for ccGLWidget (at least on Windows)
[...]

PR #1388, opened 2021-01-19
Translation command line bug fix

WargodHernandez:
call `ccApplication::initOpenGL();` whether or not in command line mode.

Also adds `ccTranslationManager::loadTranslation(QString language);` to simplify loading language from the command prompt

`C:\Program Files\CloudCompare>CloudCompare.exe -LANG translations\cloudcompare_fr`

was the previous command for french

`C:\Program Files\CloudCompare>CloudCompare.exe -LANG fr`

is the new way handled by the translation manager

### DIFF
--- a/qCC/main.cpp
+++ b/qCC/main.cpp
@@ -45,6 +45,7 @@
 #include "ccGuiParameters.h"
 #include "ccPersistentSettings.h"
 #include "mainwindow.h"
+#include "ccTranslationManager.h"
 
 //plugins
 #include "ccPluginInterface.h"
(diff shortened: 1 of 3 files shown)


qCC/main.cpp
```diff
@@ -89,14 +90,11 @@ int main(int argc, char **argv)
 	bool commandLine = (argc > 1) && (argv[1][0] == '-');
 #endif
    
-	if ( !commandLine )
```
dgirardeau, line 92: Actually, I'm not sure to see what was the advantage of removing this test? Does it have an impact on the translation stuff?

WargodHernandez: When starting cloud compare from the command line with a supplied -LANG argument it sets commandLine true, which prevents running the InitOpenGL call. That causes going into full screen mode to hard crash.
See #1387

dgirardeau: Ah yes, sorry. Can't we just add a specific test case against a single '-LANG' command? Or avoid using '-'?

WargodHernandez: we can if we want. 
looking at the code executed in InitOpenGL didn't seem to have any real impact unless you later created a GL window or widget, I couldn't find any connection to OpenGL directly, instead it seems to be configuring QT to deal with OpenGL later.

I'm ok with whichever method you prefer. :)

dgirardeau: Yes, that's probably right. I'm just a little bit fearfull of changing something that might impact the command line mode (while not being 'absolutely required').

On the other hand, we'll hear about users soon enough 😈 . Let's go with this version!

PR #1392, opened 2021-01-31
Add Chinese Translation for CloudCompare

jindili:
Add Chinese Translation for the Main Window, Almost All Menu Item and Toolbar(Except some plugins) are translated.


dgirardeau:
Thanks!

PR #1403, opened 2021-02-21
STEP file I/O plugin

dgirardeau:


WargodHernandez:
I haven't looked at the code yet obviously but STEP support would be an amazing addition, would it actually support Brep curves/surfaces? Or exclusively points and currently supported primitives.

dgirardeau:
For now it's just tesselating everything to a triangle mesh. But I guess it's definitely possible to properly import existing primitives as primitives (OpenCascade is doing all the work).

WargodHernandez:
But it is at least tesselating surfaces then, that's great. Maybe we can tesselate the curves to polyline primitives.

PR #1408, opened 2021-02-24
Single click picking option

WargodHernandez:
Adding an option to disable single click picking. 
Single click picking can be really slow on large clouds.
See forum request [Problems with a single click](http://cloudcompare.org/forum/viewtopic.php?f=9&t=5018&p=22891#p22891)


dgirardeau (approved):
Looks good!

PR #1436, opened 2021-03-29
Don't set a default color for console message item

tmontaigu:
The color used will be the default one for the current theme.

Otherwise setting the color to black overrides the default color of the
theme and it doesn't play well with dark themes.


dgirardeau (approved):

dgirardeau:
Same question

tmontaigu:
This can be merged

Issue #1415, opened 2021-03-05
Version is not displayed in the 'About' dialog


ryanthara:
**Describe the bug**

The current version is not shown in the 'About' dialog.

**How to reproduce**

Open the 'About' dialog.

**Expected behaviour**

Show the current version (major.minor.bugfix) in the bottom of the dialog window.

<img width="712" alt="Bildschirmfoto 2021-03-05 um 14 12 46" src="https://user-images.githubusercontent.com/7117072/110120086-de30dd00-7dbc-11eb-8d4a-9b6cfcc2db7e.png">

**Additional context**

**Your environment**
 - CC Version (or git hash/tag): 2.11.1 (Anoia)
 - OS & Version: macOS Big Sur 11.2.2
 - Graphics card: 
 - Qt Version (if compiling):

tmontaigu:
I would say it is likely a problem caused by the  Qt version used in the CloudCompare macOs App being outdated for macOs 11.2.2

Also are you using a mac with the new Apple Silicon processor ?

ryanthara:
> I would say it is likely a problem caused by the Qt version used in the CloudCompare macOs App being outdated for macOs 11.2.2

Ok, then a newer version would be great. :)

> Also are you using a mac with the new Apple Silicon processor ?

Unfortunately no.

tmontaigu:
I can reproduce on a mac with Big Sur 11.3 beta

dgirardeau:
I tried to fix this issue by explicitly forcing the default text color to black: 2ae7b490d9b07cdc2544b1dd117075cc9189f7ed

I'll wait for your feedback to see if this helps?

ryanthara:
> I tried to fix this issue by explicitly forcing the default text color to black: [2ae7b49](https://github.com/CloudCompare/CloudCompare/commit/2ae7b490d9b07cdc2544b1dd117075cc9189f7ed)

I think it could come from the 'dark mode' on macOS Big Sur (11.2.2 here).  
 
> I'll wait for your feedback to see if this helps?

This is how it looks in the 'light mode'.

<img width="712" alt="Bildschirmfoto 2021-03-07 um 21 27 00" src="https://user-images.githubusercontent.com/7117072/110253711-23573980-7f8c-11eb-8440-31c950e38c14.png">

Should the behavior of color selection not done by Qt functions automatically, instead of hard coding?

tmontaigu:
This should be fixed by #1437

PR #1437, opened 2021-03-29
Remove about dialog custom stylesheet

tmontaigu:
So that it respects the global stylesheet.

Related to #1415.

Commit #2ae7b490d9b07cdc2544b1dd117075cc9189f7ed forced the font color
to black, which indeed fixes the probleme of white text on white
background, but I think it is better the let the dialog use the 'global'
stylesheet so that if the theme is dark, the about window also is.

image
image
image

link color doesn't look very readable on dark themes 🤔

### DIFF
--- a/qCC/ui_templates/aboutDlg.ui
+++ b/qCC/ui_templates/aboutDlg.ui
@@ -16,10 +16,6 @@
   <property name="windowTitle">
    <string>About CloudCompare</string>
   </property>
-  <property name="styleSheet">
-   <string notr="true">background-color:white;
-color:black;</string>
-  </property>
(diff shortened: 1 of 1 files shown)


dgirardeau:
Should we change the link color to red? Or something purple?

tmontaigu:
I don't know about chosing a color ourselves as it would have to be color that works for both light & dark, maybe we should let the theme handle the color used for links (if its possible, I haven't checked).

A way to work aroung this  would be to replace the link with an image/button that act as a link when clicked just like the donate button

dgirardeau:
Should we merge this PR or should we wait?

tmontaigu:
And this one can also be merged

PR #1478, opened 2021-06-22
Small fix for PDAL 2.3.0

hobu:
PDAL 2.3.0 changed its LasReader class.


dgirardeau (approved):

nh2:
Could this be put into a new release, so that Linux distros can update to newer PDAL without carrying a patch?

Thanks!

PR #1485, opened 2021-07-09
Segmentation Cancel Fix

WargodHernandez:
Release mouse in-case the escape shortcut was pressed while mouse was captured.

Moved the check outside of the `if (m_somethingHasChanged)` check so the mouse is released regardless of whether or not something has changed


dgirardeau (approved):

dgirardeau:
Thanks ;)

PR #1510, opened 2021-09-09
Rework BUILD.md to be more up to date

tmontaigu:
I feel like the BUILD.md file is a bit old, so I reworked it to be more up to date.


dgirardeau (approved):
Wow, nice!

PR #1516, opened 2021-09-26
Add Draco IO plugin

dgirardeau:
Early version


plugins/core/Standard/CMakeLists.txt
```diff
@@ -20,6 +20,7 @@ set( submod_plugins
 		${CMAKE_CURRENT_SOURCE_DIR}/qMasonry
 		${CMAKE_CURRENT_SOURCE_DIR}/qMPlane
 		${CMAKE_CURRENT_SOURCE_DIR}/qJSonRPCPlugin
+		${CMAKE_CURRENT_SOURCE_DIR}/qUltrabeam
```
tmontaigu, line 23: Is qUltrabeam a part of the qDracoIO plugin ? 🤔

plugins/core/IO/qDracoIO/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
+option( PLUGIN_IO_QDRACO "Install qDracoIO plugin to read & write DRC files" OFF )
+
+if( PLUGIN_IO_QDRACO )
+	project( QDRACO_IO_PLUGIN )
+
+	set(CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/modules)
```
tmontaigu, line 6: It seems the files contained in this directory have not been added to the commit
And `list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/modules)` might be better to not override the `CMAKE_MODULE_PATH`

tmontaigu:
I see there is a [conda-forge package](https://anaconda.org/conda-forge/draco) for draco so we should be able to add the plugin to the CI of Windows and macOs

plugins/core/Standard/CMakeLists.txt
dgirardeau: Good point, this one shouldn't even be here 👍

plugins/core/IO/qDracoIO/CMakeLists.txt
dgirardeau: Ah, that was a (failing) attempt to help CMake find the module...

dgirardeau:
> I see there is a [conda-forge package](https://anaconda.org/conda-forge/draco) for draco so we should be able to add the plugin to the CI of Windows and macOs

I don't know!

PR #1532, opened 2021-10-31
Fix macOs DeployQt.cmake for XCode generator

tmontaigu:
XCode is a multi config generator so we have to take that
into account to point to the correct path where binaries are.


dgirardeau (approved):

dgirardeau:
Merci ;)

PR #1541, opened 2021-11-16
Fix build on FreeBSD

lbartoletti:
.so is the standard for unix-like (except macos), not only linux.

Replace Q_OS_LINUX by Q_OS_UNIX


dgirardeau (approved):

dgirardeau:
Thanks for the contribution

PR #1543, opened 2021-11-19
CMake/macOs: Fix install

tmontaigu:
Installing the CloudCompare / CCViewer executable after DeployQt broke the installation
because it overwrote in the bundle (.app) the executable that had the proper path to load liraries
(set by the macdeployqt program) by the 'base' executable that does not have the proper paths to
load libraries resulting in a broken .app

In the DeployQt for macOs use USE_SOURCE_PERMISSIONS so that the files keep the correct
permissions (like permission to be executable).

### DIFF
--- a/ccViewer/CMakeLists.txt
+++ b/ccViewer/CMakeLists.txt
@@ -82,8 +82,10 @@ if( WIN32 )
 	endif()
 endif()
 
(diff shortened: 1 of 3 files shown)


dgirardeau:
Good catch!

PR #1548, opened 2021-11-26
Split Plugins table in 3 (Gl, IO, STANDARD)

tmontaigu:
And add qDraco to the list of IO plugins

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -90,13 +90,21 @@ The following options are **Windows-only**:
 
 The available plugins are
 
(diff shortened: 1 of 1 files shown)


dgirardeau (approved):
Nice! I totally forgot about this table. The next time I add a plugin, I'll try to think about it...

tmontaigu:
I was happy to have this table to easily see the plugins that I could try to compile for the macOS executable :smiley:

PR #1561, opened 2021-12-19
Rewrite qPdalIo

tmontaigu:
This is a tentative rewrite of the qPdal IO to to make it better and fix some problems we had with it.

- Change the save dialog:
  - Add widgets to select the Las version and point format
  - Add ability for user to select which scalar field is mapped to which las dimension
  - Add dialog to so that users can chose to add some scalar fields as extra scalar fields
- Automatically tries to select the best point format when a saving a cloud that has no Las metadata
- Saves Vlrs by default

It's not yet finished.


tpwrules:
Can this be updated for compatibility with PDAL 2.4.0?

tpwrules:
Also, how does this interact with your prototype LAS plugin?

tmontaigu:
This does not interact with the prototype LAS plugin based on LASzip, this rewrite does the same thing as the prototype but it uses PDAL

tmontaigu:
Closing as #1688 is merged,

I still plan on redoing a PDAL plugin but this time with a different scope

PR #1576, opened 2022-01-19
Doc update

bramburn:
hi Daniel,
As discussed I've started to add some documentation to the code. Hope these are helpful. I am slowly adding them and I will PR.

I've also added cmake-*/ into the gitignore as my cmake kept building it inside the working dir.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -45,6 +45,10 @@ cmake -DCMAKE_PREFIX_PATH=C:\Qt\5.15.2\msvc2019_64 ..
 # macOs
 mkdir build && cd build
 cmake -DCMAKE_PREFIX_PATH=/usr/local/opt/qt@5 ..
+
+# Linux
+mkdir build && cd build
+cmake ..
 ```
 
 You can always take a look at how CloudCompare is being build on [GitHub's CI](.github/workflows/build.yml)


.gitignore
```diff
@@ -136,6 +136,7 @@ CCCoreLibExport.h
 /.vs
 .vscode
 /CMakeSettings.json
+cmake-*
```
dgirardeau, line 139: Is it really necessary?

ccViewer/ccViewerApplication.cpp
```diff
@@ -24,13 +24,22 @@
 #include "ccviewer.h"
 #include "ccViewerApplication.h"
 
-
+/*
+ * Initialise the ccViewerApp application
+ * this is version 1.40 alpha, it implements ccApplicationBase
```
dgirardeau, line 29: I don't think we want (or we will remind) to update this line each time we upgrade the version

ccViewer/ccViewerApplication.cpp
```diff
@@ -24,13 +24,22 @@
...
+ * @param reference to argument
```
dgirardeau, line 30: the format is '\param' then you have to add the name of the argument
`\param argc number of arguments`

ccViewer/ccViewerApplication.cpp
```diff
@@ -24,13 +24,22 @@
...
+ * @param argv pointer's pointer
+ * @param isCommandLine returns true if this is called from command line
```
dgirardeau, line 32: not 'return' but 'indicates if...'?

ccViewer/ccViewerApplication.cpp
```diff
@@ -24,13 +24,22 @@
...
+ * */
 ccViewerApplication::ccViewerApplication( int &argc, char **argv, bool isCommandLine )
 	: ccApplicationBase( argc, argv, isCommandLine, QStringLiteral( "1.40 alpha" ) )
 {
 	setApplicationName( "CloudCompareViewer" );
 }
-
+/*
+ * setViewer functions sets an instance of QMainWindow->ccViewer
```
dgirardeau, line 40: Not need to reuse the function name --> 'Sets an instance of QMainWindow->ccViewer'

And I'm not sure we want to remind that ccViewer is a QMainWindow here?

I would have wrote something like 'Sets the ccViewer instance associated to this application'

ccViewer/ccViewerApplication.h
```diff
@@ -36,6 +36,8 @@ class ccViewerApplication : public ccApplicationBase
 	bool event( QEvent *inEvent ) override;
 	
 private:
+    // ptr to QMainWindow->ccViewer
```
dgirardeau, line 39: Format is //! and once again I wouldn't remind here that ccViewer derives from QMainWindow (I'm not sure to understand the benefit?). 
--> '//! Associated ccViewer instance'?

libs/CCAppCommon/devices/3dConnexion/cc3DMouseManager.cpp
```diff
@@ -330,7 +332,7 @@ void cc3DMouseManager::on3DMouseReleased()
 		win->redraw();
 	}
 }
-
+// sets up the menu for 3D Mouse
```
dgirardeau, line 335: documentation the function in the header file should be sufficient? (moreover most the methods are not documented this way in cpp files, and it's hard to keep the consistency between the header and the cpp file)

libs/CCAppCommon/devices/3dConnexion/cc3DMouseManager.cpp
```diff
@@ -339,6 +341,7 @@ void cc3DMouseManager::setupMenu()
 	m_actionEnable = new QAction( tr( "Enable" ), this );
 	m_actionEnable->setCheckable( true );
 
+    // see connect menu
```
dgirardeau, line 344: not sure to see the benefit (or understand) this comment?

libs/CCAppCommon/include/ccApplicationBase.h
```diff
@@ -24,15 +24,26 @@
 //! Mimic Qt's qApp for easy access to the application instance
 #define ccApp (static_cast<ccApplicationBase *>( QCoreApplication::instance() ))
 
+
+/*
+ * main ccApplicationBase that the CCViewer which implements
```
dgirardeau, line 29: So I think that's a general comment: I dont' see the point of describing the inheritance (which can be very easily visualised in doxygen) and copying / referencing Qt's documentation explicitly (and pointing to version 5, it will be hard to maintain in the future)

libs/CCAppCommon/include/ccApplicationBase.h
```diff
@@ -24,15 +24,26 @@
...
+ * QApplication: The QApplication class manages the GUI application's control flow and main settings
+ * https://doc.qt.io/qt-5/qapplication.html
+ * QApplication inherites QGuiApplication, which forces this app to use ::event(QEvent e)
```
dgirardeau, line 32: inherites --> inherits? (but anyway, I don't think we need to document Qt ;) )

libs/CCAppCommon/devices/3dConnexion/cc3DMouseManager.cpp
```diff
@@ -90,9 +90,11 @@ void cc3DMouseManager::enableDevice(bool state, bool silent)
 
 void cc3DMouseManager::releaseDevice()
 {
+    // nothing to release
```
dgirardeau, line 93: general comment: for historical reasons, we use tabulations (4 spaces) and not spaces in this project. This is why all your comments appear shifted in github.

libs/CCAppCommon/include/ccApplicationBase.h
```diff
@@ -24,15 +24,26 @@
...
+ *
+ * */
 class CCAPPCOMMON_LIB_API ccApplicationBase : public QApplication
 {
 public:
 	//! This must be called before instantiating the application class so it
 	//! can setup OpenGL first.
+    //! this is not from the QApplication parent class
```
dgirardeau, line 40: that's obvious otherwise we would use the 'override' keyword
Plus, additional comments should be in /** **/ sections (//! is for the brief description)

libs/CCAppCommon/include/ccApplicationBase.h
```diff
@@ -24,15 +24,26 @@
...
 	static void	initOpenGL();
-	
+
+    // default constructor
```
dgirardeau, line 43: //!

libs/CCAppCommon/include/ccApplicationBase.h
```diff
@@ -24,15 +24,26 @@
...
 	ccApplicationBase( int &argc, char **argv, bool isCommandLine, const QString &version );
-	
+
+    // returns if this is commandline or not
```
dgirardeau, line 46: //!

libs/CCAppCommon/include/ccApplicationBase.h
```diff
@@ -44,9 +55,12 @@ class CCAPPCOMMON_LIB_API ccApplicationBase : public QApplication
 	void setupPaths();
 		
 	const QString c_VersionStr;
-	
+	// QString
```
dgirardeau, line 58: Well, once again, we don't need to restate the obvious, I would expect more something like 'Shader path' than 'QString' (the return type as well as the argument types clearly appear in the doxygen output).

libs/CCAppCommon/include/ccApplicationBase.h
```diff
@@ -44,9 +55,12 @@ class CCAPPCOMMON_LIB_API ccApplicationBase : public QApplication
...
+    // url: https://doc.qt.io/qt-5/qstring.html
 	QString	m_ShaderPath;
 	QString	m_TranslationPath;
```
dgirardeau, line 61: missing doc?

libs/CCAppCommon/include/ccApplicationBase.h
```diff
@@ -44,9 +55,12 @@ class CCAPPCOMMON_LIB_API ccApplicationBase : public QApplication
...
+    // QStringList of plugins
```
dgirardeau, line 62: Full path to each plugin?

dgirardeau (changes requested):
Ok, I think I'll have to stop my review here. That would be a very nice contribution but I fear that in the current state it needs a lot of rework and change of 'philosophy'. 

I won't blame you if you think it's too much or work by the way 😅

bramburn:
Sorry I will work on it and fix it. I'm new to working on c++ projects with others and especially QT.

bramburn:
thank you for taking a lot of time responding and reviewing.

libs/CCAppCommon/include/ccApplicationBase.h
tmontaigu: rewriting the type and putting a link to Qt's doc is not a really helpful comment.

QString is such an important part of Qt that people wanting to know more about QString will easily find information on the internet

bramburn: ok I will remove that.

bramburn:
I'll be working on it differently, one file at a time.

Issue #1560, opened 2021-12-17
libCCAppCommon.so cannot open shared object file


reppiZ:
Just installed CloudCompare following the directions in BUILD.md. Even built the binary out of source like suggested and used the 
"--recursive" argument when cloning the repository. It installed successfully and I verified that the shared library objects are in /usr/local/lib however, when I run the command "CloudCompare" i get the following: "error while loading shared libraries: libCCAppCommon.so: cannot open shared object file: No such file or directory"

However, "libCCAppCommon.so" is indeed in /urs/local/lib along with other needed shared libraries. 

Am i missing something? Also please forgive me if this is not in the right spot. I'm not use to GitHub.

Environment: 
 - Ubuntu 21.10 x64
 - Qt Version 5.15.2

reppiZ:
Fixed with running "sudo ldconfig"

EpicWink:
Related: #1294 , #1346

PR #1581, opened 2022-01-22
Linux: Fix shared lib install folder

tmontaigu:
On Linux the `INSTALL_RPATH` was set to
`"${CMAKE_INSTALL_FULL_LIBDIR}/cloudcompare`.

However the path where the shared libs were installed
was set to `"${destination}/${CMAKE_INSTALL_LIBDIR}/` or
`"${CMAKE_INSTALL_LIBDIR}"` which is not the same path as
the `INSTALL_RPATH`.

To illustrate, with a `CMAKE_INSTALL_PREFIX` set to
`~/CloudCompare/install` the shared libs were installed
in `~/CloudCompare/install/lib64` but the RPATH set in
each lib was pointing to `~/CloudCompare/install/lib64/cloudcompare`

Resulting in an error like this when running the installed CloudCompare:
```
error while loading shared libraries: libCCAppCommond.so: cannot open shared object file: No such file or directory
```

This fixes the problem by making the two things point to the same
path `"${CMAKE_INSTALL_FULL_LIBDIR}/cloudcompare"`.
So in the example libs properly end up in
`~/CloudCompare/install/lib64/cloudcompare`.

Although the cmake for installing shared target is still a bit ?strange?.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,12 +37,6 @@ endif()
 # Default debug suffix for libraries.
 set( CMAKE_DEBUG_POSTFIX "d" )
 
-# RPATH Linux/Unix: (dynamic) libs are put in a subdir of prefix/lib,
-# since they are only used by qCC/ccViewer
-if( UNIX AND NOT APPLE )
-	include( GNUInstallDirs )
-	set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_FULL_LIBDIR}/cloudcompare")
-endif()
 
 # Define target folders
 # (now that ccViewer can have its own plugins, qCC and ccViewer must fall in separate folders!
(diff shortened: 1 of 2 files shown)


tmontaigu:
This should fix issues like this  #1560

dgirardeau:
I believe these CMake scripts were heavily updated/upgraded by @asmaloney but he was not able to test them on Linux (only on macOS). Therefore, if you find other strange things, feel free to update/fix them.

asmaloney:
Those changes to the Linux stuff were added after my changes.

https://github.com/CloudCompare/CloudCompare/commits/35b21d28c5dc2b3aa98497f35a58945c2c3328f4/CMakeLists.txt

https://github.com/CloudCompare/CloudCompare/commits/35b21d28c5dc2b3aa98497f35a58945c2c3328f4/cmake/Install.cmake

So it might make sense to back up, look at what I had, and fix that instead of adding hacks on top of hacks.

PR #1589, opened 2022-01-27
Allow the possibility to define multiple plugin paths from CC_PLUGIN_PATH environment variable

dimitri-tdg:
A very tiny modification to allow definition of multiple plugin paths from environment variable. It can be useful when debugging or compiling a plugin (for example CC_PLUGIN_PATH=<build dir>/plugins/core/IO/qCoreIO:<build dir>/plugins/core/IO/qE57IO)

### DIFF
--- a/libs/CCAppCommon/src/ccApplicationBase.cpp
+++ b/libs/CCAppCommon/src/ccApplicationBase.cpp
@@ -259,6 +259,6 @@ void ccApplicationBase::setupPaths()
 	// included and appdata ones.
 	if ( env.contains("CC_PLUGIN_PATH") )
 	{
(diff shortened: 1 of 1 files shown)


dgirardeau (approved):

tmontaigu:
On windows `;` is more common to split variable values, so we should use `;` on windows and `:` on other OSes

tmontaigu:
Like `[...].split(QDir::listSeparator())` https://doc.qt.io/qt-5/qdir.html#listSeparator

dimitri-tdg:
> On windows `;` is more common to split variable values, so we should use `;` on windows and `:` on other OSes

You're right! I'll make the modifications

tmontaigu (approved):

PR #1596, opened 2022-02-02
ccApplicationBase.h

bramburn:
hi guys,
I've only added a few comments this time around.
I was not sure if you wanted every param in the class documented, as the previous PR comments mentioned it was useless to describe those.

I've also updated the cpp comments to match QT style comments.

My plan is to only tackle class documentation, constructors and any methods accepting arguments. Is that ok?

### DIFF
--- a/libs/CCAppCommon/include/ccApplicationBase.h
+++ b/libs/CCAppCommon/include/ccApplicationBase.h
@@ -27,6 +27,7 @@
 /*!
  * ccApplicationBase provides a QApplication class with additional Constructor param isCommandLine and version
  * The class also initialises openGL before instantiating the application class
+ * \see QApplication
  */
 class CCAPPCOMMON_LIB_API ccApplicationBase : public QApplication
 {
@@ -37,10 +38,10 @@ public:
 
     //! ccApplicationBase Constructor
 	/*!
-	 * @param argc number of arguments
-	 * @param argv array of arguments
-	 * @param isCommandLine boolean to define the application as either a CL or GUI. If true its a commandline application
-	 * @param version string of the application version set by CloudCompare viewer or app
+	 * @param argc argc must be greater than zero, this is set by QT
+	 * @param argv this contains the command line values, see QApplication, again this is set by QT
+	 * @param isCommandLine boolean to define the application as either a Commandline or GUI. If true its a commandline application
+	 * @param version string of the application version set by CloudCompare viewer or app.
 	 */
 	ccApplicationBase( int &argc, char **argv, bool isCommandLine, const QString &version );
 
--- a/libs/CCAppCommon/include/ccApplicationBase.h
+++ b/libs/CCAppCommon/include/ccApplicationBase.h
@@ -23,16 +23,32 @@
 
 //! Mimic Qt's qApp for easy access to the application instance
 #define ccApp (static_cast<ccApplicationBase *>( QCoreApplication::instance() ))
-
+//! ccApplicationBase
+/*!
+ * ccApplicationBase provides a QApplication class with additional Constructor param isCommandLine and version
+ * The class also initialises openGL before instantiating the application class
+ */
 class CCAPPCOMMON_LIB_API ccApplicationBase : public QApplication
 {
 public:
 	//! This must be called before instantiating the application class so it
 	//! can setup OpenGL first.
 	static void	initOpenGL();
-	
+
+    //! ccApplicationBase Constructor
+	/*!
+	 * @param argc number of arguments
+	 * @param argv array of arguments
+	 * @param isCommandLine boolean to define the application as either a CL or GUI. If true its a commandline application
+	 * @param version string of the application version set by CloudCompare viewer or app
+	 */
 	ccApplicationBase( int &argc, char **argv, bool isCommandLine, const QString &version );
-	
+
+    //! isCommandLine
+    /*!
+     * Method that confirms if it is running via commandline or not
+     * @return bool
+     */
 	bool isCommandLine() const { return c_CommandLine; }
 	
 	QString versionStr() const;
--- a/libs/CCAppCommon/src/ccApplicationBase.cpp
+++ b/libs/CCAppCommon/src/ccApplicationBase.cpp
@@ -48,8 +48,8 @@
 
 void ccApplicationBase::initOpenGL()
 {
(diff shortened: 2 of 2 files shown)


libs/CCAppCommon/include/ccApplicationBase.h
```diff
@@ -23,16 +23,33 @@
 
 //! Mimic Qt's qApp for easy access to the application instance
 #define ccApp (static_cast<ccApplicationBase *>( QCoreApplication::instance() ))
-
+//! ccApplicationBase
```
dgirardeau, line 26: For me, here we should add a description of what the class does / what's its purpose. We don't need to recall the name of the class (that will appear in the doxygen doc anyway)

libs/CCAppCommon/include/ccApplicationBase.h
```diff
@@ -23,16 +23,33 @@
...
+/*!
+ * ccApplicationBase provides a QApplication class with additional Constructor param isCommandLine and version
```
dgirardeau, line 28: Looks a little bit too detailed to me. For this kind of class without much intelligence, I don't think we always need a detailed description.

libs/CCAppCommon/include/ccApplicationBase.h
```diff
@@ -23,16 +23,33 @@
...
 {
 public:
 	//! This must be called before instantiating the application class so it
 	//! can setup OpenGL first.
 	static void	initOpenGL();
-	
+
+    //! ccApplicationBase Constructor
+	/*!
+	 * @param argc argc must be greater than zero, this is set by QT
```
dgirardeau, line 41: The description is simply: 'number of command line arguments' (and it could be 0, we don't really care?). This will be transferred **to** Qt by the caller of this constructor (it comes from the system).

libs/CCAppCommon/include/ccApplicationBase.h
```diff
@@ -23,16 +23,33 @@
...
+	 * @param argv this contains the command line values, see QApplication, again this is set by QT
```
dgirardeau, line 42: once again, this will be transferred to Qt (QApplication).

libs/CCAppCommon/include/ccApplicationBase.h
```diff
@@ -23,16 +23,33 @@
...
+	 * @param isCommandLine boolean to define the application as either a Commandline or GUI. If true its a commandline application
```
dgirardeau, line 43: Could we condense the documentation a little? Generally it's not useful to recall the type of the variable or class (already visible in the doxygen output). Something like:
'Whether the application is to be used in command line mode (true) or GUI mode (false).'

libs/CCAppCommon/include/ccApplicationBase.h
```diff
@@ -23,16 +23,33 @@
...
+	 * @param version string of the application version set by CloudCompare viewer or app.
```
dgirardeau, line 44: 'Application version' would be sufficient

libs/CCAppCommon/include/ccApplicationBase.h
```diff
@@ -23,16 +23,33 @@
...
+	 */
 	ccApplicationBase( int &argc, char **argv, bool isCommandLine, const QString &version );
-	
+
+    //! isCommandLine
```
dgirardeau, line 48: Simply
`//! Returns whether this application is in command line mode or not`

(should be sufficient, getters might not need a @return field since it's generally redundant with the short description)

libs/CCAppCommon/include/ccApplicationBase.h
```diff
@@ -23,16 +23,33 @@
...
+    /*!
+     * Method that confirms if it is running via commandline or not
+     * @return bool
+     */
 	bool isCommandLine() const { return c_CommandLine; }
 	
 	QString versionStr() const;
```
dgirardeau, line 55: `//! Returns the application version`

libs/CCAppCommon/src/ccApplicationBase.cpp
```diff
@@ -48,8 +48,8 @@
 
 void ccApplicationBase::initOpenGL()
 {
-	//See http://doc.qt.io/qt-5/qopenglwidget.html#opengl-function-calls-headers-and-qopenglfunctions
-	/** Calling QSurfaceFormat::setDefaultFormat() before constructing the QApplication instance is mandatory
+	//! See http://doc.qt.io/qt-5/qopenglwidget.html#opengl-function-calls-headers-and-qopenglfunctions
```
dgirardeau, line 51: You shouldn't add //! in a cpp file (that's only for short descriptions of methods or members)

Here the intent was to add a comment block (`/* ... */ `would have been sufficient though)

libs/CCAppCommon/src/ccApplicationBase.cpp
```diff
@@ -76,8 +76,8 @@ void ccApplicationBase::initOpenGL()
 		QSurfaceFormat::setDefaultFormat(format);
 	}
 
-	// The 'AA_ShareOpenGLContexts' attribute must be defined BEFORE the creation of the Q(Gui)Application
-	// DGM: this is mandatory to enable exclusive full screen for ccGLWidget (at least on Windows)
+	//! The 'AA_ShareOpenGLContexts' attribute must be defined BEFORE the creation of the Q(Gui)Application
+	//! DGM: this is mandatory to enable exclusive full screen for ccGLWidget (at least on Windows)
```
dgirardeau, line 80: Same, we don't use //! in cpp files

dgirardeau: + same below

bramburn: ok thanks I'll fix these. I'll take note of the documentation styles.

libs/CCAppCommon/include/ccApplicationBase.h
bramburn: ok I thought it would from following Doxygen's guide on classes. It was stating to put //! <class name> then /*!....*/ with the brief.

libs/CCAppCommon/include/ccApplicationBase.h
bramburn: ok, no problem was re-phrased from the QT application ctor.

libs/CCAppCommon/include/ccApplicationBase.h
bramburn: ok will change. thank you

Issue #1600, opened 2022-02-10
Unintentionally change spinbox's value in "Properties" scroll area.


hackery98:
When I scroll "Properties" widget, unintentionally change spinbox's and comboBox's value in "Properties" scroll area.

Solutions:

**First,**  install EventFilter 'FilterMouseWheelNoFocusEvent' for spinbox and comboBox to avoid such mistake.

FilterMouseWheelNoFocusEvent:  ignore MouseWheel event when the widget has no current focus.

**Second,** add m_lastFocusItemRole to  record which widget's value changed.  
Then using it to setFocus the widget in ccPropertiesTreeDelegate::updateModel(). 
 so you can continuous scroll the spinbox's value as before.

here is the main code: is that ok?
![图片](https://user-images.githubusercontent.com/6083117/153350761-84adbb59-c113-4ef3-956d-964a9c0c9be9.png)

![图片](https://user-images.githubusercontent.com/6083117/153350715-41590002-fa46-422a-87b3-7bab671a1775.png)

![图片](https://user-images.githubusercontent.com/6083117/153350859-22766cc4-edb3-403e-a3fd-4a423fb43cb6.png)

tmontaigu:
Closing because #1602 was merged

PR #1602, opened 2022-02-10
Avoid unintentionally change spinbox's and comboBox's value in "Prope…

hackery98:
#1600

### DIFF
--- a/qCC/db_tree/ccPropertiesTreeDelegate.cpp
+++ b/qCC/db_tree/ccPropertiesTreeDelegate.cpp
@@ -70,9 +70,30 @@
 #include <QSpinBox>
 #include <QStandardItemModel>
 #include <QToolButton>
+#include <QEvent>
 
 //System
 #include <cassert>
+
+class FilterMouseWheelNoFocusEvent : public QObject
+{
+public:
+	explicit FilterMouseWheelNoFocusEvent(QObject *parent)
+	{
+	}
+protected:
+	bool eventFilter(QObject* o, QEvent* e) override
+	{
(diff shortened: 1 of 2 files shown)


qCC/db_tree/ccPropertiesTreeDelegate.cpp
```diff
@@ -2341,6 +2368,7 @@ void ccPropertiesTreeDelegate::octreeDisplayedLevelChanged(int val)
 		octree->setDisplayedLevel(val);
 		updateDisplay();
 		//we must also reset the properties display!
+		m_lastFocusItemRole = OBJECT_OCTREE_LEVEL;
```
dgirardeau, line 2371: I guess these lines shouldn't be below the comment which applies to the call to 'updateModel()'. Can you move them after updateModel() or before the comment?

hackery98: > 		//record item role here to use it 'setFocus' in 'createEditor', 
> 		//than you can continuous scroll the widget's value.
> 		m_lastFocusItemRole = OBJECT_CONE_HEIGHT;
> 
> 		//we must also reset the properties display!
> 		updateModel();
is this ok? Please let me know if there is a problem with the expression .

dgirardeau: Thanks for asking. I would simply write something like:
//record item role to force the scroll focus (see 'createEditor').

dgirardeau (approved):

dgirardeau:
Thanks for the contribution

PR #1613, opened 2022-03-18
Fix: "ccGLWindow" not responding to resizeEvent.

hackery98:
When tapping between "title" and "Cascade" in "3D view" menu, the views not show right.

图片

### DIFF
--- a/libs/qCC_glWindow/src/ccGLWindow.cpp
+++ b/libs/qCC_glWindow/src/ccGLWindow.cpp
@@ -1169,6 +1169,7 @@ bool ccGLWindow::event(QEvent* evt)
 	{
 		update();
 	}
+	break;
 #endif
 
 	default:
--- a/libs/qCC_glWindow/src/ccGLWindow.cpp
+++ b/libs/qCC_glWindow/src/ccGLWindow.cpp
@@ -1137,11 +1137,6 @@ bool ccGLWindow::event(QEvent* evt)
 	}
 	break;
 
-	case QEvent::Resize:
-	{
-		update();
-	}
-
 #ifdef CC_GL_WINDOW_USE_QWINDOW
 	case QEvent::Resize:
 	{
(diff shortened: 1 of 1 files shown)


libs/qCC_glWindow/src/ccGLWindow.cpp
```diff
@@ -1137,6 +1137,11 @@ bool ccGLWindow::event(QEvent* evt)
 	}
 	break;
 
+	case QEvent::Resize:
+	{
+		update();
+	}
+
```
dgirardeau, line 1144: You would at least have to add a `return true;` statement. But mind that just after these new lines, you have a `#ifdef CC_GL_WINDOW_USE_QWINDOW` block, in which this event is already redefined. So you will need to move your own version in a #else block after the one below. i.e. line 1173.

dgirardeau: And I would be interested to know about your setup? I guess it's the non-stereo version? And with which version of Qt are you compiling?

hackery98: My Qt is 5.12.12.  
I tested both non-stereo and stereo mode, all has this issue.
It seems the event should pass to parent. Because after add `return true`, it looks more weird, can't show any elemnets.
Add I will move the code to `#else `block .

hackery98: 1ef92605780f2c9d6d728a15e88a54fb58bc62e8

libs/qCC_glWindow/src/ccGLWindow.cpp
```diff
@@ -1169,7 +1164,11 @@ bool ccGLWindow::event(QEvent* evt)
 		evt->accept();
 	}
 	return true;
-
+#else
+	case QEvent::Resize:
+	{
+		update();
+	}
```
dgirardeau, line 1171: you'll still need to add a 'break' statement.

hackery98: My bad, will add it right away.

Issue #1258, opened 2020-05-30
Cannot compile with -DCCCORELIB_USE_CGAL=ON


PE1NUT:
CC does not compile with CGAL enabled. I'm trying to run Delaunay2.5 as I have a point cloud that I would like to turn into a mesh. However, compilation fails on a seemingly trivial constexpr for RAD_TO_DEG and DEG_TO_RAD in libs/qCC_db/extern/CCCoreLib/include/CCConst.h

**How to reproduce**
Platofm: Ubuntu 20.20
git clone --recursive https://github.com/cloudcompare/CloudCompare.git
cd CloudCompare; mkdir buid; cd build
cmake -DCCCORELIB_USE_CGAL=ON -DCMAKE_BUILD_TYPE=Release ..
make

Error message:
[  1%] Building CXX object libs/qCC_db/extern/CCCoreLib/CMakeFiles/CCCoreLib.dir/src/AutoSegmentationTools.cpp.o
In file included from /home/paul/CloudCompare/libs/qCC_db/extern/CCCoreLib/include/FastMarching.h:7,
                 from /home/paul/CloudCompare/libs/qCC_db/extern/CCCoreLib/include/FastMarchingForPropagation.h:8,
                 from /home/paul/CloudCompare/libs/qCC_db/extern/CCCoreLib/src/AutoSegmentationTools.cpp:7:
/home/paul/CloudCompare/libs/qCC_db/extern/CCCoreLib/include/CCConst.h:29:38: error: ‘(1.8e+2 / 3.1415926535897931e+0)’ is not a constant expression
   29 |  constexpr double RAD_TO_DEG = (180.0/M_PI);
      |                                ~~~~~~^~~~~~
/home/paul/CloudCompare/libs/qCC_db/extern/CCCoreLib/include/CCConst.h:33:37: error: ‘(3.1415926535897931e+0 / 1.8e+2)’ is not a constant expression
   33 |  constexpr double DEG_TO_RAD = (M_PI/180.0);

**Additional context**
The reason I'm building it myself is because on Ubuntu 20.20 (using the Debian package, not the snap), the Delaunay2.5 doesn't work. but returns '[ccMesh::Triangulate] Failed to construct Delaunay mesh (Triangle lib error: CGAL library not supported)'

**Your environment**
 - CC Version (or git hash/tag): f2cdc5eeab3c2923b690bab8c0d9a093faeead67
 - OS & Version: Ubuntu 20.20 LTS
[...]

PR #1623, opened 2022-04-02
Remove uses of floating point math in constexprs

tpwrules:
Floating point math in C++ constexprs is [ill-defined in general](https://stackoverflow.com/questions/50959021/why-compile-time-floating-point-calculations-might-not-have-the-same-results-as) and [causes problems compiling with CGAL](https://github.com/CloudCompare/CloudCompare/issues/1258).

This PR removes all uses of it from the CloudCompare codebase, as there are only a few and the edits required are straightforward.


dgirardeau:
Thanks!

PR #1641, opened 2022-05-17
Add switch case 0 to ICP rotation box

### LINKED ISSUE #1640: ICP crashing when setting to allow XYZ using the interface
**Describe the bug**
CloudCompare is crashing when using ICP method with default option allowing all rotations optimization, error:
```bash
CloudCompare: /home/user/dev/src/tools/cloudcompare/qCC/ccRegistrationDlg.cpp:269: int ccRegistrationDlg::getTransformationFilters() const: Assertion `false' failed.
Aborted (core dumped)
```

**How to reproduce**
-> Select two clouds and run ICP with default options( check that Research->Rotation is at `XYZ`)
-> A Abort will happen

**Expected behaviour**
Do ICP with all degrees of freedom

**Your environment**
 - CC hash(6d4bd0af3182f6f402c733cd297e52d25df2368e - latest as I am writing this issue)
 - ubuntu 20.04
 - nvidia 1060 TI
 - Qt 5.12.8

I think that this is small bug added at commit a0ee03d973d2e7ebf8445c39cd9e35e4b459bdc7, when default was set to `assert(false)` and there is no switch to case `0` that is the default `SkipNone` at the interface. Am I missing something?

ivan1993br:
## Related issue
This PR close #1640 , that causes a `segfault` running ICP without skipping any rotation.

## Analysis
I added a `case 0` to explicit allow this values instead of modifying `default`, that I think is more coherent to not hide any other errors that my happen and can go to `default` 

As it is a really small fix and I needed to fix here to use, I already opened this PR, see what you guys think.

Thanks!

### DIFF
--- a/qCC/ccRegistrationDlg.cpp
+++ b/qCC/ccRegistrationDlg.cpp
@@ -253,6 +253,8 @@ int ccRegistrationDlg::getTransformationFilters() const
 	int filters = CCCoreLib::RegistrationTools::SKIP_NONE;
 	switch (rotComboBox->currentIndex())
 	{
+	case 0:
+		break;
 	case 1:
 		filters |= CCCoreLib::RegistrationTools::SKIP_RYZ;
 		break;


dgirardeau (approved):

dgirardeau:
Ah, good catch. Thanks for the fix.

PR #1649, opened 2022-05-24
Rasterize: scalar field statistics

dgirardeau:
Adding progress bar for rasterized grid export
Add include assert.h to ccAsprsModel.cpp to avoid compile error
Start working on progress bar for raster grid convertToCloud
Remove export "projected" SF value in rasterize, it is sort of redundant with interpolate SF function
Indentation fix
Fix export of std.dev unique and percentile value for SF in rasterize tool
Fix bug cause by renaming eight grid layer
Populating most scalar field statistics.
Create Scalar fields for SF statistics in ccRasterGrid::convertToCloud, still not populating them
Relay list of statistics parameters to be exported for each scalar field, to ccRasterGrid::convertToCloud. Need to do something about it here
Modify SF interpolation in ccRasterGrid::fillWith, from min/max/median of SF value, to SF value corespoinding to min/max/median of height value
Add UI elements for export SF statistics, load and save settings between use
Rename ccRasterGrid.h ExportableFields fields, replace HEIGHT with VALUE
Add UI elements for export SF statistics in rasterize dialog
Handle inverse_var_value case in SF resampling
Rasterize Tool: When in inverse variance mode, always export std.dev SF with same name with updated std.dev
Added model std deviation as exportable SF
Added rasterization with inverse variance weighting, some GUI cosmetics remains


dgirardeau:
@magnuan When looking at the UI, and at the code, I'm a bit confused by this new 'PROJ_INVERSE_VAR_VALUE' option...

It seems that with this mode, while it is selected as a way to compute 'cell height', it's in fact still using the 'average' point (= the closest to the cell center) and it  only used to update the other scalar fields (which is not clearly indicated anywhere).

I also have the impression that we perform a lot of computations anytime the 'Update' button is pressed, while it's not necessary. It would be better to move everything to the lower part (with clearer explanations of what is done) and only do the computations if the cloud (or a raster?) is exported.

@AccessArctic I have uploaded the version I compiled if you want to test this new branch (in its current state). It may take a while before it's actually merged in CloudCompare. Mind that it's a zip package, you might have to install the Visual Studio 2013-2017 redistributable package, and you will also need to convert the files to a limited number of formats (bin, ply, etc.). You should typically use the official version of CC to convert to .bin first.
[CloudCompare_rasterize_SF_resampling.zip](https://github.com/CloudCompare/CloudCompare/files/8765058/CloudCompare_rasterize_SF_resampling.zip)

Shivaspace4:
@dgirardeau 

I have downloaded and just extracted the zip file..cc starts and the rasterize tool that I am interested to check starts with all the new nice options..but exporting and saving the desired raster is not working...the option,  export tab is not even active..

surely  I am  doing something wrong since I have just unzipped and run the program..

It would be really nice to bring out a beta version with the new rasterize tool..since it is a bit difficult , at least for me,  to follow your instructions for compiling the software

Thanks again and also @magnuan , for this very useful improvement!

dgirardeau:
@Shivaspace4 I checked and there were indeed a lot of strange things in the code. I fixed them (hopefully) and tried to make things simpler and more efficient (by regrouping all the options, with 2 checkboxes to choose when exporting a cloud or a mesh whether to compute all the stats on the height values only, or on the scalar fields as well).

You can get the new version here: https://www.cloudcompare.org/release/CloudCompare_rasterize_SF_resampling_v2.7z

dgirardeau:
Okay I've fixed an improved many things on this PR (and more generally with the Rasterize tools, as some things were broken). Hopefully this should now be working properly. I'll merge that soon in the master branch.

magnuan:
Thanks a lot for merging this in.
Sorry I haven't been able to follow it more up.
CloudCompare helps me a whole lot analyzing sonar data, and I hope i can contribute more in the future.

Shivaspace4:
hello and happy new year!

I think that some functions of the rasterize tool are still not working

There is no way to export raster files with percentiles , unique values etc..all those functions that appear at the lowest panel of the tool s GUI

Maybe some kind of code restructuring is necessary..so in the first panel  the user can select what to project (active layer)

in the second panel  after the direction option, the cell height should be changed to the active layer defined in the previous panel and the methods, like minimum, maximum, median, average, inverse var, should be enriched with the percentile methods, unique values, those option that are basically located on the lowest panel

only then should be possible to procced with updating the grid..

it is a bit confusing, but basically all the options that appear on the lowest panel, should be available at the second panel as projection methods

I hope this helps a bit

rasterize_tool

dgirardeau:
About the raster export: yes, this is a current limitation of the feature, and the horizontal bar (just below the red line ;) is here to indicate that the 'export statistics' options currently only for 'Cloud' and 'Mesh' export. Of course this can be changed in the future.

Otherwise I agree that things could be reorganized in order to clarify what is really computed and what is not.

Shivaspace4:
Hello,

I have just installed the current version of CC, and the univariate statistics of the rasterize tool works nice giving as a result a new point cloud (cloud export) which can be again rasterized and exported as a tif raster set!!

thanks a lot to all contributors,
I believe this is a very useful improvement-feature and it is fast :)

PR #1678, opened 2022-08-23
qPDALIO bug resolved

p-leroy:
The minPointFormat value is not changed by GetLASFields::GetLASFields. In my opinion, an ampersand is missing.
This can lead to missing standard fields (GpsTime for instance) when exporting in las/laz. In my case, GpsTime was saved as an extra field instead of being saved directly in the point record.


dgirardeau:
Good catch!

It looks like the spacing in front of the method declaration changed?

p-leroy:
It seems like QtCreator has changed a tab into several spaces

dgirardeau:
Can you fix it? :p

p-leroy:
Yes I will!

dgirardeau:
Thanks!

PR #1687, opened 2022-09-24
reorder actions done at startup

### LINKED ISSUE #1665: Unhandled file extension ‘e57‘
**Describe the bug**

**How to reproduce**

1. Double click on a .e57 file
2. Receive this error message
grafik
3. The file will not open.

**Expected behaviour**
The file should open properly.

**Additional context**

**Your environment**
 - CC Version (or git hash/tag): 2.13 alpha
 - OS & Version: MacOS 64bit 
 - Graphics card: M1
 - Qt Version (if compiling): n/a

tmontaigu:
This reorders the order at which we do things at startup.

- We init and load plugins just after the splash screen is shown
- Don't call QApplication::ProcessEvents as its preferable to wait that we are more ready (plugins are loaded and other things) before processing events.
- Remove the splash screen timer

Fixes #1665

### DIFF
--- a/qCC/main.cpp
+++ b/qCC/main.cpp
@@ -56,6 +56,30 @@
 #endif
 
 
+bool isCommandLine(int argc, char **argv)
+{
+#ifdef Q_OS_MAC
+    // On macOS, when double-clicking the application, the Finder (sometimes!) adds a command-line parameter
+    // like "-psn_0_582385" which is a "process serial number".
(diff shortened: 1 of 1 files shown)


dgirardeau:
Interesting. Instead of making comments, I directly updated a few minor things (syntax mostly)

tmontaigu:
👍 

Btw, I've sent a message on slack about LAS plugin

PR #1688, opened 2022-09-24
Add new LAS plugin based on LASZip

tmontaigu:
Meant to replace QLASFWFIO that is based on LASlib as LASZip is available on most distro
Should replace replace the PDAL plugin as the main LAS filter 
The PDAL plugin should become a more general plugin converting from cc to pdal

It needs some real world testing especially for global shift, FWF


dgirardeau:
Ok, once again I made a quick syntax cleaning commit, to avoid having to flood you with comments. I'll add some comments still.

plugins/core/IO/qLASIO/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
+option(PLUGIN_IO_QLAS "LAS IO plugin using laszip" OFF)
+
+if (PLUGIN_IO_QLAS)
```
dgirardeau, line 3: I renamed the plugin so that it's consistent with the others

plugins/core/IO/qLASIO/include/LasDetails.h
```diff
@@ -0,0 +1,356 @@
...
+class ccPointCloud;
+class ccScalarField;
+
+class QDataStream;
+
+struct laszip_header;
+struct laszip_vlr;
+typedef laszip_vlr laszip_vlr_struct;
+
+#define LAS_VLR_HEADER_SIZE 54
```
dgirardeau, line 37: could we switch to constexpr int or doubles?

plugins/core/IO/qLASIO/include/LasDetails.h
```diff
@@ -0,0 +1,356 @@
...
+
+        template <class T> static constexpr Range ForType()
+        {
+            return Range(std::numeric_limits<T>::min(), std::numeric_limits<T>::max());
+        }
+
+        static Range ForBitSize(uint8_t numBits)
+        {
+            Range range(0.0, 0.0);
+            range.max = static_cast<ScalarType>(std::pow(2, numBits) - 1.0);
```
dgirardeau, line 117: '(1 << numBits) - 1' maybe?

plugins/core/IO/qLASIO/include/LasDetails.h
```diff
@@ -0,0 +1,356 @@
...
+
+  public: // Static functions
+    static constexpr const char *NameFromId(LasScalarField::Id id);
+    static LasScalarField::Id IdFromName(const char *name, unsigned int targetPointFormat);
+    static LasScalarField::Range ValueRange(LasScalarField::Id id);
+
+    // TODO They should be private
+  public: // Members
+    /// The Id of the LAS field this relates to.
+    Id id;
```
dgirardeau, line 138: default initialize to something? Or to an 'invalid' enum?

plugins/core/IO/qLASIO/include/LasDetails.h
```diff
@@ -0,0 +1,356 @@
...
+    ccScalarField *sf{nullptr};
+    Range range;
+};
+
+/// Returns point the formats available for the given version.
+///
+/// If the version does not exists or is not supported a nullptr is returned.
+///
+/// \param version version string, must be "major.minor" e.g. "1.2"
+const std::vector<unsigned int> *PointFormatsAvailableForVersion(const char *version);
```
dgirardeau, line 155: Ok, this is a general comment: I am really not a big fan of all these global functions. They should at least all leave in a namespace, or even better as static members of a common struct or class. It kills a bit the 'object oriented' aspect of C++, and you have to make sure they have unique names, etc.

plugins/core/IO/qLASIO/include/LasDetails.h
```diff
@@ -0,0 +1,356 @@
...
+
+/// Returns the scalar fields that correspond the the pointFormatId
+/// as per the LAS specification.
+std::vector<LasScalarField> LasScalarFieldForPointFormat(unsigned int pointFormatId);
+
+/// Array containing the available versions
+extern const char *AvailableVersions[3];
```
dgirardeau, line 162: Same, 'extern' shouldn't be used

plugins/core/IO/qLASIO/include/LasIOFilter.h
```diff
@@ -0,0 +1,34 @@
...
+
+#include "FileIOFilter.h"
+
+class LasIOFilter : public FileIOFilter
+{
+  public:
+    LasIOFilter();
+
+    // Inherited from FileIOFilter
+    CC_FILE_ERROR
```
dgirardeau, line 28: the limit of 120 characters in the clang format file may not help to make the code more readable?

plugins/core/IO/qLASIO/include/LasScalarFieldLoader.h
```diff
@@ -0,0 +1,99 @@
...
+#include <QFileInfo>
+
+#include <ccPointCloud.h>
+
+#include <laszip/laszip_api.h>
+
+/// Class with the logic on how to load LAS dimensions values
+/// from the LAS file into a ccPointCloud's scalar field.
+///
+/// This also handle LAS extra scalar fields, as well as RGB.
```
dgirardeau, line 35: handles

plugins/core/IO/qLASIO/include/LasScalarFieldLoader.h
```diff
@@ -0,0 +1,99 @@
...
+    ///
+    /// sfInfo: Info about the current scalar field we are loading the value into
+    /// pointCloud: The point cloud where the scalar field will be loaded into
+    /// currentValue: The current value of the LAS field we are loading.
+    template <typename T>
+    CC_FILE_ERROR handleScalarField(LasScalarField &sfInfo, ccPointCloud &pointCloud, T currentValue);
+
+    /// Same thing as `handleScalarField` but for RGB.
+    ///
+    /// In LAS files, the red, green and blue channels are normal LAS fiels,
```
dgirardeau, line 66: fields

plugins/core/IO/qLASIO/include/LasScalarFieldLoader.h
```diff
@@ -0,0 +1,99 @@
...
+    /// Handles loading of LAS value into the scalar field that will be part
+    /// of the pointCloud.
+    ///
+    /// sfInfo: Info about the current scalar field we are loading the value into
+    /// pointCloud: The point cloud where the scalar field will be loaded into
+    /// currentValue: The current value of the LAS field we are loading.
+    template <typename T>
+    CC_FILE_ERROR handleScalarField(LasScalarField &sfInfo, ccPointCloud &pointCloud, T currentValue);
+
+    /// Same thing as `handleScalarField` but for RGB.
```
dgirardeau, line 64: Isn't this method dealing with GPS data?

plugins/core/IO/qLASIO/src/LasIOFilter.cpp
```diff
@@ -0,0 +1,664 @@
...
+            useLasOffset = true;
+            shift = -lasOffset;
+            if (csModeBackup != ccGlobalShiftManager::NO_DIALOG_AUTO_SHIFT)
+            {
+                parameters.shiftHandlingMode = ccGlobalShiftManager::ALWAYS_DISPLAY_DIALOG;
+            }
+        }
+    }
+
+    FileIOFilter::HandleGlobalShift(firstPoint, shift, preserveCoordinateShift, parameters, useLasOffset);
```
dgirardeau, line 70: Too be on the safe side, if HandleGlobalShift returns false, you should force 'preserveCoordinateShift' to false, and 'shift' to (0, 0, 0).

plugins/core/IO/qLASIO/src/LasScalarFieldLoader.cpp
```diff
@@ -0,0 +1,403 @@
...
+#include <ccPointCloud.h>
+#include <ccScalarField.h>
+
+#include <utility>
+
+// TODO take by move
+LasScalarFieldLoader::LasScalarFieldLoader(std::vector<LasScalarField> standardScalarFields,
+                                           std::vector<LasExtraScalarField> extraScalarFields,
+                                           ccPointCloud &pointCloud)
+    : m_standardFields(std::move(standardScalarFields)), m_extraScalarFields(std::move(extraScalarFields))
```
dgirardeau, line 31: Aren't you moving the already copied vectors this way? Shouldn't you take the vectors as references (or even double &&?). Or is it on purpose that you want to copy these vectors? (in which case I don't see the point of using std::move).

And if the aim is to have only one instance of these vectors, I don't really see the benefit compared to managing references of the input vectors in this class?

plugins/core/IO/qLASIO/src/LasScalarFieldSaver.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+#include <ccScalarField.h>
+
+LasScalarFieldSaver::LasScalarFieldSaver(std::vector<LasScalarField> standardFields,
+                                         std::vector<LasExtraScalarField> extraFields)
+    : m_standardFields(std::move(standardFields)), m_extraFields(std::move(extraFields))
+{
+}
+
+void LasScalarFieldSaver::handleScalarFields(size_t i, laszip_point &point)
```
dgirardeau, line 30: 'i' is not super explicit ;)

plugins/core/IO/qLASIO/src/LasScalarFieldSaver.cpp
```diff
@@ -0,0 +1,253 @@
...
+//#                                                                        #
+//##########################################################################
+
+#include "LasScalarFieldSaver.h"
+
+#include <laszip/laszip_api.h>
+
+#include <ccScalarField.h>
+
+LasScalarFieldSaver::LasScalarFieldSaver(std::vector<LasScalarField> standardFields,
```
dgirardeau, line 24: same questions as for the loader

plugins/core/IO/qLASIO/src/LasWaveformLoader.cpp
```diff
@@ -0,0 +1,244 @@
...
+        {
+            ccLog::Warning(QString("[LAS] Invalid waveform data packet size (0). We'll load all the "
+                                   "remaining part of the file!"));
+            fwfDataCount = fwfDataSource.size() - fwfDataSource.pos();
+        }
+    }
+    else if (laszipHeader.global_encoding & 4)
+    {
+        QFileInfo info(lasFilename);
+        QString wdpFilename = info.path() + "/" + info.completeBaseName() + ".wdp";
```
dgirardeau, line 126: I fixed a bug here (completeBaseName should be used instead of baseName for files with '.' inside their name).

Have you re-used any of the code in the LASFWFIO plugin, or have you really redone everything from scratch?

plugins/core/IO/qPDALIO/src/LASFilter.cpp
```diff
@@ -140,8 +140,8 @@ LASFilter::LASFilter()
                     3.0f,	// priority
                     QStringList{ "las", "laz" },
                     "las",
-                    QStringList{ "LAS cloud (*.las *.laz)" },
-                    QStringList{ "LAS cloud (*.las *.laz)" },
+                    QStringList{ "LAS cloud with PDAL (*.las *.laz)" },
```
dgirardeau, line 143: That was to make things clearer on my side (temporarily). This can be removed of course.

dgirardeau (changes requested):

dgirardeau:
And here is a sample file to test FWF and other things. In this file, there are 2 extended scalar fields. I haven't checked what happens, but for the first one ('Amplitude' I believe), the scalar field once loaded doesn't seem to be properly initialized (it remains gray). But maybe it's because all values are the same?

[FWF.zip](https://github.com/CloudCompare/CloudCompare/files/9641195/FWF.zip)

tmontaigu:
The extra scalar fields in this file where 'scaled extra dimensions', it was not properly handled but it should now be fixed

plugins/core/IO/qLASIO/src/LasScalarFieldLoader.cpp
tmontaigu: Yes this is moving the copy of the vectors from the arguments to the member variable.
Without the std::move, we would be copying the vector from the argiments to the member variable, introducing an unneeded extra copy step.


But I agree that in case we can simply use simple references / directly move with &&

plugins/core/IO/qLASIO/src/LasScalarFieldSaver.cpp
tmontaigu: Changed to && here

plugins/core/IO/qLASIO/include/LasDetails.h
tmontaigu: I explicitely disabled the default constructor of the `LasScalarField` class, which means that to create an instance of it
one must use the constructor that initializes this id to a proper value.

Otherwise if I were to introduce an "Invalid" variant I'd have to check at runtime for this variant in a bunch of places,
so it seems better to me to not allow an invalid / un-initialized value by deleting the default ctor

plugins/core/IO/qLASIO/include/LasDetails.h
tmontaigu: sure

plugins/core/IO/qLASIO/include/LasDetails.h
tmontaigu: I'll probably move them into a namespace

plugins/core/IO/qLASIO/include/LasScalarFieldLoader.h
tmontaigu: yup

plugins/core/IO/qLASIO/src/LasWaveformLoader.cpp
tmontaigu: From what I remember (this plugin was actually written 1-2 years ago) most of it is just redone from scratch with the LAS spec on the 2nd screen. 

I did look at how it was done in the FWF plugin to get started on how to put the fwf data and sometimes check the ideas were the same.

tmontaigu:
It could be helpful to have a clang format file tailored for the formatting of cloud compare, so that formatting is something 
we would not have to care about anymore.

asmaloney wanted to introduce sucj a config file a few years ago

plugins/core/IO/qLASIO/include/LasDetails.h
dgirardeau: Ah ok, it was just surprising to see that it was the only member that was not default initialized.

plugins/core/IO/qLASIO/include/LasDetails.h
dgirardeau: Thanks

plugins/core/IO/qLASIO/src/LasScalarFieldLoader.cpp
dgirardeau: Yes, even passing the vector as const references would only end in a single copy.

plugins/core/IO/qLASIO/src/LasWaveformLoader.cpp
dgirardeau: What about the PDAL plugin GUI? And all the features that come with it? (being able to split the classification field, tiling the cloud, etc.)? Do you plan to reuse the GUI at least? (or at least part of it?)

dgirardeau:
Yes, I remember this clang file, and that's still a good idea. We just need to apply it once in a big commit 😅

tmontaigu:
> Yes, I remember this clang file, and that's still a good idea. We just need to apply it once in a big commit 😅

I think should be done then, and since you are, I believe the most active of the few developpers, you are likely the only / most affected by the conflicts it could create with other existing branches

plugins/core/IO/qLASIO/src/LasWaveformLoader.cpp
tmontaigu: The GUI is also from scratch

- For the loading file part, the fields from classification are already split by default (with no possibility of re-merging them) as in the las standard, you don't have the choice the classification is on 5 bits, the 3 following bits are there own 1-bit fields (for LAS version < 1.4)

- For the writing, it is also from scratch and it provides more features, like
   -  selecting the version you wish to save the file into
   - which cc scalar field you want to map to LAS field with warnings If data loss due to precision would occur. 
   - And later I plan to add  the ability to add cc scalarfields as LAS extra scalar fields (as of now, cc scarlar fields are saved as LAS extra fields only if they were LAS extra scalar field in the first place)

- I didn't add tiling, I always found it weird that the LAS filter was the only one that had this option.
 But if its really necessary it can be re-added

tmontaigu:
All global functions where moved to namespaces or into classes as static functions

dgirardeau:
Ok, I'll try to pressure test this new plugin ASAP

dgirardeau:
(do you think you'll have some time to add some comments and code documentation later? 👼 )

tmontaigu:
Yes, are there specific points you see that particularly needs commenta ?

plugins/core/IO/qLASIO/src/LasWaveformLoader.cpp
dgirardeau: Let's see how this is received then ;)

dgirardeau:
No it was more a general comment, and especially regarding the doxygen documentation. I see that you already added some comments 👍 

Otherwise I have rebased the branch. This way I'll be able to test all the recent changes at once.

dgirardeau:
I updated some code to get a better look for the UI and to fix some simple things.

And I did some tests. Here is what I observed so far:
- when loading several files at once, making a selection of fields to open, and eventually hitting the "Apply all" button, the dialog will show up again and ask me to redo the selection
- when saving, the point format doesn't seem to be chosen in a smart way (in LasDetails::SelectBestVersion ?). For instance if I add a RGB field or an additional scalar field, I still get the original format (Type 1, Version 1.2). That's quite dangerous to me. The best point format should be selected by default. Especially since most of the users don't know which point format corresponds to what.
- I have not understood how to save a custom made scalar field as an 'extended' LAS field?
- when loading a LAS file, all the meta data pieces or stored in a big 'savedInfo' struct. Two issues here:
  - the user can't read some interesting pieces of info anymore (the LAS scale, etc.)
  - it breaks the compatibility with the old plugins, especially for users who have BIN files containing clouds loaded with the 2 other plugins. We should use the same Meta data keys as before for these pieces of data.

Nice to have (not mandatory, but it would have been great to have that...):
- the doxygen comments style does not match CC's style (//! for the short decsription/title, and /** ... **/ for the blocks)
- about the clangformat file:
  - it does not indent the contents in namespaces (} (that doesn't help readability)
  - it places the * or & for pointers and reference next to the variable name instead of the type, which is not CloudCompare's default syntax
  - I guess we can fix all that by retrieving and applying Andy's clangformat file ;)

dgirardeau:
And I almost forgot one of the newest feature added recently: the option to set the GPS time shift
image

tmontaigu:
> * when loading several files at once, making a selection of fields to open, and eventually hitting the "Apply all" button, the dialog will show up again and ask me to redo the selection

will fix

> * when saving, the point format doesn't seem to be chosen in a smart way (in LasDetails::SelectBestVersion ?). For instance if I add a RGB field or an additional scalar field, I still get the original format (Type 1, Version 1.2). That's quite dangerous to me. The best point format should be selected by default. Especially since most of the users don't know which point format corresponds to what.

`LasDetails::SelectBestVersion` Selects the point format in a smart way, its just that for clouds with LAS metadata
I used this metadata and did not call the `LasDetails::SelectBestVersion`. The fix is to simply call this function everytime.

> * I have not understood how to save a custom made scalar field as an 'extended' LAS field?

This is not yet possible I will add it.

> * when loading a LAS file, all the meta data pieces or stored in a big 'savedInfo' struct. Two issues here:
>   
>   * the user can't read some interesting pieces of info anymore (the LAS scale, etc.). And this big struct surprisingly shows 'QWidget*' in the description part 😅.
>   * it breaks the compatibility with the old plugins, especially for users who have BIN files containing clouds loaded with the 2 other plugins. We should use the same Meta data keys as before for these pieces of data.

Ok, I'll break the `SavedInfo` class into multiple metadata keys.
I haven't found a way so that, the big struct shows something else than just  "QWidget*".

tmontaigu:
> (and I don't know what happened to the 2 other options? To ignore the fields with only one constant value, and to force the colors to be considered as 8bits colors? 

These can be re added

> And I almost forgot one of the newest feature added recently: the option to set the GPS time shift

Will also try to port this

dgirardeau:
Hi Thomas, do you intend to make new commits on this PR? Do you want help for any of the remaining points above?

tmontaigu:
I have stashed changes that splits the metadatas that i need to finish and push

Then i will tackle the ui to add custom scalar field

The thing i'm not sure how to correctly do is the "apply for all"

tmontaigu:
I split the metadatas and added a ui system that allows to add extra scalar fields

tmontaigu:
Ok, the famous problem where I don't need some `#include` on my machine but on other systems/compiler i do 😅

tmontaigu:
What should be the behaviour of "applyForAll" when the current file to load does not have the same fields (due to not being same point format id and/or having different extra scalar field) ?

Should it ignore and go on? Or should the dialog pop up again just like it does if the last shift used is not valid ?

dgirardeau:
Ah good question. In the ASCII filer, we display the dialog again as soon as the previous 'opening sequence' does not apply to the new file:
https://github.com/CloudCompare/CloudCompare/blob/master/libs/qCC_io/src/AsciiFilter.cpp#L467

tmontaigu:
Ok so I added the apply for all, on file loading, the dialog will show up again if the file does not have the same fields || extra scalar fields as the previous one

What happens of the commandline mode, do FileIOFilters get a automatic command line option? 
Can they set their own one ?

dgirardeau:
Yes normally there's nothing to do

dgirardeau:
I just rebased the branch, and I'll try to test it again.

dgirardeau:
When I open a file multiple times, the checkboxes pile up :D

image

dgirardeau:
Have you looked at this discussion by the way?
https://github.com/CloudCompare/CloudCompare/issues/1695

There's probably an opportunity to improve that in this new plugin?

tmontaigu:
There is 2 aspect to this issue: standard fields and extra fields.


Extra fields have a mechanism to specify the value to consider as no data, so it would be possible to add
the ui stuff to let the user tell allow customizing it.

Standard fields however do not have a "no data" concept, so we can only provide a way to let the user select to which
value the ccScalarFields NaN will be converted to. 
That does not means that upon reading that file again CC or other software will know the no data values for the different fields, as again, standard fields do not have a way to mark values as no data.

Some fields have values which can 'fit' as being used for no data and it often is 0, (eg classification == 0, means unclassified, rgb == 0 makes sense for no data).

We could add the ui stuff in the second tab of the save dialog to let user chose on a per field basis the value to which NaNs
will be converted.

dgirardeau:
Yes, maybe that's a bit heavy for something rarely needed. For now we could just use these default and 'obvious' values to replace NaN values in standard fields.

And of course use the default mechanism for extended fields.

tmontaigu:
I don't think its that much work, and can be made not that intrusive.

However I think that this issue should be tackled once this plugin is accepted as a replacement for the two others

dgirardeau:
Yes, makes sense. And it should be accepted very soon from what I see today ;)

dgirardeau:
Okay, I've made a pass on the code and I'm ready to merge if you are good with my last changes

tmontaigu:
Looks good to me

Issue #107, opened 2013-10-30
Getting segfault when loading back a cube


luca-penasa:
How to reproduce:
1. Open primitive factory. Create a cube
2. Save the cube to BIN file
3. Reload.
4. Segfault!

anybody could confirm?
tracing the segfault comes out that cc crashes trying recompute the bounding box...

The same segfault happens for cylinder (and probably other primitives)

PR #1690, opened 2022-09-28
Bug fix: Check for valid E57 pose rotation

nigels-com:
CloudCompare seems to be having trouble reading E57 that have a pose rotation quaternion set to all zeros.
Likely this is malformed according to the E57 specification, but this change adds a check.
See also: [libE57Format #107](https://github.com/asmaloney/libE57Format/issues/107)

### DIFF
--- a/plugins/core/IO/qE57IO/src/E57Filter.cpp
+++ b/plugins/core/IO/qE57IO/src/E57Filter.cpp
@@ -1399,6 +1399,10 @@ static bool GetPoseInformation(const e57::StructureNode& node, ccGLMatrixd& pose
 	bool validPoseMat = false;
 	if (node.isDefined("pose"))
 	{
+		CCCoreLib::SquareMatrixd rotMat(3);
+		rotMat.toIdentity();
+		rotMat.toGlMatrix(poseMat.data());
+
 		e57::StructureNode pose(node.get("pose"));
 		if (pose.isDefined("rotation"))
 		{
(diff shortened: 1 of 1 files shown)


plugins/core/IO/qE57IO/src/E57Filter.cpp
```diff
@@ -1399,6 +1399,10 @@ static bool GetPoseInformation(const e57::StructureNode& node, ccGLMatrixd& pose
 	bool validPoseMat = false;
 	if (node.isDefined("pose"))
 	{
+		CCCoreLib::SquareMatrixd rotMat(3);
```
dgirardeau, line 1402: Is it necessary do that here since 'validPoseMat' is false? You could simply call 'validPoseMat.toIdentity()' at the beginning of the method, just in case.

nigels-com: I was considering the case that translation is defined, but not rotation.  Seems reasonable to initialise to identity matrix, in the absence of rotation?

dgirardeau: Yes it's enough to initialize poseMat to identity directly.

nigels-com: Updated as discussed.

dgirardeau (approved):

dgirardeau:
Thanks!

nigels-com:
Cheers!  Thanks!

PR #1697, opened 2022-10-13
ccConsole: avoid printing 2 times the messages to stdout

tmontaigu:
When CloudCompare is compiled in debug mode,
and run in command line mode with the -SILENT option the messages "printed" to the ccConsole would be output 2 times in the users console (stdout).

Once due du to -SILENT  flag, then a second time due to being compiled in DEBUG mode (I guess to be able to debug the ccConsole itself).

This makes the output harder to read.

This commit makes it so that messages are only printed once in this configuration


dgirardeau (approved):

dgirardeau:
Ah ah nice catch ;)

PR #1715, opened 2022-11-29
LASFWFFilter bug

p-leroy:
return number, number of returns and scan angle rank were not properly handled with point format equal to six or above.

### DIFF
--- a/plugins/core/IO/qLASFWFIO/src/LASFWFFilter.cpp
+++ b/plugins/core/IO/qLASFWFIO/src/LASFWFFilter.cpp
@@ -691,10 +691,24 @@ CC_FILE_ERROR LASFWFFilter::saveToFile(ccHObject* entity, const QString& filenam
 					laspoint.set_intensity(static_cast<U16>(it->getSafeValue(i)));
 					break;
 				case LAS_RETURN_NUMBER:
-					laspoint.set_return_number(static_cast<U8>(it->getSafeValue(i)));
+					if (pointFormatSixOrAbove)
+					{
+						laspoint.set_extended_return_number(static_cast<U8>(it->getSafeValue(i)));
+					}
+					else
+					{
+						laspoint.set_return_number(static_cast<U8>(it->getSafeValue(i)));
+					}
 					break;
 				case LAS_NUMBER_OF_RETURNS:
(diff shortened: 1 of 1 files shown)


dgirardeau:
Nice catch!

By the way, this plugin is now deprecated, and the new versions will now rely on the unified "LAS I/O" plugin (https://github.com/CloudCompare/CloudCompare/tree/master/plugins/core/IO/qLASIO) developed by @tmontaigu. This new plugin relies on LASzip (and hopefully this bug is not present anymore, but you may want to double check that?).

p-leroy:
At first sight, it think that the problem is also present in the new plugin (I have done a little test trying to save in point format 9). But doing the testing, I find a more important bug when you want to save full waveforms, I will send a PR for that.

dgirardeau:
I believe you can manually change the point format for the output file in the new plugin?

p-leroy:
Yes, you can. If to chose 9 for the output and if the input was in format 4, some fields are not saved properly.

dgirardeau:
@tmontaigu will you be able to look at that? Or should I give it a try?

tmontaigu:
The fields not properly saved were the fwf i think, https://github.com/CloudCompare/CloudCompare/pull/1716

dgirardeau:
Oh, there's not the same concept in LASzip as in LASlib (with set_extended_return_number / set_return_number?). I though that's what Paul was referring to in his last comment.

tmontaigu:
It is alreay using the equivalents https://github.com/CloudCompare/CloudCompare/blob/88807812f5956dccec66450e0b13517a265a2dfc/plugins/core/IO/qLASIO/src/LasScalarFieldSaver.cpp#L140

dgirardeau:
Ah good. @p-leroy, can you be more precise on your latest comment? Which field are not properly saved?

p-leroy:
I put a little dataset online to show the issue. I can open it with the plugin, it is in point format 4. If I save it in format 4, it is OK, If I save it in format 9,  it is not possible to open it properly, there are errors at the opening. At the current time, only a very limited number of points are read (297/10988439). Will need to solve that before having a closer look to the scalar fields.
https://filesender.renater.fr/?s=download&token=6da14fb4-0e44-4e04-b566-d8477cfce66f

PR #1717, opened 2022-12-02
QLAS_IO : one bug corrected and a feature added to the command line.

p-leroy:
1) The format was not correctly handled when saving float or double extra fields.
2) Added the automatic saving of additional scalar fields when using the command line : usefull when computing c2c distances and saving to LAZ format.

### DIFF
--- a/plugins/core/IO/qLASIO/include/LasScalarFieldSaver.h
+++ b/plugins/core/IO/qLASIO/include/LasScalarFieldSaver.h
@@ -65,24 +65,14 @@ class LasScalarFieldSaver
 		{
 			*reinterpret_cast<T*>(dest) = std::numeric_limits<T>::max();
 		}
-		else if (value < std::numeric_limits<T>::min())
+		else if (value < std::numeric_limits<T>::lowest())
 		{
-			*reinterpret_cast<T*>(dest) = std::numeric_limits<T>::min();
+			*reinterpret_cast<T*>(dest) = std::numeric_limits<T>::lowest();
 		}
 		else
 		{
 			*reinterpret_cast<T*>(dest) = static_cast<T>(value);
 		}
-	}
-	template <typename T>
-	static void WriteScalarValueAsFltDbl(ScalarType value, uint8_t* dest)
-	{
-		if (value > std::numeric_limits<T>::max())
-			*reinterpret_cast<T*>(dest) = std::numeric_limits<T>::max();
-		else if (value < -std::numeric_limits<T>::max())
-			*reinterpret_cast<T*>(dest) = -std::numeric_limits<T>::max();
-		else
-			*reinterpret_cast<T*>(dest) = static_cast<T>(value);
 	}
 
   private:
--- a/plugins/core/IO/qLASIO/include/LasScalarFieldSaver.h
+++ b/plugins/core/IO/qLASIO/include/LasScalarFieldSaver.h
@@ -74,6 +74,16 @@ class LasScalarFieldSaver
 			*reinterpret_cast<T*>(dest) = static_cast<T>(value);
 		}
 	}
+	template <typename T>
+	static void WriteScalarValueAsFltDbl(ScalarType value, uint8_t* dest)
+	{
+		if (value > std::numeric_limits<T>::max())
+			*reinterpret_cast<T*>(dest) = std::numeric_limits<T>::max();
+		else if (value < -std::numeric_limits<T>::max())
+			*reinterpret_cast<T*>(dest) = -std::numeric_limits<T>::max();
+		else
+			*reinterpret_cast<T*>(dest) = static_cast<T>(value);
+	}
 
   private:
 	std::vector<LasScalarField>      m_standardFields;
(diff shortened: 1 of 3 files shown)


plugins/core/IO/qLASIO/include/LasScalarFieldSaver.h
```diff
@@ -74,6 +74,16 @@ class LasScalarFieldSaver
...
+	template <typename T>
+	static void WriteScalarValueAsFltDbl(ScalarType value, uint8_t* dest)
+	{
+		if (value > std::numeric_limits<T>::max())
+			*reinterpret_cast<T*>(dest) = std::numeric_limits<T>::max();
+		else if (value < -std::numeric_limits<T>::max())
+			*reinterpret_cast<T*>(dest) = -std::numeric_limits<T>::max();
+		else
+			*reinterpret_cast<T*>(dest) = static_cast<T>(value);
+	}
```
tmontaigu, line 86: I would suggest changing `std::numeric_limits<T>::min()` to `std::numeric_limits<T>::lowest()` in the `WriteScalarValueAs` function it would help us avoid duplicating

tmontaigu (changes requested):

plugins/core/IO/qLASIO/src/LasIOFilter.cpp
```diff
@@ -598,6 +599,50 @@ CC_FILE_ERROR LasIOFilter::saveToFile(ccHObject* entity, const QString& filename
...
+					found = true;
+			if (!found)
+			{
+				ccLog::Print("[LAS] scalar field " + QString(sfName) + " will be saved automatically in the extra fields of the output file");
+				LasExtraScalarField field;
+				const std::string stdName = sfName;
+				strncpy(field.name, stdName.c_str(), LasExtraScalarField::MAX_NAME_SIZE);
+
+				const std::string stdDescription = "";
+				strncpy(field.description, stdDescription.c_str(), LasExtraScalarField::MAX_DESCRIPTION_SIZE);
```
tmontaigu, line 625: `stdDescription` will always be empty, so this is not needed

plugins/core/IO/qLASIO/src/LasIOFilter.cpp
```diff
@@ -598,6 +599,50 @@ CC_FILE_ERROR LasIOFilter::saveToFile(ccHObject* entity, const QString& filename
...
+				if (stdName.size() > LasExtraScalarField::MAX_NAME_SIZE)
+				{
+					ccLog::Warning("[LAS] Extra Scalar field name '%s' is too long and will be truncated",
+								   stdName.c_str());
+				}
+				if (stdDescription.size() > LasExtraScalarField::MAX_DESCRIPTION_SIZE)
+				{
+					ccLog::Warning("[LAS] Extra scalar field description '%s' is too long and will be truncated",
+								   stdDescription.c_str());
+				}
```
tmontaigu, line 636: `stdDescription` will always be empty, so this is not needed

plugins/core/IO/qLASIO/include/LasScalarFieldSaver.h
dgirardeau: you mean `-std::numeric_limits<T>::max()` by `std::numeric_limits<T>::lowest()` I guess?

plugins/core/IO/qLASIO/src/LasIOFilter.cpp
```diff
@@ -598,6 +599,50 @@ CC_FILE_ERROR LasIOFilter::saveToFile(ccHObject* entity, const QString& filename
...
+		uint sfCount = pointCloud->getNumberOfScalarFields();
+		for (uint index = 0; index < sfCount; index++)
+		{
+			ccScalarField *sf = static_cast<ccScalarField*>(pointCloud->getScalarField(index));
+			const char* sfName = sf->getName();
+			bool found = false;
+			for (auto& el : params.standardFields)
+				if (strcmp(sfName, el.name()) == 0)
+					found = true;
+			for (auto& el : params.extraFields)
```
dgirardeau, line 614: if (!found)
 ...

plugins/core/IO/qLASIO/src/LasIOFilter.cpp
dgirardeau: you mean `-std::numeric_limits<T>::max()` by `std::numeric_limits<T>::lowest()` I guess?

plugins/core/IO/qLASIO/src/LasIOFilter.cpp
p-leroy: I do not understand the comment, can you explain what you mean?

tmontaigu: I think he meant to do

```C++
for (auto& el : params.standardFields)
    if (strcmp(sfName, el.name()) == 0)
        found = true;


if (!found) {
   for (auto& el : params.extraFields)
     // ...
}
```

p-leroy: OK, got it now!

dgirardeau (approved):

dgirardeau:
Thanks for the update, Paul

PR #1729, opened 2022-12-20
LAS: Add tiling back

tmontaigu:
This adds the LAS tiling option back

- [x] Need to add progress bar
- [x] Change dialog buttons when tiling tab is selected


tmontaigu:
should be good now

plugins/core/IO/qLASIO/include/LasTiler.h
```diff
@@ -0,0 +1,69 @@
...
+//#                                                                        #
+//##########################################################################
+
+#include <QString>
+
+#include <FileIOFilter.h>
+
+#include <laszip/laszip_api.h>
+
+enum class LasTilingDimensions {
```
dgirardeau, line 26: the convention is CloudCompare is generally to have the '{' symbol on the next line 👼

dgirardeau: here and below

plugins/core/IO/qLASIO/include/LasTiler.h
```diff
@@ -0,0 +1,69 @@
...
+        }
+        return 0;
+    }
+
+    inline size_t index1() const {
+        switch (dims) {
+            case LasTilingDimensions::XY:
+                return 1;
+            case LasTilingDimensions::XZ:
+                return 2;
```
dgirardeau, line 55: Q_FALLTHROUGH() instead?

plugins/core/IO/qLASIO/src/LasTiler.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+    std::vector<laszip_POINTER> writers;
+    writers.resize(options.numTiles0 * options.numTiles1, nullptr);
+
+    CC_FILE_ERROR error = CC_FERR_NO_ERROR;
+
+    const size_t index0 = options.index0();
+    const size_t index1 = options.index1();
+
+    double cloudMins[3] = {
```
dgirardeau, line 63: I guess you can remove the '=' signs here and below?

plugins/core/IO/qLASIO/src/LasTiler.cpp
```diff
@@ -0,0 +1,208 @@
...
+    };
+
+    double cloudBBox[3] = {
+            laszipHeader->max_x - laszipHeader->min_x,
+            laszipHeader->max_y - laszipHeader->min_y,
+            laszipHeader->max_z - laszipHeader->min_z,
+    };
+
+    double tileSize[2] = {
+            cloudBBox[index0] / static_cast<double>(options.numTiles0),
```
dgirardeau, line 76: no need to use static_cast since cloudBBox is already composed of double values

plugins/core/IO/qLASIO/src/LasTiler.cpp
```diff
@@ -0,0 +1,208 @@
...
+        tileI = std::min(tileI, static_cast<size_t>(options.numTiles0 - 1));
+        auto tileJ = static_cast<size_t>((laszipCoordinates[index1] - cloudMins[index1]) / tileSize[1]);
+        tileJ = std::min(tileJ, static_cast<size_t>(options.numTiles1 - 1));
+
+        const size_t tileIndex = (tileI * options.numTiles1) + tileJ;
+
+        laszip_POINTER &writer = writers[tileIndex];
+
+        if (writer == nullptr) {
+            std::string outputName;
```
dgirardeau, line 120: Wouldn't it be simpler to use QStrings to do this and just convert it to stdString when needed? (and baseName and extension can remain QStrings)
if (...)
{
    outputName = options.outputDir + '/';
}
outputName += baseName + '\_' + QString::number(tileI) + '\_'  + QString::number(tileK) + '.' + extension;

(or equivalent)

plugins/core/IO/qLASIO/src/LasTiler.cpp
```diff
@@ -0,0 +1,208 @@
...
+    laszip_close_reader(laszipReader);
+    laszip_clean(laszipReader);
+    laszip_destroy(laszipReader);
+
+
+    if (error == CC_FERR_NO_ERROR)
+    {
+        timer.elapsed();
+        qint64 elapsed = timer.elapsed();
+        int32_t minutes = timer.elapsed() / (1000 * 60);
```
dgirardeau, line 199:
```suggestion
        int32_t minutes = elapsed / (1000 * 60);
```

plugins/core/IO/qLASIO/src/LasTiler.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+    laszip_close_reader(laszipReader);
+    laszip_clean(laszipReader);
+    laszip_destroy(laszipReader);
+
+
+    if (error == CC_FERR_NO_ERROR)
+    {
+        timer.elapsed();
+        qint64 elapsed = timer.elapsed();
```
dgirardeau, line 198: elapsed_ms? (to make the more units explicit)

plugins/core/IO/qLASIO/src/LasTiler.cpp
dgirardeau: And you can probably retrieve baseName and extension from fileInfo just here (as you don't need them elsewhere)

plugins/core/IO/qLASIO/include/LasTiler.h
dgirardeau: Don't we have a clang file in this plugin/project by the way?

tmontaigu: No there is only a `.clang-format` file in the qLASIO plugin, but ideally there should one at the root of the cloudcompare project and it would be used for every file within the project

plugins/core/IO/qLASIO/src/LasTiler.cpp
tmontaigu: I did it that way to avoid too many allocation from QString to std::string, but I can change that

dgirardeau: Yes, it's a bit 'heavy' for something that simple

plugins/core/IO/qLASIO/include/LasTiler.h
dgirardeau: Isn't this file part of the qLASIO plugin?

plugins/core/IO/qLASIO/src/LasTiler.cpp
tmontaigu: You mean 
```C++
 double cloudMins[3] { ... }
```
?

That is not legal syntax 🤔

plugins/core/IO/qLASIO/include/LasTiler.h
tmontaigu: A yes sure, I had a brain fart  🧠 💨

plugins/core/IO/qLASIO/src/LasTiler.cpp
dgirardeau: Ah yes, there's no 'constructor' for C-arrays, my bad. I'm more used to std::array and std::vectors now...

dgirardeau (approved):

PR #1733, opened 2023-01-04
Fixed a variable in the copy constructor call to maxValidIndex

michael-salmas:
Noticed the wrong variable assigned to maxValidIndex when using the copy constructor for the Grid structure


dgirardeau:
Good catch! Thanks.

PR #1741, opened 2023-01-21
Various updates

dgirardeau:


plugins/core/IO/qLASIO/src/LasOpenDialog.cpp
```diff
@@ -31,13 +31,19 @@ static QListWidgetItem* CreateItem(const char* name)
...
+	}
+
+	for (int i = 0; i < list->count(); ++i)
+	{
+		if (list->item(i)->text() == name)
 		{
-			return list.item(i)->checkState() == Qt::Checked;
+			return list->item(i)->checkState() == Qt::Checked;
 		}
 	}
```
tmontaigu, line 48: Not sure its worth changing the input from `const QListWidget& list` to `const QListWidget* list`

qCC/ccGraphicalSegmentationTool.cpp
```diff
@@ -903,7 +903,6 @@ void ccGraphicalSegmentationTool::segment(bool keepPointsInside, ScalarType clas
 
 		// we project each point and we check if it falls inside the segmentation polyline
 #if defined(_OPENMP)
-		omp_set_num_threads(std::max(1, omp_get_max_threads() - 1)); // always leave one thread/core to let the application breath
```
tmontaigu, line 906: Is removing this line intentional ?

libs/qCC_db/src/ccPointCloud.cpp
```diff
@@ -6120,3 +6120,119 @@ bool ccPointCloud::shiftPointsAlongNormals(PointCoordinateType shift)
...
+	: w(0)
+	, h(0)
+	, validCount(0)
+	, minValidIndex(0)
+	, maxValidIndex(0)
+{
+	sensorPosition.toIdentity();
+}
+
+ccPointCloud::Grid::Grid(const Grid& grid)
```
tmontaigu, line 6134: Wouldn't having `Grid(const Grid& grid) = default` in the `Grid` class declarion be shorter ?

plugins/core/IO/qLASIO/src/LasOpenDialog.cpp
dgirardeau: It was to be able to test the pointer at one place (instead of dereferencing it without testing it, or testing it at 2 different places).

qCC/ccGraphicalSegmentationTool.cpp
dgirardeau: Yes, it's not worth the sweat to do this for some short operations (and on some processors with a few cores, it makes the computation a lot slower). I should probably add a minimum to the number of threads used... Let me fix that (again)

libs/qCC_db/src/ccPointCloud.cpp
dgirardeau: That would probably work indeed.

PR #1748, opened 2023-02-12
Command line improvements

dgirardeau:


dgirardeau:
I wonder if the initial step to manage single quotes in the command line (in ccCommandLineParser::Parse) is actually required on Linux? Or are single quotes properly managed by default? (hopefully double quotes are also behaving just as on Windows 😅 )

rjanvier:
On Unix systems (including macOS) all type of shell (at least bash and zsh) will somehow indicate to the user if single or a double quote sequence is incomplete.

dgirardeau:
But will the arguments passed to CC be properly split as single 'arguments' when between single quotes? Because on Windows, only double quotes behave like this. For instance "my file.bin" will be one argument, but 'my file.bin' will be two arguments (passed to main(arc, argv))

rjanvier:
It's "fine" to use single quote. Using escape character is also possible in the case of a space in a path.

rjanvier (commented):
I would wait for @tmontaigu input on this, but this change should be harmless on Linux/macOS.

tmontaigu:
On bash-like shells, `'` and `"` will make it so that "my file.bin" or 'my file.bin' as considered as one argument so this change should be harmless

dgirardeau:
Thanks guys!

PR #1765, opened 2023-04-01
LAS: Allow entering custom scale values

### LINKED ISSUE #1764: Can't change custom scale when I save LAS file
**Describe the bug**
When saving the point cloud (las) I cannot activate the "Custom scale" output option (the field is locked).

**How to reproduce**
1. Open LAZ file + select it
2. Save > LAS type
3. Basic Params > Custom scale > I can't change the value

export-option

**Expected behaviour**
I want to change the custom value

**Your environment**
 - CC Version (or git hash/tag): 2.13 alpha (Mar 30 2023)
 - OS & Version: Windows 11 x64
 - Graphics card: -
 - Qt Version (if compiling): -

tmontaigu:
Clicking on the "custom scale" radio button did not enable the spin box, so the user could not enter the custom scale it wanted to use.

This commits adds 2 spin box, so that there a 3 (one per dimension X, Y, Z). Also this commits makes it so that when clicking on the custom scale radio button the 3 spin box are enabled and value can be changed, when the radio button is no longer selected, the 3 spin box are disabled

Fixes #1764

### DIFF
--- a/plugins/core/IO/qLASIO/include/LasSaveDialog.h
+++ b/plugins/core/IO/qLASIO/include/LasSaveDialog.h
@@ -75,6 +75,7 @@ class LasSaveDialog : public QDialog
 	void handleSelectedVersionChange(const QString&);
 	void handleSelectedPointFormatChange(int index);
 	void handleComboBoxChange(int index);
+	void handleCustomScaleButtontoggled(bool checked);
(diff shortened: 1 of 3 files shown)


dgirardeau (approved):

dgirardeau:
Thanks for the fix!

PR #1775, opened 2023-04-17
Don't canonicalize OBJ path for MTL loading. Fixes #1772

### LINKED ISSUE #1772: Symlink following is too aggressive when opening OBJ+MTL
**Describe the bug**

CloudCompare fails to open a pair of `.obj`+`.mtl` files when these two are symlinks into separate directories:

image

Error: `[OBJ] Failed to load material file!`

Other programs, such as Meshlab, succeed opening in this case.

The issue here is most likely that CloudCompare resolves symlinks too aggressively itself.
Likely, CloudCompare should not follow a symlink and then open the OBJ from the dereferenced file's parent folder; instead it should simply `open()` the symlink itself and let the OS derefence it.

**How to reproduce**

Repro scenario:

```
% tree
.
├── joined-symlinks-dir
│   ├── untitled.mtl -> ../mtl-dir/untitled.mtl
│   └── untitled.obj -> ../obj-dir/untitled.obj
├── mtl-dir
│   └── untitled.mtl
└── obj-dir
    └── untitled.obj
```

Repro as a tar: [obj-symlink-test.tar.zip](https://github.com/CloudCompare/CloudCompare/files/11203061/obj-symlink-test.tar.zip)

**Expected behaviour**

CloudCompare opens symlinked OBJs just as well as Meshlab.

**Your environment**
 - CC Version (or git hash/tag): 2.12.1
 - OS & Version: Ubuntu 22.04.1

Thanks!

nh2:
Fixes #1772.

### DIFF
--- a/plugins/core/IO/qCoreIO/src/ObjFilter.cpp
+++ b/plugins/core/IO/qCoreIO/src/ObjFilter.cpp
@@ -1134,7 +1134,7 @@ CC_FILE_ERROR ObjFilter::loadFile(const QString& filename, ccHObject& container,
 						mtlFilename = mtlFilename.left(mtlFilename.size() - 1);
 					}
 					ccLog::Print(QString("[OBJ] Material file: ") + mtlFilename);
-					QString mtlPath = QFileInfo(filename).canonicalPath();
+					QString mtlPath = QFileInfo(filename).path();
 					//we try to load it
 					if (!materials)
 					{


dgirardeau:
Redundant with https://github.com/CloudCompare/CloudCompare/commit/0fb46ddb6fd3ed35ddb7c92a8b8befb141b87668

PR #1790, opened 2023-05-24
a small fix on Salome polylines

prascle:
last commit on Salome polylines always closes the polylines...


dgirardeau (approved):
Oops! Good catch!

PR #1798, opened 2023-06-04
Loss of accuracy when calculating the GPS time shift.

p-leroy:
This loss has been experienced when selecting a point with several returns and trying to get all points with the same time with the filter tool. Too many points are returned.
This suggestion is similar to what were used in an old version of qPDALIO (changed in commit 90b2423).

### DIFF
--- a/plugins/core/IO/qLASIO/include/LasWaveformLoader.h
+++ b/plugins/core/IO/qLASIO/include/LasWaveformLoader.h
@@ -33,8 +33,8 @@ struct LasWaveformLoader
 
 	void loadWaveform(ccPointCloud& pointCloud, const laszip_point& currentPoint) const;
 
-	unsigned                       fwfDataCount{0};
-	unsigned                       fwfDataOffset{0};
+	uint64_t                       fwfDataCount{0};
+	uint64_t                       fwfDataOffset{0};
 	bool                           isPointFormatExtended{false};
 	ccPointCloud::FWFDescriptorSet descriptors;
 };


plugins/core/IO/qLASIO/src/LasScalarFieldLoader.cpp
```diff
@@ -235,10 +235,12 @@ LasScalarFieldLoader::handleGpsTime(LasScalarField& sfInfo, ccPointCloud& pointC
 			return CC_FERR_NOT_ENOUGH_MEMORY;
 		}
 
+
+
 		double timeShift;
 		if (std::isnan(m_manualTimeShiftValue))
 		{
-			timeShift = static_cast<int64_t>(currentValue / 10000.0) * 10000.0;
+			timeShift = currentValue;
```
dgirardeau, line 243: The aim of this rounded shift is to be easier for users to copy it around, or to mentally or numerically apply it to their data. So I'm not very happy to let it go ;)

Can you explain better where does the accuracy loss come from in your case? Is it because the shift is not large enough maybe?

dgirardeau (changes requested):

plugins/core/IO/qLASIO/src/LasScalarFieldLoader.cpp
dgirardeau: So I used your file to look at what happens, and while using a shift without digits makes the problem worse, the problem still exists with all the digits (the limitation for float values is an issue of ratio between the min and max values, so as soon as the GPS time span is over 100, you only have 10^-6 worth of information).

Here is a snapshot of the raw GPS values and the shifted GPS values (even after your change):
[10:30:00] (-2478134, 5479949, -37562) --> GPS time = 366883269.549650251865 --> 100.064750671387
[10:30:00] (-2478206, 5479886, -37580) --> GPS time = 366883269.549655914307 --> 100.064758300781
[10:30:00] (-2478164, 5479890, -37050) --> GPS time = 366883269.549661636353 --> 100.064758300781
[10:30:00] (-2478231, 5479830, -37048) --> GPS time = 366883269.549667358398 --> 100.064765930176
[10:30:00] (-2478271, 5479786, -36920) --> GPS time = 366883269.549673020840 --> 100.064773559570
[10:30:00] (-2478320, 5479737, -36843) --> GPS time = 366883269.549678742886 --> 100.064781188965
[10:30:00] (-2478381, 5479681, -36819) --> GPS time = 366883269.549684464931 --> 100.064781188965
[10:30:00] (-2478505, 5479587, -37113) --> GPS time = 366883269.549690127373 --> 100.064788818359
[10:30:00] (-2478542, 5479545, -36985) --> GPS time = 366883269.549695849419 --> 100.064796447754
[10:30:00] (-2478605, 5479488, -36986) --> GPS time = 366883269.549701571465 --> 100.064804077148

dgirardeau: My point is that this GPS time shift is not sufficient sadly 😅

dgirardeau: And with this change, the GUI needs to be updated as well!

p-leroy: Could the points with similar GPS time belong to the same shot? If you have examples with three points, it can be checked with the GUI. If so, that means that points belonging to the same shot have different GPS time (is it normal in a LAS file? I do not know), but that also means that they end up with the same time after reading, what is not so bad...

dgirardeau: Sorry I forgot to say that I filtered the points with a return index of 1.

p-leroy: Argl

PR #1803, opened 2023-06-13
assign q3DMASC and qQTreeIso to the correct submod_plugins variable in build file

rjanvier:
This fixes the scenario where one would clone CC with "minimal" submodule depedencies (i.e. CCCorelib)


dgirardeau (commented):
Good catch!

dgirardeau (approved):

PR #1807, opened 2023-06-19
Update LibE57Format to v3.0.1

rjanvier:


dgirardeau:
For an unknown reason, I totally missed this PR 😅 

I'll close it since we now move to v3.1.1

PR #1842, opened 2023-08-22
qlasio bug correction: unsigned is too short for a size, use long long

p-leroy:


plugins/core/IO/qLASIO/include/LasWaveformLoader.h
```diff
@@ -33,8 +33,8 @@ struct LasWaveformLoader
 
 	void loadWaveform(ccPointCloud& pointCloud, const laszip_point& currentPoint) const;
 
-	unsigned                       fwfDataCount{0};
-	unsigned                       fwfDataOffset{0};
+	long long                       fwfDataCount{0};
```
dgirardeau, line 36: maybe use the more explicit uint64_t type? (and did you mean to remove the 'unsigned' part?)

tmontaigu: I was going to propose this also, uint64_t would be better

p-leroy: yep!

tmontaigu (approved):

PR #1844, opened 2023-08-25
Better to have "1." as a default value for the scales.

p-leroy:
Better to have "1." as a default value for the scales.
fwfDataOffset was not correct for internal waveforms

### DIFF
--- a/plugins/core/IO/qLASIO/include/LasExtraScalarField.h
+++ b/plugins/core/IO/qLASIO/include/LasExtraScalarField.h
@@ -125,7 +125,7 @@ class LasExtraScalarField
 	uint8_t noData[MAX_DIM_SIZE][8]           = {0};
 	uint8_t mins[MAX_DIM_SIZE][8]             = {0};
 	uint8_t maxs[MAX_DIM_SIZE][8]             = {0};
-	double  scales[MAX_DIM_SIZE]              = {0.0};
+	double  scales[MAX_DIM_SIZE]              = {1.0};
 	double  offsets[MAX_DIM_SIZE]             = {0.0};
 
 	// These are added by us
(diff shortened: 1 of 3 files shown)


dgirardeau (approved):

plugins/core/IO/qLASIO/include/LasExtraScalarField.h
```diff
@@ -125,7 +125,7 @@ class LasExtraScalarField
 	uint8_t noData[MAX_DIM_SIZE][8]           = {0};
 	uint8_t mins[MAX_DIM_SIZE][8]             = {0};
 	uint8_t maxs[MAX_DIM_SIZE][8]             = {0};
-	double  scales[MAX_DIM_SIZE]              = {0.0};
+	double  scales[MAX_DIM_SIZE]              = {1.0};
```
tmontaigu, line 128: {1.0, 1.0, 1.0} otherwise, only scales[0] is 1.0.

Also, to be pedantic, the standard says that 
> Any unused "no_data", "min","max", "scale", or "offset" fields must be set to zero.

Although I don't think not respecting this would break other reading software/libs, if we want to be conforming we need to add a `if scaleIsRelevant() { ... } else { }` in the function that write extra scalar filed vlr

dgirardeau: That's a good point!

p-leroy: Sorry, I was wrong with this scale thing. I made a mistake during some of my tests. So, this suggestion is not appropriate. But the fwfDataOffset suggestion of correction should be OK.

tmontaigu (approved):

dgirardeau (approved):

dgirardeau:
Thanks for the cleanup!

PR #1852, opened 2023-09-14
New option "Toggle draw normals as lines" in the contextual menu of a cloud

p-leroy:
This is a little tool to display cloud normals as lines.
A widget allows to set the color and the length of the normals.

What do you think about it?


dgirardeau (changes requested):
I think it's possible to make it work, but more work is required 👼

libs/qCC_db/include/ccPointCloud.h
```diff
@@ -362,6 +363,27 @@ class QCC_DB_LIB_API ccPointCloud : public CCCoreLib::PointCloudTpl<ccGenericPoi
...
+	void setNormalLength(float value);
+
+	//! Set the length of the normals
+	void setNormalColor(QColor color);
+
+	//! Do the drawing of normals
+	bool drawNormals(CC_DRAW_CONTEXT &context);
+
+	//! When normals are drawn, a widget is open to set somme parameters
+	void setDrawNormalsWidget(QSharedPointer<QWidget> &&widget);
```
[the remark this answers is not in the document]
p-leroy: This member function is obsolete (not used anywhere... and not implemented in the cpp!). At the beginning I have used it but I dropped it when I decided to remove all references to the GUI. Sorry.

libs/qCC_db/include/ccPointCloud.h
```diff
@@ -766,6 +788,7 @@ class QCC_DB_LIB_API ccPointCloud : public CCCoreLib::PointCloudTpl<ccGenericPoi
 
 	//! Normals (compressed)
 	NormsIndexesTableType* m_normals;
+	std::vector<CCVector3> m_decompressed_normals; // used for drawing normals if needed
```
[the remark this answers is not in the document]
p-leroy: If the normals are modified, normalsHaveChanged() calls updateDecompressedNormals() in ccPoinCloud.h

libs/qCC_db/src/ccPointCloud.cpp
```diff
@@ -5748,6 +5754,140 @@ bool ccPointCloud::orientNormalsWithFM(	unsigned char level,
...
+	glFunc->glGetFloatv(GL_PROJECTION_MATRIX, projection.data());
+	glFunc->glGetFloatv(GL_MODELVIEW_MATRIX, modelView.data());
+	QMatrix4x4 projectionModelView = projection * modelView; // QMatrix4x4 expects row major data
+
+	if (m_programDrawNormals.isNull())
+	{
+		m_programDrawNormals = QSharedPointer<QOpenGLShaderProgram>(new QOpenGLShaderProgram(context.qGLContext));
+
+		// create vertex shader
+		QString vertexShaderFile = ("C:/dev/CloudCompare/qCC/shaders/DrawNormals/DrawNormals.vs");
```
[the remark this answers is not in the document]
p-leroy: Argl, hardcoded path!
Right, a "shaders" directory already exists. I have modified a CMakeLists.txt to have a copy of my shaders in the shaders directory and I have used a static variable to get the path.

libs/qCC_db/src/ccPointCloud.cpp
```diff
@@ -5748,6 +5754,140 @@ bool ccPointCloud::orientNormalsWithFM(	unsigned char level,
...
+	// if the normals are drawn and they have changed, we need to update the array
+	if (normalsAreDrawn())
+	{
+		// we need to decompress the normals
+		m_decompressed_normals.clear();
+		m_decompressed_normals.resize(size());
+		for (unsigned idx = 0; idx < size(); idx++)
+		{
+			const CCVector3 N = getPointNormal(idx);
+			m_decompressed_normals[idx].x = N.x;
```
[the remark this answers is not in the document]
p-leroy: Far better!

qCC/ccDrawNormalsWidget.cpp
```diff
@@ -0,0 +1,109 @@
+#include "ccDrawNormalsWidget.h"
+#include "ui_ccDrawNormalsWidget.h"
+
+#include <QSettings>
+
+ccDrawNormalsWidget::ccDrawNormalsWidget(ccPointCloud *cloud, QWidget *parent) :
+	QWidget(parent),
+	ui(new Ui::ccDrawNormalsWidget),
+	m_cloud(cloud)
```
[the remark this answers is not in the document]
p-leroy: Sorry, I will take care of the coding style.

p-leroy:
Thank you a lot for the review Daniel!

libs/qCC_db/include/ccPointCloud.h
```diff
@@ -364,24 +364,21 @@ class QCC_DB_LIB_API ccPointCloud : public CCCoreLib::PointCloudTpl<ccGenericPoi
 									ccProgressDialog* pDlg = nullptr);
 
 	//! Toggle the drawing of normals as small lines
-	bool setNormalsAreDrawn(bool state);
+	void showNormalsAsLines(bool state);
 
 	//! Are normals drawn as vectors?
-	bool normalsAreDrawn();
+	const bool normalsAreDrawn();
```
dgirardeau, line 370: Nope, the method should be const ;)

bool normalsAreDrawn() const;

libs/qCC_db/include/ccPointCloud.h
```diff
@@ -364,24 +364,21 @@ class QCC_DB_LIB_API ccPointCloud : public CCCoreLib::PointCloudTpl<ccGenericPoi
...
 
 	//! Set the length of the normals
 	void setNormalLength(float value);
 
-	//! Set the length of the normals
-	void setNormalColor(QColor color);
+	//! Set the color of the normals
+	void setNormalLineColor(ccColor::Rgba color);
```
dgirardeau, line 376: void setNormalLineColor(const ccColor::Rgba& color);

libs/qCC_db/include/ccPointCloud.h
```diff
@@ -900,18 +902,21 @@ class QCC_DB_LIB_API ccPointCloud : public CCCoreLib::PointCloudTpl<ccGenericPoi
...
 	SharedFWFDataContainer m_fwfData;
 
-	bool m_normalsAreDrawn;
-	float m_normalLength;
-	QColor m_normalColor;
+	bool m_normalsDrawnAsLines;
+
+	struct normalLineParameters
+	{
+		float length = 1.;
```
dgirardeau, line 909: float length = 1.0f;

libs/qCC_db/include/ccPointCloud.h
```diff
@@ -900,18 +902,21 @@ class QCC_DB_LIB_API ccPointCloud : public CCCoreLib::PointCloudTpl<ccGenericPoi
 	//! Waveforms raw data storage
 	SharedFWFDataContainer m_fwfData;
 
-	bool m_normalsAreDrawn;
-	float m_normalLength;
-	QColor m_normalColor;
+	bool m_normalsDrawnAsLines;
+
+	struct normalLineParameters
```
dgirardeau, line 907: NormalLineParameters (classes or strucs start with a capital letter)

libs/qCC_db/src/ccPointCloud.cpp
```diff
@@ -59,6 +59,14 @@
 
 static const char s_deviationSFName[] = "Deviation";
 
+// Shader path
+static QString s_shaderPath;
```
dgirardeau, line 63: Does it need to be in the header? Can it move to the cpp (with the implementation of the method below)

libs/qCC_db/src/ccPointCloud.cpp
```diff
@@ -5803,36 +5803,40 @@ bool ccPointCloud::drawNormals(CC_DRAW_CONTEXT& context)
 
 	if (m_programDrawNormals.isNull())
 	{
-		m_programDrawNormals = QSharedPointer<QOpenGLShaderProgram>(new QOpenGLShaderProgram(context.qGLContext));
+		m_programDrawNormals.reset(new QOpenGLShaderProgram(context.qGLContext));
```
dgirardeau, line 5806: So each point cloud has its own program? Can't you share one for all the clouds?

libs/qCC_db/src/ccPointCloud.cpp
```diff
@@ -5803,36 +5803,40 @@ bool ccPointCloud::drawNormals(CC_DRAW_CONTEXT& context)
...
 		}
 
 		// create fragment shader
-		QString fragmentShaderFile = ("C:/dev/CloudCompare/qCC/shaders/DrawNormals/DrawNormals.fs");
+		QString fragmentShaderFile(s_shaderPath + "/DrawNormals/DrawNormals.fs");
 		if (!m_programDrawNormals->addShaderFromSourceFile(QOpenGLShader::Fragment, fragmentShaderFile))
 		{
 			error = m_programDrawNormals->log();
 			ccLog::Error(error);
+			return false;;
```
dgirardeau, line 5832: spurious semicolon

qCC/ccDrawNormalsWidget.cpp
```diff
@@ -45,65 +45,76 @@ ccDrawNormalsWidget::~ccDrawNormalsWidget()
...
+		m_cloud->setNormalLineColor(ccColor::green);
+	else if (this->ui->pushButtonBlue->isChecked())
+		m_cloud->setNormalLineColor(ccColor::blue);
+	else if (this->ui->pushButtonBlack->isChecked())
+		m_cloud->setNormalLineColor(ccColor::black);
 	else
-		m_cloud->setNormalColor(QColorConstants::Yellow);
+		m_cloud->setNormalLineColor(ccColor::yellow);
+
+	if (m_cloud->getDisplay())
```
dgirardeau, line 118: Ah, if you use redrawDisplay, then it's safe to call it directly (no need to test 'getDisplay', I believe it's already done by redrawDisplay)

qCC/db_tree/ccDBRoot.cpp
```diff
@@ -2202,16 +2202,20 @@ void ccDBRoot::drawNormals()
...
 			else
 			{
 				if (cloud->hasNormals())
 				{
-					openDrawNormalsWidget(cloud);
+					openPointCloudNormalsDisplayParametersWidget(cloud);
 					QApplication::processEvents();
-					cloud->setNormalsAreDrawn(true);
+					cloud->showNormalsAsLines(true);
+					if (cloud->getDisplay())
```
dgirardeau, line 2217: same comment as above

qCC/db_tree/ccDBRoot.cpp
```diff
@@ -2378,23 +2382,14 @@ void ccDBRoot::showContextMenu(const QPoint& menuPos)
...
 
-void ccDBRoot::openDrawNormalsWidget(ccPointCloud *cloud)
+void ccDBRoot::openPointCloudNormalsDisplayParametersWidget(ccPointCloud *cloud)
 {
-	map_cloud_drawNormalsWidget[cloud] = new ccDrawNormalsWidget(cloud, MainWindow::TheInstance());
+	m_normalLineParametersWidget.reset(new ccDrawNormalsWidget(cloud, MainWindow::TheInstance()));
 }
 
-void ccDBRoot::closeDrawNormalsWidget(ccPointCloud *cloud)
+void ccDBRoot::closePointCloudNormalsDisplayParametersWidget(ccPointCloud *cloud)
```
dgirardeau, line 2390: I guess you don't need the cloud as input anymore?

dgirardeau (changes requested):
Sorry for the delay, I was busy this week. Here is my second review!

qCC/ccDrawNormalsWidget.cpp
p-leroy: Yes it is!

libs/qCC_db/src/ccPointCloud.cpp
p-leroy: It is already in the cpp. Did I miss something?

libs/qCC_db/src/ccPointCloud.cpp
p-leroy: Not sure about my changes here as the IDE is now complaining about "non-POD static". What do you think about it?

libs/qCC_db/src/ccPointCloud.cpp
dgirardeau: No, apparently I read too quickly!

dgirardeau:
Hi again Paul, I had the time to test the actual feature.

First, I rebased your branch and made a few fixes (see the 3 additional commits).

Now, I like the feature, but I wonder why it needs a separate dialog? This dialog remains 'always on top', even if CC is minimized, and even when one tries to close CC (so it can be in front of the small confirmation dialog that asks the user whether CC should be closed or not, preventing the user to click on it ;) ). Last, the only way to toggle this feature is via the right click (which is not the most straightforward).

Why not simply integrating that in the cloud properties? Just as the scalar field section, with some fields, combo boxes, etc. (the choice of color could be in a combo box to make things simpler).

p-leroy:
Hi Daniel,
Thank you for your work! Good idea to integrate this tool in the cloud properties. I will look at that asap.

dgirardeau:
Very nice. On my side, when I select a cloud, the 'Draw' checkbox is not checked, even if the normals are drawn as lines.

p-leroy:
It should work now!

dgirardeau (approved):

dgirardeau:
Thanks for this nice addition!

p-leroy:
Thank you for your careful reviews!

Issue #1871, opened 2023-10-18
qMeshBoolean compilation failed


Loris-L:
Hi, i tried to compile cloudcompare 2.13.beta on my ubuntu 22.04.3 LTS from master branch. I followed the BUILD.md and, after i've installed dependencies requested, i compile cloudcompare with cmake 3.22.1. The cloudcompare compilation has failed because qMeshBoolean plugin compilation has failed, and this is the error code:

```
home/loris/Scaricati/CloudCompare/plugins/core/Standard/qMeshBoolean/src/qMeshBoolean.cpp:53:9: error: ‘Eigen’ does not name a type
   53 |         Eigen::MatrixXd V; //!< Vertices
      |         ^
/home/loris/Scaricati/CloudCompare/plugins/core/Standard/qMeshBoolean/src/qMeshBoolean.cpp:54:9: error: ‘Eigen’ does not name a type
   54 |         Eigen::MatrixXi F; //!< Triangles
      |         ^
/home/loris/Scaricati/CloudCompare/plugins/core/Standard/qMeshBoolean/src/qMeshBoolean.cpp: In function ‘bool ToIGLMesh(const ccMesh*, IGLMesh&, ccMainAppInterface*)’:
/home/loris/Scaricati/CloudCompare/plugins/core/Standard/qMeshBoolean/src/qMeshBoolean.cpp:109:21: error: ‘struct IGLMesh’ has no member named ‘F’
  109 |                 out.F.resize(triCount, 3);
      |                     ^
/home/loris/Scaricati/CloudCompare/plugins/core/Standard/qMeshBoolean/src/qMeshBoolean.cpp:110:21: error: ‘struct IGLMesh’ has no member named ‘V’
  110 |                 out.V.resize(vertCount, 3);
      |                     ^
/home/loris/Scaricati/CloudCompare/plugins/core/Standard/qMeshBoolean/src/qMeshBoolean.cpp:124:29: error: ‘struct IGLMesh’ has no member named ‘F’
  124 |                         out.F(i, 0) = tsi->i1;
      |                             ^
/home/loris/Scaricati/CloudCompare/plugins/core/Standard/qMeshBoolean/src/qMeshBoolean.cpp:125:29: error: ‘struct IGLMesh’ has no member named ‘F’
  125 |                         out.F(i, 1) = tsi->i2;
      |                             ^
[...]

PR #1872, opened 2023-10-21
Updated qMeshBoolean.cpp to prevent compilation failure

Loris-L:
After the qMeshBoolean plugin compilation failure (see #1871), i changed the mesh_boolean.h path position in qMeshBolean.cpp and then the compilation finished without errors, CloudCompare works. It was probably hidden by ifdef-endif code part.

### DIFF
--- a/plugins/core/Standard/qMeshBoolean/src/qMeshBoolean.cpp
+++ b/plugins/core/Standard/qMeshBoolean/src/qMeshBoolean.cpp
@@ -39,12 +39,12 @@
 #endif
 
 //libIGL
+#include <igl/copyleft/cgal/mesh_boolean.h>
 #ifdef _MSC_VER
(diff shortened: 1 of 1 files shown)


plugins/core/Standard/qMeshBoolean/src/qMeshBoolean.cpp
```diff
@@ -39,12 +39,12 @@
 #endif
 
 //libIGL
+#include <igl/copyleft/cgal/mesh_boolean.h>
```
dgirardeau, line 42: To respect the spirit of the warnings disabling lines below, you should move the #include AFTER the #ifdef _MSC_VER section (otherwise the #pragma warning calls will be useless).

dgirardeau (changes requested):

dgirardeau (approved):

PR #1874, opened 2023-10-23
Cmd option SF_ADD_ID with AS_INT supboption

p-leroy:
SF_ADD_ID allows to add the index of each point as a scalar field, initially as a float 32.
Possibility to store the index as a raw int32 to avoid a loss of accuracy for large integers and store values from 0 up to 2**32.

### DIFF
--- a/qCC/ccEntityAction.cpp
+++ b/qCC/ccEntityAction.cpp
@@ -1118,6 +1118,7 @@ namespace ccEntityAction
 					else
 					{
 #if defined CC_CORE_LIB_USES_DOUBLE
+#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
 						unsigned char *valuePtr = (unsigned char *)(&idValue);
 						valuePtr[7] = static_cast<unsigned char>(j & 0xff);
 						valuePtr[6] = static_cast<unsigned char>(j >> 8 & 0xff);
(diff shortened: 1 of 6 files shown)


qCC/ccEntityAction.cpp
```diff
@@ -1107,11 +1107,39 @@ namespace ccEntityAction
...
+
+				for (size_t j=0 ; j<cloud->size(); j++)
 				{
-					ScalarType idValue = static_cast<ScalarType>(j);
-					sf->setValue(j, idValue);
+					ScalarType idValue;
+					if (false)
+					{
+						idValue = static_cast<ScalarType>(j);
+						sf->setValue(j, idValue);
```
dgirardeau, line 1117: I guess you can call this only once (after the if then else statement)

qCC/ccEntityAction.cpp
```diff
@@ -1107,11 +1107,39 @@ namespace ccEntityAction
...
 				Q_ASSERT(sf->currentSize() == pc->size());
-				
-				for (unsigned j=0 ; j<cloud->size(); j++)
+
+				for (size_t j=0 ; j<cloud->size(); j++)
 				{
-					ScalarType idValue = static_cast<ScalarType>(j);
-					sf->setValue(j, idValue);
+					ScalarType idValue;
+					if (false)
```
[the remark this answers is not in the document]
p-leroy: Sure, you're right

qCC/ccEntityAction.cpp
p-leroy: Yes

dgirardeau:
What is the exact use case of this export as an int? (as it will give strange results within CloudCompare). I am mostly wondering about systems with big endianness. Should it be properly handled here?

p-leroy:
My use case is the following: 

- I use CloudCompare in command line to extract points in a laz file (with waveforms as wdp) which are close to a another classified point cloud.
- In this selection, for each point, I want to go back to the original las file and get all echoes belonging to the same laser pulse (I do that in Python). As there is a loss of accuracy when recording the GPS time, I have found this index trick to be able to go back in the original file, get the GPS time at full resolution, and so get all other echoes belonging to the same shot, if any.
- You are right that this "index" should not be used for display in CC

Big endianness should be handled properly, you are right. Do you compile CC for big endian architectures?

dgirardeau:
Ok understood. I don't compile it myself, but I know some people do...

dgirardeau (approved):

dgirardeau:
Thanks again ;)

PR #1876, opened 2023-10-28
Added preference to use GLVND libraries for OpenGL and updated draco cmake file

Loris-L:
These modifies are related to Linux OS.
I added the preference to use GLVND libraries for OpenGL rather than Legacy library that it's commonly used by cmake 3.10 and below.
I also updated the draco cmake file to set include and library directorories, because in Linux OS, after CC compilation, libQDRACO_IO_PLUGIN isn't loaded due to missing link with draco libraries.


rjanvier:
Thank you Loris. I think it won’t hurt 🙂.

rjanvier (approved):

PR #1880, opened 2023-11-06
New option for the "Segment" tool: "Clone selection".

p-leroy:


dgirardeau:
Not fully sure what this option does? It exports the subset of visible points as a new cloud? (and only the vertices if a mesh is currently segmented?). If that's the case, I guess the meshes should be properly handled, and I'm not sure I would call that 'clone'.

p-leroy:
I use this button when dealing with point clouds. I often want to inspect a specific area and it saves me from having to merge back the remaining/segmented parts after inspection.

dgirardeau:
So can you simply call it 'export'? And should you disable it if meshes are selected? (else you'll have to properly manage the mesh export)

qCC/ccGraphicalSegmentationTool.cpp
```diff
@@ -77,13 +80,14 @@ ccGraphicalSegmentationTool::ccGraphicalSegmentationTool(QWidget* parent)
 
 	connect(inButton,				&QToolButton::clicked, this, &ccGraphicalSegmentationTool::segmentIn);
 	connect(outButton,				&QToolButton::clicked, this, &ccGraphicalSegmentationTool::segmentOut);
+	connect(cloneSelectionButton,	&QToolButton::clicked, this, &ccGraphicalSegmentationTool::cloneSelection);
```
dgirardeau:
```suggestion
	connect(exportSelectionButton,	&QToolButton::clicked, this, &ccGraphicalSegmentationTool::exportSelection);
```

qCC/ccGraphicalSegmentationTool.cpp
```diff
@@ -828,7 +850,12 @@ void ccGraphicalSegmentationTool::segmentOut()
 	segment(false);
 }
 
-void ccGraphicalSegmentationTool::segment(bool keepPointsInside, ScalarType classificationValue/*=CCCoreLib::NAN_VALUE*/)
+void ccGraphicalSegmentationTool::cloneSelection()
+{
+	segment(true, CCCoreLib::NAN_VALUE, true);
+}
+
+void ccGraphicalSegmentationTool::segment(bool keepPointsInside, ScalarType classificationValue/*=CCCoreLib::NAN_VALUE*/, bool cloneSelection/*=false*/)
```
dgirardeau, line 863:
```suggestion
void ccGraphicalSegmentationTool::segment(bool keepPointsInside, ScalarType classificationValue/*=CCCoreLib::NAN_VALUE*/, bool exportSelection/*=false*/)
```

qCC/ccGraphicalSegmentationTool.cpp
```diff
@@ -828,7 +850,12 @@ void ccGraphicalSegmentationTool::segmentOut()
 	segment(false);
 }
 
-void ccGraphicalSegmentationTool::segment(bool keepPointsInside, ScalarType classificationValue/*=CCCoreLib::NAN_VALUE*/)
+void ccGraphicalSegmentationTool::cloneSelection()
```
dgirardeau:
```suggestion
void ccGraphicalSegmentationTool::exportSelection()
```

qCC/ccGraphicalSegmentationTool.cpp
```diff
@@ -886,6 +913,7 @@ void ccGraphicalSegmentationTool::segment(bool keepPointsInside, ScalarType clas
 	int errorCount = 0;
 	for (QSet<ccHObject *>::const_iterator p = m_toSegment.constBegin(); p != m_toSegment.constEnd(); ++p)
 	{
+		ccHObject* entity = (*p);
```
dgirardeau, line 921: is it used?

qCC/ccGraphicalSegmentationTool.cpp
```diff
@@ -908,7 +942,7 @@ void ccGraphicalSegmentationTool::segment(bool keepPointsInside, ScalarType clas
 			ccPointCloud* pc = ccHObjectCaster::ToPointCloud(*p);
 			if (!pc)
 			{
-				ccLog::Warning("Can't apply classification to cloud " + (*p)->getName());
+				ccLog::Warning("Can't apply classification to object " + (*p)->getName());
```
dgirardeau, line 950:
```suggestion
				ccLog::Warning("Can't apply classification to entity" + (*p)->getName());
```

qCC/ccGraphicalSegmentationTool.cpp
```diff
@@ -931,11 +965,11 @@ void ccGraphicalSegmentationTool::segment(bool keepPointsInside, ScalarType clas
 
 		// we project each point and we check if it falls inside the segmentation polyline
 #if defined(_OPENMP)
-        omp_set_num_threads(omp_get_max_threads());
+		omp_set_num_threads(omp_get_max_threads());
 		#pragma omp parallel for
 #endif
 		for (int i = 0; i < cloudSize; ++i)
-		{
+		{
```
dgirardeau, line 977: restore to its original state

qCC/ccGraphicalSegmentationTool.cpp
```diff
@@ -931,11 +965,11 @@ void ccGraphicalSegmentationTool::segment(bool keepPointsInside, ScalarType clas
 
 		// we project each point and we check if it falls inside the segmentation polyline
 #if defined(_OPENMP)
-        omp_set_num_threads(omp_get_max_threads());
+		omp_set_num_threads(omp_get_max_threads());
```
dgirardeau, line 973: Not sure what is this change? To be restored?

qCC/ccGraphicalSegmentationTool.cpp
```diff
@@ -953,18 +987,36 @@ void ccGraphicalSegmentationTool::segment(bool keepPointsInside, ScalarType clas
...
 				}
-				else
+				else // standard segmentation mode
 				{
-					// standard segmentation mode
-					visibilityArray[i] = (keepPointsInside != pointInside ? CCCoreLib::POINT_HIDDEN : CCCoreLib::POINT_VISIBLE);
+					unsigned char visibility = (keepPointsInside != pointInside ? CCCoreLib::POINT_HIDDEN : CCCoreLib::POINT_VISIBLE);
+					visibilityArray[i] = visibility;
+
+					if (cloneSelection) // specify an output visibility array for the export mode (exported points will be hidden until the Segment tool is closed)
```
dgirardeau, line 1007: I don't understand why you need to use the 'segment' method when exporting the current selection? (this the visibility array is already initialized on each entity that needs one, and you can query it with 'getVisibilityArray'.

Why don't you simply add a dedicated 'export' method that goes through the entities and calls createNewCloudFromVisibilitySelection or createNewMeshFromSelection?

qCC/ccGraphicalSegmentationTool.cpp
```diff
@@ -987,9 +1094,9 @@ void ccGraphicalSegmentationTool::segment(bool keepPointsInside, ScalarType clas
 		}
 	}
 
-	if (classificationMode)
+	if (classificationMode || cloneSelection)
```
dgirardeau, line 1102: replace 'clone' by 'export' everywhere

qCC/ui_templates/graphicalSegmentationDlg.ui
```diff
@@ -147,6 +147,20 @@
         </property>
        </widget>
       </item>
+      <item>
+       <widget class="QToolButton" name="cloneSelectionButton">
```
dgirardeau, line 151:
```suggestion
       <widget class="QToolButton" name="exportSelectionButton">
```

qCC/ccGraphicalSegmentationTool.cpp
```diff
@@ -60,6 +60,9 @@
 #include <omp.h>
 #endif
 
+std::set<ccHObject*> enableAtTheClose;
```
dgirardeau, line 63: Are those static?! Consider using members of the tool instance instead.

And rename them m_enableOnClose and m_disableOnClose

qCC/ccGraphicalSegmentationTool.cpp
```diff
@@ -828,7 +850,12 @@ void ccGraphicalSegmentationTool::segmentOut()
 	segment(false);
 }
 
-void ccGraphicalSegmentationTool::segment(bool keepPointsInside, ScalarType classificationValue/*=CCCoreLib::NAN_VALUE*/)
+void ccGraphicalSegmentationTool::cloneSelection()
+{
+	segment(true, CCCoreLib::NAN_VALUE, true);
```
dgirardeau, line 860: see my other comment on the usage of 'segment'

dgirardeau (changes requested):

qCC/ccGraphicalSegmentationTool.cpp
p-leroy: now it is used!

qCC/ccGraphicalSegmentationTool.cpp
p-leroy: oups

qCC/ccGraphicalSegmentationTool.cpp
p-leroy: sorry, this was a wrong indentation

qCC/ccGraphicalSegmentationTool.cpp
p-leroy: All the machinery with the omp stuff of the selection tool is in the segment method, I did not want to duplicate too much code. It is right that the code is a little bit complicated. What is your thoughts?

qCC/ccGraphicalSegmentationTool.cpp
```diff
@@ -344,23 +341,23 @@ void ccGraphicalSegmentationTool::stop(bool accepted)
 
 	ccOverlayDialog::stop(accepted);
 
-	for (auto item : enableAtTheClose) // in export mode, all parts are enabled at the close
+	for (auto item : m_enableAtTheClose) // in export mode, all parts are enabled at the close
```
dgirardeau, line 344:
```suggestion
	for (auto item : m_enableOnClose) // in export mode, all parts are enabled at the close
```

dgirardeau: Same everywhere

qCC/ccGraphicalSegmentationTool.cpp
```diff
@@ -344,23 +341,23 @@ void ccGraphicalSegmentationTool::stop(bool accepted)
...
 		if (item != nullptr)
 		{
 			item->setEnabled(true);
 		}
 	}
-	enableAtTheClose.clear();
+	m_enableAtTheClose.clear();
 
-	for (auto item : disableAtTheClose) // in export mode, the original entities are disabled at the close to show the parts
+	for (auto item : m_disableAtTheClose) // in export mode, the original entities are disabled at the close to show the parts
```
dgirardeau, line 353:
```suggestion
	for (auto item : m_disableOnClose) // in export mode, the original entities are disabled on close to make sure the newly created parts are visible
```

qCC/ccGraphicalSegmentationTool.cpp
```diff
@@ -344,23 +341,23 @@ void ccGraphicalSegmentationTool::stop(bool accepted)
...
 	{
 		if (item != nullptr)
 		{
 			item->setEnabled(false);
 		}
 	}
-	disableAtTheClose.clear();
+	m_disableAtTheClose.clear();
```
dgirardeau, line 360:
```suggestion
	m_disableOnClose.clear();
```

dgirardeau: same everywhere

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ v2.13.alpha (???) - (??/??/????)
 ----------------------
 - - New features:
 
+	- Segment
+		- new button in the toolbar to export the current selection
```
dgirardeau, line 9:
```suggestion
		- new button in the toolbar to export the current selection as a new cloud or mesh (without needing to close the tool)
```

qCC/ccGraphicalSegmentationTool.h
```diff
@@ -148,6 +151,12 @@ class ccGraphicalSegmentationTool : public ccOverlayDialog, public Ui::Graphical
 
 	//! Whether to delete hidden parts after segmentation
 	bool m_deleteHiddenParts;
+
+	//! In export mode, it switches the visibility of the exported parts
```
dgirardeau, line 155:
```suggestion
	//! In export mode, entities in this set will be enabled/visible
```

qCC/ccGraphicalSegmentationTool.h
```diff
@@ -148,6 +151,12 @@ class ccGraphicalSegmentationTool : public ccOverlayDialog, public Ui::Graphical
...
+	std::set<ccHObject*> m_enableAtTheClose;
+
+	//! In export mode, it switches off the original cloud to show the exported parts
```
dgirardeau, line 158:
```suggestion
	//! In export mode, entities in this set will be disabled/invisible
```

qCC/ccGraphicalSegmentationTool.h
```diff
@@ -148,6 +151,12 @@ class ccGraphicalSegmentationTool : public ccOverlayDialog, public Ui::Graphical
 
 	//! Whether to delete hidden parts after segmentation
 	bool m_deleteHiddenParts;
+
+	//! In export mode, it switches the visibility of the exported parts
+	std::set<ccHObject*> m_enableAtTheClose;
```
dgirardeau, line 156:
```suggestion
	std::set<ccHObject*> m_enableOnClose;
```

qCC/ccGraphicalSegmentationTool.h
```diff
@@ -148,6 +151,12 @@ class ccGraphicalSegmentationTool : public ccOverlayDialog, public Ui::Graphical
...
+
+	//! In export mode, it switches off the original cloud to show the exported parts
+	std::set<ccHObject*> m_disableAtTheClose;
```
dgirardeau, line 159:
```suggestion
	std::set<ccHObject*> m_disableOnClose;
```

dgirardeau (changes requested):

qCC/ccGraphicalSegmentationTool.cpp
dgirardeau: So my understanding is that you use this button to export the current selection. So I was expecting the visibility array to be already up to date (no need to call the 'segment' method again. If the selection is not done, which means the user has not chosen whether the points should be inside or outside the polygon, I would find the workflow a bit clunky.

So for me, the visibility array should be already set properly, and you don't need to call 'segment' again. Am I missing something?

dgirardeau:
Sorry for the delay, I was travelling the last 2 weeks.

qCC/ccGraphicalSegmentationTool.cpp
p-leroy: In my use case, I want to export several small patches and I click on the export button each time I close a poly-line. So I activate the segment tool each time. What is practical is that I can see which patches have been exported as they are removed from the screen and I can also pause the segmentation to move my cloud. 
It is very practical when I want to select patches for manual classification. Once I have all my patches, I quit the segmentation tool and my patches are displayed (thanks to m_enableOnClose, m_disableOnClose).

dgirardeau: How do you choose whether to export inside or outside then?

My general comment is that the way you use this tool is not clearly expressed by a simple 'export/clone' button.

p-leroy:
Your more that welcome, thank you for your careful review and for the time to spent on it.

qCC/ccGraphicalSegmentationTool.cpp
p-leroy: I thought that the use of the dolly sheep was sufficiently meaningful to say that you will duplicate the selection (inside). The original segmentation tool allows you to split into several parts, to remove points (segment out), but not to copy the points you selected without modifying the initial cloud.
Maybe something like copy/paste could be more clear?

dgirardeau: So my issue is that it's really counter intuitive (for me) that this option applies to what's inside the polygon only but not yet segmented. It's almost a different tool in its spirit. The '#' icon (classification) was already a bit weird in this sense (even though the fact that the points are not disappearing makes things clearer).

I think we could compromise with an icon to the right of the '#' icon (instead of to the left), and that would look like this:
export

And in the code, don't lump the standard segmentation mode and the export mode together, but make them 2 clearly separated 'else' cases. See my other comment.

dgirardeau (changes requested):
Second pass

qCC/ccGraphicalSegmentationTool.cpp
```diff
@@ -953,18 +989,36 @@ void ccGraphicalSegmentationTool::segment(bool keepPointsInside, ScalarType clas
...
+				if (classifSF) // classification mode
 				{
-					// classification mode
 					if (pointInside)
 					{
 						classifSF->setValue(i, classificationValue);
 					}
 				}
-				else
+				else // standard segmentation mode + export mode
```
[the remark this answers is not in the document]
dgirardeau: I tried to simply things at the maximum, assuming keepPointsInside is always true in this mode.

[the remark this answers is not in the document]
p-leroy: Amazing!

dgirardeau (approved):

PR #1888, opened 2023-11-24
LasSaveDialog fix initial state of normals check box

### LINKED ISSUE #1887: CloudCompare problem with last updated feature for LAS files
This time i've opened an issue for the last windows installer available on CC site: i've also windows 11 in my PC and with the last CC, when i try to open las files, CC crashes after click on a button of the first window about the setting of scalar fields. Instead, when i try to save e.g. a bin file in las file, CC isn't able to save it with the error described in the next image.

I saw the last @tmontaigu 's PR, maybe is it something related to that?

tmontaigu:
The checkbox was initialized to checked even when the cloud had no normals, leading to the saver trying to save non existant normals

Fixes #1887


dgirardeau:
Don't you miss some defensive code in the 'saver' function to avoid the crash as well? (even with this fix)

dgirardeau (approved):

dgirardeau:
Thanks

PR #1890, opened 2023-11-24
Fix Loading files without normals

tmontaigu:


dgirardeau:
Ah! I pushed my fix already. Can you rebase?

tmontaigu:
It fixes the same bug but in a different way so I don't think its usefull to rebase + merge

dgirardeau:
I like the // safeguard part though ;)

dgirardeau:
Ah, I tried the online conflict solver of github, and it seems it didn't handled well spacings 😅

dgirardeau:
Ok, the history was also weird with this merge commit... Should be fixed now.

tmontaigu:
Yep its looks ok

PR #1903, opened 2023-12-09
Create a Command Line command RENAME_CLOUDS

HeadLessHUN:
RENAME_CLOUDS "new_name", command will rename clouds and meshes, to "new_name" and if there are multiple entities then it gives an increasing suffix padded to 3 digit. cc forum reference https://www.danielgm.net/cc/forum/viewtopic.php?f=14&t=6697&sid=545a7ceae939eb9fb09d6115f195c8b1&p=28692#p28692

It will produce a log something like this.
```
[RENAME CLOUDS]
Cloud 'pointcloud1 - Cloud' renamed to 'newBaseName-001'
Mesh 'Mesh' renamed to 'newBaseName-002'
Mesh 'Mesh' renamed to 'newBaseName-003'
```

if the name contains invalid chars it will thrown an error, and exit.
```
[RENAME CLOUDS]
Name cannot contain any of these characters: :/\*?"\|<>
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,9 @@ v2.13.alpha (???) - (??/??/????)
 		- DEBUG
 			- display various information to help one debug and tweak the command line
 			- can be placed at difference positions
+		- RENAME_ENTITIES {base name}
+			- rename all loaded entities (clouds or meshes) with the provided base name.
+				A numerical suffix is added if multiple entities are loaded.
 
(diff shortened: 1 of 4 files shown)


dgirardeau:
Very good, this kind of contributions help a lot!

I renamed the option 'RENAME_ENTITIES' since it applies to both clouds and meshes now.

dgirardeau (approved):

Issue #1908, opened 2023-12-13
CLI APPLY transformation high coordinates


HeadLessHUN:
Description:
using command line -APPLY_TRANS function if the transformation coordinates are high enough it will loss precision. 

Expected:
if possible use global shift for the translation process.

**Additional context**

**Your environment**
 - CC Version c6eba51139a86eba8a1e414dfaec2deb118a6a3d
 - windows 10
 - AMD rx570
 - online beta

PR #1926, opened 2023-12-27
CLI Apply_trans use global shift

HeadLessHUN:
#1908
- -APPLY_TRANS uses (default) global shift to maintain precision
- New sub-option for -APPLY_TRANS command: -NOT_APPLY_TO_GLOBAL it will not use global shift.

And a little correction for -SET_GLOBAL_SHIFT

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -267,7 +267,12 @@ v2.13.alpha (???) - (??/??/????)
 		- New sub-option for the -C2M_DIST command: -UNSIGNED, to compute unsigned distances
 		- New sub-option for the -SF_ARITHMETIC command: -IN_PLACE, to update the scalar field in place, without creating a new SF
 		- Most methods using scalar fields as input will now also accept the scalar field name (in lieu of the SF index)
-		- New sub-option for -APPLY_TRANS command: -INVERSE to inverse the transformation matrix before it is applied.
+		- Changes to the -APPLY_TRANS command:
+			- -APPLY_TRANS will now apply the transformation to the global coordinate system by default
+				Warning: the Global Shift may be automatically updated to maintain precision
+			- -NOT_APPLY_TO_GLOBAL: to force CC to apply the transformation to the local coordinate system
+				(i.e. after the Global Shift has been applied)
+			- -INVERSE: to inverse the transformation matrix before applying it.
 		- New sub-options for -SS command:
 			* -SS OCTREE NUMBER_OF_POINTS {number} to subsample with the highest octree level for which the resulting point count won't exceed the given number of points
 			* -SS OCTREE NUMBER_OF_POINTS PERCENT {number} to calculate NUMBER_OF_POINTS from PERCENT. PERCENT should be a decimal number between 0 and 100.
--- a/qCC/ccCommandLineCommands.cpp
+++ b/qCC/ccCommandLineCommands.cpp
@@ -2271,7 +2271,8 @@ bool CommandApplyTransformation::process(ccCommandLineInterface& cmd)
 			}
 		}
 
-		shiftedEntity->applyGLTransformation_recursive(&ccGLMatrix(transMat.data()));
+		ccGLMatrix matrixToApply = ccGLMatrix(transMat.data());
+		shiftedEntity->applyGLTransformation_recursive(&matrixToApply);
 		QString nameSuffix = "_TRANSFORMED";
 		if ((&desc)->getCLEntityType() == CL_ENTITY_TYPE::MESH)
 		{
--- a/qCC/ccCommandLineCommands.cpp
+++ b/qCC/ccCommandLineCommands.cpp
@@ -185,6 +185,7 @@ constexpr char OPTION_CLOUD[]							= "CLOUD";
 constexpr char OPTION_MESH[]							= "MESH";
 constexpr char OPTION_PERCENT[]							= "PERCENT";
 constexpr char OPTION_NUMBER_OF_POINTS[]				= "NUMBER_OF_POINTS";
+constexpr char OPTION_FORCE[]							= "FORCE";
 
 static void GetSFIndexOrName(ccCommandLineInterface& cmd, int& sfIndex, QString& sfName)
 {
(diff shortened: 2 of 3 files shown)


qCC/ccCommandLineCommands.cpp
```diff
@@ -2094,7 +2094,8 @@ bool CommandApplyTransformation::process(ccCommandLineInterface& cmd)
 			}
 		}
 
-		shiftedEntity->applyGLTransformation_recursive(&ccGLMatrix(transMat.data()));
+		ccGLMatrix matrixToApply = ccGLMatrix(transMat.data());
```
dgirardeau, line 2097:
```suggestion
		ccGLMatrix matrixToApply(transMat.data());
```

dgirardeau (changes requested):

dgirardeau (approved):

dgirardeau:
I rebased this PR and made the same syntax cleanup as to the other PR.

I'm still a bit confused by the change of behavior that this change implies to existing scripts. And also to the fact that the Global Shift is automatically updated without any user control.

Can't we switch the -NOT_APPLY_TO_GLOBAL sub-option to a -APPLY_TO_GLOBAL sub-option so that the user actively requests this behavior?

HeadLessHUN:
Well, i wanted to apply by default, because nobody wants a cloud with precision loss, but yes existing scripts that rely on similar global shift like merging. Could cause issue, but if you want to reverse, this behaviour we can change that.

dgirardeau:
It's probably less disruptive. And we could add a warning if we see that the output coordinates are too big (something like "Transformed coordinates are too large. Consider using '-APPLY_TO_GLOBAL' to prevent accuracy loss")

HeadLessHUN:
A warning might be good idea. Could you do it since i suppose you already did some changes. It is probably easier/faster for you right now.

dgirardeau:
I'll rebase it again ;)

Meanwhile, are you able to test my last changes on your side?

HeadLessHUN:
You removed use previous shift if possible? I think that part was a good addition, because if you have multiple clouds somewhat close together it would use the same shift, even if the clouds on the edge of "bestshift". All of that was cherrypicked from the gui version. Even if i was not a fan of skipping global shift if the local coordinates already "too high".

Today i don't really able to test things.

dgirardeau:
Well, it was not used? Or did I miss anything?

dgirardeau:
Oh my bad, I missed the loop! Let me fix that...

dgirardeau:
Should be restored now... I'll wait for you to test before merging this PR.

qCC/ccCommandLineCommands.cpp
```diff
@@ -2124,36 +2133,172 @@ bool CommandApplyTransformation::process(ccCommandLineInterface& cmd)
...
+			// we switch to a local transformation matrix
+			transMat.setTranslation(localTranslation);
+
+			//test if the translated cloud coordinates were already "too large"
+			ccBBox localBBox = shiftedEntity->getOwnBB();
+			CCVector3d Pl = localBBox.minCorner();
+			double Dl = localBBox.getDiagNormd();
+			if (!ccGlobalShiftManager::NeedShift(Pl) && !ccGlobalShiftManager::NeedRescale(Dl))
+			{
+				//test if the translated (local) cloud coordinates are too large
```
HeadLessHUN, line 2192: Cannot we skip this whole condition? Or at least create a -FORCE argument to pass through it?

If you say yes, i'll do it when i have opportunity probably next year.

HeadLessHUN (commented):
Cannot we skip this whole condition? Or at least create a -FORCE argument to pass through it?

If you say yes, i'll do it when i have opportunity jan 2nd/3rd

qCC/ccCommandLineCommands.cpp
dgirardeau: Yes why not

HeadLessHUN: skip or argument?:)

dgirardeau: Ah, so we can skip it, but in this case we should add a warning. Let me do it.

HeadLessHUN: it's all yours till january :)

dgirardeau: Ah but in this case the Global Shift will change in an uncontrolled way 😅 

Ok, so it's better to add an argument. I'll let you do it.

HeadLessHUN: and can you tell me why we check minCorner instead of maxCorner? probably we should check the highest absolute value, but usually that is the maxCorner since state grids never goes to negative territorry, i think.

dgirardeau: Well, it's just by consistency with the rest of the code, but it would make more sense in most cases indeed 👍

HeadLessHUN: ok, then happy new year. Probably i'm gonna be rush into real work tomorrow, we have a deadline on the first week. And i "wasted" (of course not)  my holidays with CC :) So probably i could take a look at this on the next half of the week.

HeadLessHUN: I've run some test. For me it is ready to merge

dgirardeau:
Ok I made 2 minor syntax updates, and I'll merge!

dgirardeau (approved):

CHANGELOG.md
```diff
@@ -267,7 +267,12 @@ v2.13.alpha (???) - (??/??/????)
 		- New sub-option for the -C2M_DIST command: -UNSIGNED, to compute unsigned distances
 		- New sub-option for the -SF_ARITHMETIC command: -IN_PLACE, to update the scalar field in place, without creating a new SF
 		- Most methods using scalar fields as input will now also accept the scalar field name (in lieu of the SF index)
-		- New sub-option for -APPLY_TRANS command: -INVERSE to inverse the transformation matrix before it is applied.
+		- Changes to the -APPLY_TRANS command:
+			- -APPLY_TRANS will now apply the transformation to the global coordinate system by default
+				Warning: the Global Shift may be automatically updated to maintain precision
+			- -NOT_APPLY_TO_GLOBAL: to force CC to apply the transformation to the local coordinate system
```
dgirardeau, line 273: Oops, I forgot to update this 😅 ... I'll fix that in a separate commit...

Issue #1929, opened 2024-01-02
Examples for command lines for testing and for the public


HeadLessHUN:
Hello Daniel.

I was thinking about to save COMMANDS_FILE for testing purposes and for making them public.

Can we save somewhere in the git repo these kind of COMMAND_FILEs? And few small clouds to test (I mean about 100-1000 points in a bin file, it can be enough to show how the most majority of commands works)

And i was thinking about to allow comment lines in command file, so it could be commented, whats done.

these example files would be 2 part first ( Prerequisites ) (Actual command/option usage.)

#TODO List
https://github.com/CloudCompare/CloudCompare/blob/master/qCC/TODO.txt

can we cross out things in the list or flag them as solved?

dgirardeau:
Yes, comments would definitely help (maybe with # and/or // as marker?)

Else, don't hesitate to clean/update the TODO list whenever you see something already implemented.

And last but not least, the example command scripts could fall either in a 'qCC/cli' or equivalent subfolder, or equivalently in a 'scripts/cli' subfolder.

HeadLessHUN:
I'm planning to to do 
# and // and it will affect the whole line after the comments, like in other languages.

And i was thinking about to make some kind of inline comments /* */, but don't want to reimplement the wheel and support the same kind of functionality what is done already. 

So maybe if someone wants to use inline comments they should do it inside quotes. Other wise it would be a little bit harder to implement.  But i'll see maybe i do inline/multiline comments available with /* */.

-COMMAND_TO_RUN '/*my comments goes here with spaces*/' -COMMAND_TO_RUN

PR #1930, opened 2024-01-03
Add support for comments inside a command file

HeadLessHUN:
syntax:
```
# whole line comment
// whole line comment
-SS OCTREE 10 //subsample cloud with octree level 10
 # this will run octree 10 rename_entities is commented out then save 
-SS OCTREE 10 '/* -RENAME_ENTITIES "rename" */' -SAVE_CLOUDS #we can use hashtags here as well
```

#1929

### DIFF
--- a/qCC/ccCommandLineCommands.cpp
+++ b/qCC/ccCommandLineCommands.cpp
@@ -720,7 +720,7 @@ bool CommandLoadCommandFile::process(ccCommandLineInterface& cmd)
 						processedArgs.append(arg);
 					}
 				}
-				if (insideSingleQuoteSection||insideDoubleQuoteSection)
+				if (insideSingleQuoteSection || insideDoubleQuoteSection)
 				{
 					// the single/double quote section was not closed...
 					cmd.warning("Probably malformed command (missing closing quote)");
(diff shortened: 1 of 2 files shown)


qCC/ccCommandLineCommands.cpp
```diff
@@ -730,12 +730,31 @@ bool CommandLoadCommandFile::process(ccCommandLineInterface& cmd)
...
+			{
 				QString processedArg = processedArgs.takeFirst();
-				if (!processedArg.isEmpty()) {
-					cmd.print(QObject::tr("\t[%1] %2").arg(insertingIndex).arg(processedArg));
-					cmd.arguments().insert(insertingIndex, processedArg);
-					insertingIndex++;
+				if (!processedArg.isEmpty())
+				{
+					bool commentArgument = false;
+					QString commentStr = "[COMMENT] ";
```
dgirardeau, line 740: I wouldn't initialize this string each time, as creating a QString object is quite expensive... Let me change this and you'll tell me if it's ok with you.

dgirardeau (changes requested):

dgirardeau:
Tell me what you think about the new version

HeadLessHUN:
if we are concerned about this then, we could do early aborts on lines, and breaking out from the loop as soon as we got a line comment identifier. But in that case we won't be able to print them out at least in place. if you want i could prepare a version for this.

dgirardeau:
Ok I'll let you test first.

I was indeed surprised that a 'full line' comment (which can actually be a partial line comment, as it can appear in the middle of the line - but that's ok, apart for the name) would still be processed in multiple chunks. I guess the chunks could be re-merged at this point, and displayed in one go (with the 'join' method of QStringList?). And then the loop can be ended earlier indeed. That's maybe the only change worthy of an effort.

dgirardeau (approved):

Issue #1898, opened 2023-12-05
ICP in CLI- Translation parameters not adjustable


SchmidL:
**Describe the bug**
Not sure if it's a bug, or not implemented at all, but the translation parameters, which can be turn on/off cannot be set via the CLI

**Additional context**
The rotation  are forwarded to the CLI via `constexpr char COMMAND_ICP_ROT[]`

**Your environment**
 - CC Version (or git hash/tag): v12.3 beta
 - OS & Version: Win 10

dgirardeau:
That's not implemented indeed. Someone would need to do it ;)

HeadLessHUN:
I think this could be closed. :)

dgirardeau:
See https://github.com/CloudCompare/CloudCompare/pull/1931

PR #1931, opened 2024-01-03
Adding command line options to -ICP to skip translations

HeadLessHUN:
Add suboptions for icp to skip each translation.

-SKIP_TX
-SKIP_TY
-SKIP_TZ

#1898

### DIFF
--- a/qCC/ccCommandLineCommands.cpp
+++ b/qCC/ccCommandLineCommands.cpp
@@ -6295,27 +6295,28 @@ bool CommandICP::process(ccCommandLineInterface& cmd)
 				QString rotation = cmd.arguments().takeFirst().toUpper();
 				if (rotation == "XYZ")
 				{
-					//keep this to not make breaking changes
+					transformationFilters &= (~CCCoreLib::RegistrationTools::SKIP_ROTATION);
+					cmd.print(QObject::tr("[ICP] Use all rotations"));
 				}
 				else if (rotation == "X")
 				{
 					transformationFilters |= CCCoreLib::RegistrationTools::SKIP_RYZ;
-					cmd.print(QObject::tr("[ICP] Skip RYZ, current filter: %1").arg(transformationFilters));
+					cmd.print(QObject::tr("[ICP] Skip RYZ"));
 				}
 				else if (rotation == "Y")
 				{
 					transformationFilters |= CCCoreLib::RegistrationTools::SKIP_RXZ;
-					cmd.print(QObject::tr("[ICP] Skip RXZ, current filter: %1").arg(transformationFilters));
+					cmd.print(QObject::tr("[ICP] Skip RXZ"));
 				}
 				else if (rotation == "Z")
 				{
 					transformationFilters |= CCCoreLib::RegistrationTools::SKIP_RXY;
-					cmd.print(QObject::tr("[ICP] Skip RXY, current filter: %1").arg(transformationFilters));
+					cmd.print(QObject::tr("[ICP] Skip RXY"));
 				}
 				else if (rotation == "NONE")
 				{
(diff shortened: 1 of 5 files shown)


qCC/ccCommandLineCommands.cpp
```diff
@@ -6292,23 +6295,27 @@ bool CommandICP::process(ccCommandLineInterface& cmd)
 				QString rotation = cmd.arguments().takeFirst().toUpper();
 				if (rotation == "XYZ")
 				{
-					transformationFilters = CCCoreLib::RegistrationTools::SKIP_NONE;
+					//keep this to not make breaking changes
 				}
 				else if (rotation == "X")
 				{
-					transformationFilters = CCCoreLib::RegistrationTools::SKIP_RYZ;
+					transformationFilters |= CCCoreLib::RegistrationTools::SKIP_RYZ;
```
dgirardeau, line 6302: why the "|' operator? Did you imagine a use case where -ROT would be used multiple times for a single ICP command? It won't be considered by the registration algorithm anyway... The algorithm only considers one SKIP_R?? option at a time.

qCC/ccCommandLineCommands.cpp
```diff
@@ -6292,23 +6295,27 @@ bool CommandICP::process(ccCommandLineInterface& cmd)
...
+					cmd.print(QObject::tr("[ICP] Skip RYZ, current filter: %1").arg(transformationFilters));
```
dgirardeau, line 6303: Is it a 'debug' log?

dgirardeau (changes requested):

qCC/ccCommandLineCommands.cpp
HeadLessHUN: it was kind of, but it is not that useless :), There were no logs about confirmations of arguments.

qCC/ccCommandLineCommands.cpp
HeadLessHUN: you are right, but what if someone send a -SKIP_TX before -ROT?

qCC/ccCommandLineCommands.cpp
dgirardeau: Ah ok, it's just that the value of 'transformationFilters' might not be super meaningful to the users... (so maybe the second part could be 'printDebug')

qCC/ccCommandLineCommands.cpp
dgirardeau: Ah ok, that's for the SKIP_T?, my bad...

dgirardeau:
Ok, I realized that the 'Debug' versions of the print/warning/error methods of the logger were not readily accessible via the command line parser. So I added them, and moved the logging of 'transformationFilters' in a single final call to printDebug.

I also made sure that explicitly calling '-ROT XYZ' would reset the rotations (just in case).

HeadLessHUN:
> Ok, I realized that the 'Debug' versions of the print/warning/error methods of the logger were not readily accessible via the command line parser. So I added them, and moved the logging of 'transformationFilters' in a single final call to printDebug.
> 
> I also made sure that explicitly calling '-ROT XYZ' would reset the rotations (just in case).

isn't it more resource if we call QObject::tr() without argument or just use the quoted string in this case? (for the future)

dgirardeau:
So 'tr' is to allow translators to translate the string. So it's all a question of whether this is user facing and someone might want to translate it one day. That something that should have done everywhere, but sadly it's missing in a lot of places...

HeadLessHUN:
I think I'm done with this as well.

dgirardeau (dismissed):

dgirardeau (approved):

PR #1932, opened 2024-01-03
Second version of the -APPLY_TRANS updates

dgirardeau:
This is related to https://github.com/CloudCompare/CloudCompare/pull/1926


dgirardeau:
@HeadLessHUN: writing the documentation in the Wiki, I realized the initial implementation had 2 issues (first I forgot to update the CHANGELOG file, and then the 'FORCE' sub-option should only be considered after -APPLY_TO_GLOBAL (as it doesn't make sense alone).

HeadLessHUN (commented):
maybe we could just check the presence of APPLY_TO_GLOBAL after processing all the known arguments. Invalidate the force if appplyToGlobal == false.
Keep the argument handling as is. 

But it is ok like this as well.

Issue #1469, opened 2021-06-01
Subsampling in Commandline


jambus1986:
Hi there,

can you make Subsampling   scalarfield-sensitive  make available in Commandline please?

thank you,
ps. noise works already very fine in commandline, well done :)

jambus1986:
i mean scalar field is used to define the densitiy of the subsampling...

scalar field value 0 -->  subsampling distance 0.001
scalar field value 0.5 --> subsampling distance 0.005
scalar field value 1 --> subsampling distance 0.2 
values between could be calculated linearly
for example.

or is this in commandline already possible and i only dont know how?
thx

lorinczszabolcs:
Hi @dgirardeau !

I have just tried subsampling and I have originally 4 scalar fields in a ply file. However, it does not retain all 4 of them, only the first one. (I don't know if it is because of the subsampling or am I opening the file incorrectly)

The command I am using: cloudcompare.CloudCompare -SILENT -AUTO_SAVE OFF -O ./labeled_00051_16E.ply -C_EXPORT_FMT PLY -SS SPATIAL 0.005 -SAVE_CLOUDS FILE "try.ply" -CLEAR -CLEAR_CLOUDS

Looking forward to your response!

dgirardeau:
The only thing we could add, is what is available via the GUI version: allowing to set a SS distance for the min scalar field value, and another SS distance for the max scalar field value (the SS distance is interpolated between the 2):
![image](https://user-images.githubusercontent.com/2735509/123538450-496c3980-d735-11eb-9a8c-e39977954dc3.png)

Jean-Roc:
@dgirardeau yes these additional parameters would be great at disposal of the command lines

HeadLessHUN:
available in 2.13 beta, can be closed.

PR #1937, opened 2024-01-07
Command line new sub-option for -SS command:

HeadLessHUN:
-SS SPATIAL {fall_back_step} USE_ACTIVE_SF {value_at_SF_min} {value_at_SF_max}

to use spatial step interpolation on active scalar field

#1469

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -285,7 +285,7 @@ v2.13.beta (???) - (??/??/2024)
 			* -SS OCTREE NUMBER_OF_POINTS PERCENT {number} to calculate NUMBER_OF_POINTS from PERCENT. PERCENT should be a decimal number between 0 and 100.
 			* -SS OCTREE CELL_SIZE {size} to deduce the octree level from the given cell size.
 			* -SS RANDOM PERCENT {number} to calculate the number of sampled points from PERCENT. PERCENT should be a decimal number between 0 and 100.
-			* -SS SPATIAL {fall_back_step} USE_ACTIVE_SF {value_at_SF_min} {value_at_SF_max} to use spatial step interpolation on active scalar field
+			* -SS SPATIAL {fallback_step} USE_ACTIVE_SF {step_at_SF_min} {step_at_SF_max} to use spatial step interpolation on active scalar field (fallback step will be used if active SF has invalid values)
 		- New sub-options for -ICP command:
 			* -SKIP_TX to skip translation along X axis
 			* -SKIP_TY to skip translation along Y axis
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -285,6 +285,7 @@ v2.13.beta (???) - (??/??/2024)
 			* -SS OCTREE NUMBER_OF_POINTS PERCENT {number} to calculate NUMBER_OF_POINTS from PERCENT. PERCENT should be a decimal number between 0 and 100.
 			* -SS OCTREE CELL_SIZE {size} to deduce the octree level from the given cell size.
 			* -SS RANDOM PERCENT {number} to calculate the number of sampled points from PERCENT. PERCENT should be a decimal number between 0 and 100.
+			* -SS SPATIAL {fall_back_step} USE_ACTIVE_SF {value_at_SF_min} {value_at_SF_max} to use spatial step interpolation on active scalar field
 		- New sub-options for -ICP command:
 			* -SKIP_TX to skip translation along X axis
 			* -SKIP_TY to skip translation along Y axis
@@ -295,7 +296,6 @@ v2.13.beta (???) - (??/??/2024)
 		- The current/default active scalar field can now be designated by index '-1' with the following commands:
 			REMOVE_SF, SF_TO_COORD, SF_ARITHMETIC, SF_OP, SF_INTERP, RENAME_SF, ICP (USE_MODEL_SF_AS_WEIGHT/USE_DATA_SF_AS_WEIGHT)
 		- The logging should be more responsive (both in the standard console-based mode and the silent mode)
-
 	- New entity picking mechanism (to not rely on the deprecated OpenGL 'names' pushing mechanism)
 		- Should hopefully solve most of the random issues with picking
 
--- a/qCC/ccCommandLineCommands.cpp
+++ b/qCC/ccCommandLineCommands.cpp
@@ -189,6 +189,7 @@ constexpr char OPTION_MESH[]							= "MESH";
 constexpr char OPTION_PERCENT[]							= "PERCENT";
 constexpr char OPTION_NUMBER_OF_POINTS[]				= "NUMBER_OF_POINTS";
 constexpr char OPTION_FORCE[]							= "FORCE";
+constexpr char OPTION_USE_ACTIVE_SF[]					= "USE_ACTIVE_SF";
 
 static bool GetSFIndexOrName(ccCommandLineInterface& cmd, int& sfIndex, QString& sfName, bool allowMinusOne = false)
 {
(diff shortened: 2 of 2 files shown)


qCC/ccCommandLineCommands.cpp
```diff
@@ -1348,19 +1349,114 @@ bool CommandSubsample::process(ccCommandLineInterface& cmd)
...
 			cmd.print(QObject::tr("\tProcessing cloud %1").arg(!desc.pc->getName().isEmpty() ? desc.pc->getName() : "no name"));
-			
+
 			CCCoreLib::CloudSamplingTools::SFModulationParams modParams(false);
+
+			//handle Use Active SF on each cloud
+			if (useActiveSF)
+			{
+				//look for the min and max sf values
+				ccScalarField* sf = desc.pc->getCurrentDisplayedScalarField();
```
dgirardeau, line 1399: So I had some doubts about this when looking at the code yesterday: since we are in the command line, I can't guarantee that there's always a 'displayed' scalar field active. It would be safer to use 'getCurrentInScalarFieldIndex' instead.

I see that there may be similar issues with other tools, I'll change that later.

qCC/ccCommandLineCommands.cpp
```diff
@@ -1348,19 +1349,114 @@ bool CommandSubsample::process(ccCommandLineInterface& cmd)
...
+					//found active scalar field
+					ScalarType sfMin = CCCoreLib::NAN_VALUE;
+					ScalarType sfMax = CCCoreLib::NAN_VALUE;
+					if (sf->countValidValues() > 0)
+					{
+						if (!ccScalarField::ValidValue(sfMin) || sfMin > sf->getMin())
+							sfMin = sf->getMin();
+						if (!ccScalarField::ValidValue(sfMax) || sfMax < sf->getMax())
+							sfMax = sf->getMax();
+						if( !(ccScalarField::ValidValue(sfMin) && ccScalarField::ValidValue(sfMax)) )
```
dgirardeau, line 1416:
```suggestion
						if (!ccScalarField::ValidValue(sfMin) || !ccScalarField::ValidValue(sfMax))
```

qCC/ccCommandLineCommands.cpp
```diff
@@ -1348,19 +1349,114 @@ bool CommandSubsample::process(ccCommandLineInterface& cmd)
...
+								.arg(sfMin)
+								.arg(sfMinSpacing)
+								.arg(sfMax)
+								.arg(sfMaxSpacing));
+						}
+					}
+					else
+					{
+						//warn the user, not use Use Active SF and keep going
+						cmd.warning(QObject::tr("\tUse active SF error: scalar field (%2) does not have any valid values.").arg(COMMAND_SET_ACTIVE_SF).arg(sf->getName()));
```
dgirardeau, line 1449:
```suggestion
						cmd.warning(QObject::tr("\tCan't use 'Use active SF': scalar field '%2' does not have any valid value.").arg(COMMAND_SET_ACTIVE_SF).arg(sf->getName()));
```

qCC/ccCommandLineCommands.cpp
```diff
@@ -1348,19 +1349,114 @@ bool CommandSubsample::process(ccCommandLineInterface& cmd)
...
+					if (sf->countValidValues() > 0)
+					{
+						if (!ccScalarField::ValidValue(sfMin) || sfMin > sf->getMin())
+							sfMin = sf->getMin();
+						if (!ccScalarField::ValidValue(sfMax) || sfMax < sf->getMax())
+							sfMax = sf->getMax();
+						if( !(ccScalarField::ValidValue(sfMin) && ccScalarField::ValidValue(sfMax)) )
+						{
+							//warn the user, not use Use Active SF and keep going
+							cmd.warning(QObject::tr("\tUse active SF error: scalar field (%1) has invalid min/max values.").arg(sf->getName()));
```
dgirardeau, line 1419:
```suggestion
							cmd.warning(QObject::tr("\tCan't use 'Use active SF': scalar field '%1' has invalid min/max values.").arg(sf->getName()));
```

qCC/ccCommandLineCommands.cpp
```diff
@@ -1348,19 +1349,114 @@ bool CommandSubsample::process(ccCommandLineInterface& cmd)
...
+					ScalarType sfMax = CCCoreLib::NAN_VALUE;
+					if (sf->countValidValues() > 0)
+					{
+						if (!ccScalarField::ValidValue(sfMin) || sfMin > sf->getMin())
+							sfMin = sf->getMin();
+						if (!ccScalarField::ValidValue(sfMax) || sfMax < sf->getMax())
+							sfMax = sf->getMax();
+						if( !(ccScalarField::ValidValue(sfMin) && ccScalarField::ValidValue(sfMax)) )
+						{
+							//warn the user, not use Use Active SF and keep going
```
dgirardeau, line 1418:
```suggestion
							//warn the user, don't use 'Use Active SF' and keep going
```

qCC/ccCommandLineCommands.cpp
```diff
@@ -1348,19 +1349,114 @@ bool CommandSubsample::process(ccCommandLineInterface& cmd)
...
+
+			//handle Use Active SF on each cloud
+			if (useActiveSF)
+			{
+				//look for the min and max sf values
+				ccScalarField* sf = desc.pc->getCurrentDisplayedScalarField();
+				if (!sf)
+				{
+					//warn the user, not use active SF and keep going
+					cmd.warning(QObject::tr("\tUse active SF error: not found active scalar field. Set one with '-%1'").arg(COMMAND_SET_ACTIVE_SF));
```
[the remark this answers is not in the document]
dgirardeau: More generally, I would treat this kind of error as real 'errors' and bail out (as that's not what the user asked for, and it changes the result a lot...)

qCC/ccCommandLineCommands.cpp
dgirardeau: real error?

qCC/ccCommandLineCommands.cpp
dgirardeau: real error?

qCC/ccCommandLineCommands.cpp
```diff
@@ -1348,19 +1349,114 @@ bool CommandSubsample::process(ccCommandLineInterface& cmd)
...
+						//warn the user, not use Use Active SF and keep going
+						cmd.warning(QObject::tr("\tUse active SF error: scalar field (%2) does not have any valid values.").arg(COMMAND_SET_ACTIVE_SF).arg(sf->getName()));
+
+					}
+				}
+			}
+
+			if (useActiveSF && !modParams.enabled)
+			{
+				cmd.print(QObject::tr("\tUse active SF: disabled and fall back to static spacing."));
```
dgirardeau, line 1457:
```suggestion
				cmd.print(QObject::tr("\t'Use active SF' disabled. Falling back to constant spacing."));
```

dgirardeau: But that may be deprecated if the warnings above are treated as real errors.

dgirardeau (changes requested):

qCC/ccCommandLineCommands.cpp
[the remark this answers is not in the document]
HeadLessHUN: In my tests. If the cloud has an original spacing of >0.01  and you run a feature calculation with the same kernel size {0.01} then it will produce an all NAN sf. i don't really feel like we should stop processing because one of the cloud did not have enough points to calculate a proper sf. That is why i kept a {fall_back_value}.

Or should the user make sure that every cloud has the same valid fields? with select_entities or something like that?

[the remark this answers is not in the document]
dgirardeau: Ah ok. Can you detect if they provide a fallback value or not then? So as to be sure they are aware and made it on purpose?

qCC/ccCommandLineCommands.cpp
HeadLessHUN: but that is gonna be an issue inside the called subsampling tools as well, because it calls the same displayedscalarfield.

So better would be to just make sure there is a displayed scalar field?
In this case
desc.pc->setDisplayedScalarField(index)

dgirardeau: Oh, I see it's a big mess, as normally the displayed scalar field should be the 'out' scalar field in fact... (not the 'in'). But that's not consistent with how tools should work. Typically the subsampling tool should look at the 'in' scalar field 😅 (which should be properly set before calling the tool).

I guess you can forget about this for now then. I'll have to make a big cleanup pass later.

qCC/ccCommandLineCommands.cpp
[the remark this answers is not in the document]
HeadLessHUN: well the command will throw an error since the fallback value is a must. Not optional, only the USE_ACTIVE_SF and the rest is optional.

-SS SPATIAL {fall_back_step} USE_ACTIVE_SF {value_at_SF_min} {value_at_SF_max} to use spatial step interpolation on active scalar field

CHANGELOG.md
```diff
@@ -285,6 +285,7 @@ v2.13.beta (???) - (??/??/2024)
 			* -SS OCTREE NUMBER_OF_POINTS PERCENT {number} to calculate NUMBER_OF_POINTS from PERCENT. PERCENT should be a decimal number between 0 and 100.
 			* -SS OCTREE CELL_SIZE {size} to deduce the octree level from the given cell size.
 			* -SS RANDOM PERCENT {number} to calculate the number of sampled points from PERCENT. PERCENT should be a decimal number between 0 and 100.
+			* -SS SPATIAL {fall_back_step} USE_ACTIVE_SF {value_at_SF_min} {value_at_SF_max} to use spatial step interpolation on active scalar field
```
dgirardeau, line 288:
```suggestion
			* -SS SPATIAL {fallback_step} USE_ACTIVE_SF {step_at_SF_min} {step_at_SF_max} to use spatial step interpolation on active scalar field (fallback step will be used if active SF has invalid values)
```

qCC/ccCommandLineCommands.cpp
HeadLessHUN: I'll keep that in my mind for later :)

dgirardeau (changes requested):

dgirardeau (approved):

PR #1944, opened 2024-01-26
Snap and snapcraft builds are no longer supported

rjanvier:
We now fully moved all our linux packaging effort to flatpak. The snapcraft file is not updated for years and is broken. This commit would avoid user facing/reporting issue if ever someone tries to build from this legacy file.


VVD:
Hi!
You removed the icons from the repository, where can I find them now?

PR #1946, opened 2024-01-28
dlib for MacOS

prascle:
I have already made pull requests on this one but first using the wrong repository for dlib, then using my clone but without the patch. This one is now with the good commit! All that for commenting one line :-)

I hope it's OK now!
Paul


dgirardeau:
Silly question, but could we have simply defined the DLIB_NO_GUI_SUPPORT macro in the project? (via CMake)

prascle:
> Silly question, but could we have simply defined the DLIB_NO_GUI_SUPPORT macro in the project? (via CMake)

Hum,  I am afraid you are right !

Issue #1949, opened 2024-01-31
[CLI] implement a way to get less verbosity


ryukisai:
**Describe the bug**
Hi, i'm using CLI for most of my pre-processing steps. I'm very happy with it but even though i'm using the -SILENT option, i find there is a ton of verbose left that i find unwanted (especially 60 lines just for the plugins), see the caption:
![2024-01-31 17_58_08-Export](https://github.com/CloudCompare/CloudCompare/assets/6038397/9c598357-2451-4f32-945e-5c072322b383)

**How to reproduce**
My win CMD batch uses series of commands and the call to CC is this:
set CloudComparePath="C:\Program Files\CloudCompare\CloudCompare.exe"
%CloudComparePath% -SILENT -O "blabla.e57" -AUTO_SAVE OFF -MERGE_CLOUDS […]

**Expected behaviour**
If i'm not using the option wrong, i'd appreciate an option to further reduce the verbose level.
Thank you!

**Your environment**
 - CC Version (or git hash/tag): 2.13 β build from 2024-01-27
 - OS & Version: win 11 22H2 & 23h2

dgirardeau:
Yes, I believe this is because we load the plugins before anything else (even before interpreting the commands...). I guess we should check if -SILENT is the very first option at least.

ryukisai:
Ok this makes sense to load the plug-ins first as some commands might rely on them.
This request is definitely low priority so let me know whenever this reaches the to-do list.
Merci!

HeadLessHUN:
-SILENT means that it won't load the gui. And not a verbosity level :)
I was thinking about a verbosity level for command line but probably too much work would be to sort every message to a correct level.
Currently only 3 level are available info/warning/error

We could easily create a command to set these levels, but there will be little to none meaning because almost all message is at info level, unless it is an error and then it just halts the command so you will notice that.

ryukisai:
Hi! I took another look and the wiki, the description for -SILENT is:
> enables silent mode (no console will appear)
> Warning: must be first if required. 
Indeed it has nothing to do with verbose level. However console for me meant the text console, not the whole GUI.

Well it would be good if we could skip the plugins info messages. But it would require creating another level just for these, which does not make a lot of sense, i agree.
Otherwise you're right my batch is not in debugging phase anymore so i could ignore all messages. It's helpful to know which command is currently running to have an idea of the progress, but it's not mandatory.

HeadLessHUN:
[...]

PR #1956, opened 2024-02-06
Verbosity level and other related changes

HeadLessHUN:
#1949

### DIFF
--- a/libs/CCPluginAPI/include/ccCommandLineInterface.h
+++ b/libs/CCPluginAPI/include/ccCommandLineInterface.h
@@ -295,7 +295,9 @@ public: //file I/O
 public: //logging
 
 	//logging
+	virtual void printVerbose(const QString& message) const = 0;
 	virtual void print(const QString& message) const = 0;
+	virtual void printHigh(const QString& message) const = 0;
 	virtual void printDebug(const QString& message) const = 0;
 	virtual void warning(const QString& message) const = 0;
 	virtual void warningDebug(const QString& message) const = 0;
--- a/qCC/ccCommandLineCommands.cpp
+++ b/qCC/ccCommandLineCommands.cpp
@@ -4427,7 +4427,7 @@ bool CommandSampleMesh::process(ccCommandLineInterface& cmd)
 }
 
 CommandCompressFWF::CommandCompressFWF()
-	: ccCommandLineInterface::Command(QObject::tr("Sample mesh"), COMMAND_COMPRESS_FWF)
+	: ccCommandLineInterface::Command(QObject::tr("Compress FWF"), COMMAND_COMPRESS_FWF)
 {}
 
 bool CommandCompressFWF::process(ccCommandLineInterface& cmd)
@@ -6262,7 +6262,7 @@ bool CommandRenameEntities::process(ccCommandLineInterface& cmd)
 			{
 				currentMeshName += "-" + QString::number(index).rightJustified(3, '0');
 			}
-			cmd.error("Mesh '" + desc.mesh->getName() + "' renamed to '" + currentMeshName + "'");
+			cmd.print("Mesh '" + desc.mesh->getName() + "' renamed to '" + currentMeshName + "'");
 			//rename the Mesh entity
 			desc.mesh->setName(currentMeshName);
 			//rename the Root entity
@@ -7362,7 +7362,7 @@ bool CommandClearClouds::process(ccCommandLineInterface& cmd)
 }
 
 CommandPopClouds::CommandPopClouds()
-	: ccCommandLineInterface::Command(QObject::tr("Pop clouds"), COMMAND_POP_CLOUDS)
+	: ccCommandLineInterface::Command(QObject::tr("Pop cloud"), COMMAND_POP_CLOUDS)
 {}
 
 bool CommandPopClouds::process(ccCommandLineInterface& cmd)
@@ -7384,7 +7384,7 @@ bool CommandClearMeshes::process(ccCommandLineInterface& cmd)
 }
 
 CommandPopMeshes::CommandPopMeshes()
-	: ccCommandLineInterface::Command(QObject::tr("Pop meshes"), COMMAND_POP_MESHES)
+	: ccCommandLineInterface::Command(QObject::tr("Pop mesh"), COMMAND_POP_MESHES)
 {}
 
 bool CommandPopMeshes::process(ccCommandLineInterface& cmd)
@@ -7590,8 +7590,8 @@ bool CommandDebugCmdLine::process(ccCommandLineInterface& cmd)
 	cmd.print("[DEBUG]");
 
 	cmd.print("******************************************");
-	cmd.print("Number of clouds: " + QString::number(cmd.clouds().size()));
-	cmd.print("Number of meshes: " + QString::number(cmd.meshes().size()));
+	cmd.print("Number of selected clouds: " + QString::number(cmd.clouds().size()));
+	cmd.print("Number of selected meshes: " + QString::number(cmd.meshes().size()));
 
 	cmd.print("******************************************");
 	const QStringList& arguments = cmd.arguments();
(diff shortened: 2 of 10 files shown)


qCC/ccCommandLineCommands.cpp
```diff
@@ -7620,3 +7513,28 @@ bool CommandDebugCmdLine::process(ccCommandLineInterface& cmd)
...
+	}
+
+	bool ok = false;
+	int verbosityLevel = cmd.arguments().takeFirst().toInt(&ok);
+	if (!ok) {
+		return cmd.error(QObject::tr("Invalid verbosity level %s").arg(verbosityLevel));
+	}
+	else
+	{
+		cmd.print(QObject::tr("Set verbosity level to %i").arg(verbosityLevel));
```
dgirardeau, line 7535:
```suggestion
		cmd.print(QObject::tr("Set verbosity level to %1").arg(verbosityLevel));
```

qCC/ccCommandLineCommands.cpp
```diff
@@ -1600,11 +1591,11 @@ bool CommandSubsample::process(ccCommandLineInterface& cmd)
 		for (CLCloudDesc& desc : cmd.clouds())
 		{
 			//calculate octree before subsampling, it is passed to subsampling, so it won't be recalculated there.
-			CCCoreLib::DgmOctree* octree = desc.pc->computeOctree(nullptr, false).data();
+			CCCoreLib::DgmOctree* octree = desc.pc->getOctree().data();
+			if (!octree)
+				octree = desc.pc->computeOctree(nullptr, false).data();
 			if (!octree)
-			{
```
dgirardeau, line 1605: to be restored

qCC/ccCommandLineCommands.cpp
```diff
@@ -7620,3 +7513,28 @@ bool CommandDebugCmdLine::process(ccCommandLineInterface& cmd)
...
+
+CommandSetVerbosity::CommandSetVerbosity()
+	: ccCommandLineInterface::Command(QObject::tr("Set Verbosity"), COMMAND_VERBOSITY)
+{}
+
+bool CommandSetVerbosity::process(ccCommandLineInterface& cmd)
+{
+	if (cmd.arguments().empty())
+	{
+		return cmd.error(QObject::tr("Missing parameter: verbosity level after: %s").arg(COMMAND_VERBOSITY));
```
dgirardeau, line 7525:
```suggestion
		return cmd.error(QObject::tr("Missing parameter: verbosity level after: %1").arg(COMMAND_VERBOSITY));
```

qCC/ccCommandLineCommands.cpp
```diff
@@ -7620,3 +7513,28 @@ bool CommandDebugCmdLine::process(ccCommandLineInterface& cmd)
...
+	}
+
+	bool ok = false;
+	int verbosityLevel = cmd.arguments().takeFirst().toInt(&ok);
+	if (!ok) {
```
dgirardeau, line 7530:
```suggestion
	if (!ok)
	{
```

qCC/ccCommandLineCommands.cpp
```diff
@@ -7620,3 +7513,28 @@ bool CommandDebugCmdLine::process(ccCommandLineInterface& cmd)
...
+		return cmd.error(QObject::tr("Invalid verbosity level %s").arg(verbosityLevel));
```
dgirardeau, line 7531:
```suggestion
		return cmd.error(QObject::tr("Invalid verbosity level %1").arg(verbosityLevel));
```

qCC/main.cpp
```diff
@@ -116,6 +116,23 @@ int main(int argc, char **argv)
             commandLine = false;
             lastArgumentIndex += 2;
         }
+		if (argumentsLocal8Bit[lastArgumentIndex].toUpper() == "-VERBOSITY")
```
dgirardeau, line 119: I guess you should check that 'lastArgumentIndex < argumentsLocal8Bit.size()' first?

qCC/main.cpp
```diff
@@ -116,6 +116,23 @@ int main(int argc, char **argv)
...
+		{
+			//remove verified local option
+			argumentsLocal8Bit.removeAt(lastArgumentIndex);
+			bool ok = false;
+			int verbosityLevel = argumentsLocal8Bit[lastArgumentIndex + 1].toInt(&ok);
+			if (!ok)
+			{
+				ccLog::Error(QObject::tr("Invalid verbosity level %s").arg(verbosityLevel));
```
dgirardeau, line 127:
```suggestion
				ccLog::Error(QObject::tr("Invalid verbosity level %1").arg(verbosityLevel));
```

dgirardeau (changes requested):

HeadLessHUN:
Daniel, i wanted to have higher level for command name print/subprocess time. That is why i used warning, maybe we should/could add a new level instead?:)

dgirardeau:
Yes I realized that. I think we need to add more levels indeed.

Warning and Errors can't be 'ignored' (at least not errors). And they have to be reserved for real warnings and errors.

I think we will have to work a little bit more on that, and sadly we might not be able to add that to the release...

HeadLessHUN:
I was thinking about flipping debug and print as well, and allow debug in release mode as well, just make the standard as default.

Maybe we could add a higher lvl print. and that should be enough?

dgirardeau:
Yes, Debug could become the 'verbose' mode, disabled by default (unless you are in Debug mode).

What was your intent by putting the timing logs as 'warning'? Do you think they are more important than standard messages? Or on the contrary part of the verbose/debug logs?

HeadLessHUN:
well i wanted to have a start/end of a command so that should be higher level i think, At least for me these are more important, then messages inside commands.

dgirardeau:
Ok, so something like that?
Debug = 0 (=Verbose)
Print = 1
Important/High = 2
Warning = 3
Error = 4

HeadLessHUN:
for me that looks fine :) I don't really know if we want to mess with gui mode logs or not... I usually never look the console log in gui mode. So i wouldn't bother. But maybe if we are implementing this, it worth a select menu inside the preferences menu.

dgirardeau:
Yes, for now it's ok to only give this option to the command line users (as in the GUI mode, you can simply hide the console).

I'll try something quick so as to ship that with the release.

dgirardeau:
Ok, here's the new version:
image

I've tested it quickly. You can try it on your side. And we should be good to add it to the release!

qCC/ccCommandLineParser.h
```diff
@@ -55,7 +55,9 @@ class ccCommandLineParser : public ccCommandLineInterface
 	const QStringList& arguments() const override { return m_arguments; }
 	bool registerCommand(Command::Shared command) override;
 	QDialog* widgetParent() override { return m_parentWidget; }
+	void printVerbpse(const QString& message) const override;
```
HeadLessHUN, line 58: typoo here too

libs/CCPluginAPI/include/ccCommandLineInterface.h
```diff
@@ -295,7 +295,9 @@ class CCPLUGIN_LIB_API ccCommandLineInterface
 public: //logging
 
 	//logging
+	virtual void printVerbpse(const QString& message) const = 0;
```
HeadLessHUN, line 298: typoo
printVerbose

qCC/ccCommandLineParser.h
dgirardeau: Wow, I was tired 😫

HeadLessHUN:
currently i have build error, but i'm sure it just my environment went wrong somewhere. but i was not able to test these changes yet.

dgirardeau:
Ah ok, keep me updated!

HeadLessHUN:
> Ah ok, keep me updated!

well something is really f*cked up :) my env however is clean and reset to the verbosity_level branch, but they contain changes from the master branch which you pushed yesterday. So there are missing files. bcadd49365e51e75836529fee64c5253e3e6932c

dgirardeau:
Ok I need to merge some stuff in the PythonPlugin first.

ryukisai:
Thanks for this guys! I had not realized it was already implemented.

PR #1961, opened 2024-02-12
qPCL: add support for (u)int64 pcd field types

SamuelH91:
- update dependencies in conda.yml
- PCL 1.13.1 supports int64 and uint64 field types
- PCL dependency from vtk forced also to update ffmpeg
- qAnimation: update QVideoEncoder to use newer ffmpeg API

### DIFF
--- a/.ci/conda.yml
+++ b/.ci/conda.yml
@@ -2,12 +2,12 @@ name: CloudCompareDev
 channels:
   - conda-forge
 dependencies:
-  - eigen=3.3.*
-  - ffmpeg=4.2.*
+  - eigen=3.4.*
+  - ffmpeg=6.1.*
   - ninja
-  - pcl=1.9.*
-  - pdal=2.1.*
-  - qt=5.12.*
+  - pcl=1.13.1
+  - pdal=2.6.*
+  - qt=5.15.*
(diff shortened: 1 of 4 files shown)


dgirardeau:
Hey. That's interesting, but this PR doesn't compile on my side (with FFMPEG 4.3 and PCL 1.12).

While I can upgrade to FFMPEG 6.1 easily, would it be possible to add some preprocessors to keep the compatibility with PCL 1.12?

@rjanvier @prascle Would moving to FFMPEG 6.1 work on your Linux and macOS builds?

dgirardeau:
Apparently you have forced compilation with Qt5.15, which is not supported yet

dgirardeau:
For the records, the QVideoEncoder part of this PR was merged here: https://github.com/CloudCompare/CloudCompare/pull/2006

dgirardeau:
The Linux builds failed on the CI, but sadly there's no more log, so I'm not sure of the reason.

@rjanvier do you think it is safe to merge this?

dgirardeau (dismissed):

dgirardeau:
Ah no, I see the branch is also in conflict...

PR #1972, opened 2024-02-28
Add an "Invert" button to the clipping box tool

### LINKED ISSUE #1971: Allow to invert option the clipping box selection

Add an "invert" option to the clipping box tool that would keep the points outside of the box instead of inside the box.

**Describe alternatives you've considered**
We could also have the Clipping Box output a "remaining" point cloud just like the segment tool does.

**Is your feature request related to a problem? Please describe.**
I often need to remove some points inside a closed point cloud. For example, a box in the middle of a room

I can place the view inside the room and use the segment (scissors) tool, but it is a bit tedious and often remove too much of the geometry behind the object. Another option would be to play with the Z-near/Z-Far, but the shortcuts are not precise enough and modifying them using the menu is slow. Sometimes, I can also split the room into three point clouds : the floor, the ceiling and the rest, so I can use the orthographic view and the segment tool to remove the obstacles easily. But this is quite tedious.

Hugal31:
Add a button on the top of the clipping box dialog. This allows to remove the points inside the box instead of removing the points outside.

clippingboxinvert

I am not familiar with the CloudCompare code, so it is possible the code is very bad.

Closes #1971.


qCC/ccClippingBoxTool.cpp
```diff
@@ -428,6 +428,12 @@ ccHObject* GetSlice(ccHObject* obj, ccClipBox* clipBox, bool silent)
 			return nullptr;
 		}
 		clipBox->flagPointsInside(inputCloud, &selectionTable);
+
+		if (invert)
```
dgirardeau, line 432: Could you maybe add a 'invert' boolean parameter to the 'flagPointsInside' method and do this directly there?

(or maybe even rename the method ```flagPoints(inputCloud, &selectionTable, bool inside);```?

I can do it if you prefer.

Hugal31: Makes more sens, I'll do it

dgirardeau:
Thanks for the update.

So for meshes, it happens inside  ccCropTool::Crop (the call on line 464). You already have an 'inside' boolean in 3rd position. So you should simply have to set the right value here instead of true.

And due to where you added the button, it seems to me that you also wanted to change the way things are displayed? Because if it applies only to the exported slices, it may be better to move this option closer or inside the 'Slices' frame?

PR #1975, opened 2024-03-05
little fix plugin CSF

retepik:
fix QString::arg: Argument missing: Custom max iteration set: % , 500


dgirardeau (approved):

dgirardeau:
Good catch!

dgirardeau:
I've merged it in the 'various_improvements' branch which is the branch for the next version 2.13.1 (to be released soon).

PR #1993, opened 2024-04-10
Update LasDetails.cpp

p-leroy:
The smallest valid point format value is 4 for a cloud with waveforms. The dialog may be wrongly initialized with a point format of 0 if this is not taken into account. In command line this could result in missing waveforms when saving.

### DIFF
--- a/plugins/core/IO/qLASIO/src/LasDetails.cpp
+++ b/plugins/core/IO/qLASIO/src/LasDetails.cpp
@@ -264,8 +264,8 @@ namespace LasDetails
 		{
 			bool hasGpsTime   = cloud.getScalarFieldIndexByName(LasNames::GpsTime) != -1;
 			int  pointFormat  = 0;
-			int  minorVersion = 2;
-			if (hasWaveform || previousMinorVersion >= 3)
+			int  minorVersion = previousMinorVersion == 3 ? 3 : 2;
+			if (hasWaveform)
 			{
 				minorVersion = 3;
 				pointFormat = 4;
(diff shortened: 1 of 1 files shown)


plugins/core/IO/qLASIO/src/LasDetails.cpp
```diff
@@ -268,11 +268,8 @@ namespace LasDetails
 			if (hasWaveform || previousMinorVersion >= 3)
 			{
 				minorVersion = 3;
-				if (hasGpsTime)
```
dgirardeau, line 271: Wouldn't the test be instead:
```
if (hasWaveform || hasGpsTime)
{
    pointFormat = 4;
}
```

p-leroy: Not sure of that. As soon as you have waveforms, to should get either 4 or 5 as a point format, GpsTime being in both.
Maybe the test
`if (hasWaveform || previousMinorVersion >= 3) `
should besimply  changed to
`if (hasWaveform)`
as minorVersion is set to 3 after the test

dgirardeau: @tmontaigu any preference?

tmontaigu: I think it should be
```c++
int  minorVersion = previousMinorVersion == 3 ? 3 : 2;
if (hasWaveform) {
   pointFormat = 4;
   // ...
}
```

p-leroy: OK, let's do it that way. I will change the test.

rjanvier (approved):

PR #2002, opened 2024-05-03
Substantially improve speed of cloud merge operation

tpwrules:
Improves the speed of merging 25 clouds with 6M points each and 7 scalar fields from over 2 minutes to under 10 seconds, an order of magnitude improvement.

Please see commit messages for details.

### DIFF
--- a/libs/qCC_db/src/ccScalarField.cpp
+++ b/libs/qCC_db/src/ccScalarField.cpp
@@ -205,7 +205,7 @@ void ccScalarField::computeMinAndMax()
 
 						if (ValidValue(val))
 						{
-							unsigned bin = static_cast<unsigned>(floor(static_cast<double>(val - m_displayRange.min())*step));
+							unsigned bin = static_cast<unsigned>((val - m_displayRange.min())*step);
 							++m_histogram[std::min(bin, numberOfClasses - 1)];
 						}
 					}
--- a/libs/qCC_db/src/ccScalarField.cpp
+++ b/libs/qCC_db/src/ccScalarField.cpp
@@ -151,7 +151,7 @@ void ccScalarField::setLogScale(bool state)
 	if (m_logScale != state)
 	{
 		m_logScale = state;
-		if (m_logScale && m_minVal < 0)
+		if (m_logScale && getMin() < 0)
 		{
 			ccLog::Warning("[ccScalarField] Scalar field contains negative values! Log scale will only consider absolute values...");
 		}
@@ -164,7 +164,7 @@ void ccScalarField::computeMinAndMax()
 {
 	ScalarField::computeMinAndMax();
 
-	m_displayRange.setBounds(m_minVal, m_maxVal);
+	m_displayRange.setBounds(getMin(), getMax());
 
 	//update histogram
 	{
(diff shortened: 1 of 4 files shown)


libs/qCC_db/src/ccScalarField.cpp
```diff
@@ -205,7 +205,7 @@ void ccScalarField::computeMinAndMax()
 
 						if (ValidValue(val))
 						{
-							unsigned bin = static_cast<unsigned>(floor(static_cast<double>(val - m_displayRange.min())*step));
+							unsigned bin = static_cast<unsigned>((val - m_displayRange.min())*step);
```
dgirardeau, line 208: why removing the call to 'floor'? This changes the result... and I believe 'floor' was the right behavior 👼

dgirardeau (changes requested):

dgirardeau:
And it seems something must be fixed on macOS? (even though I'm not sure this is related to your PR 🤔 )

rjanvier:
@dgirardeau macOS-latest GHA runner is now `macOS-14`, not `macOS-13` anymore. You can restore the old CI behavior by replacing `macOS-lates`t by `macOS-13` in our GHA yaml file. I will try to look at the issue more in depth today.

libs/qCC_db/src/ccScalarField.cpp
tpwrules: I do not believe this changes the result. According to [here](https://cplusplus.com/doc/tutorial/typecasting/), a float to int cast truncates the value, i.e. removes the decimal part. The expression is never negative, so that produces the same result as flooring it.

As for why to remove it, the promotion to double and the call to floor cause a substantial overhead. Removing both roughly doubles execution speed of this function.

Can you help me understand why you believe the result changes?

rjanvier:
@dgirardeau : macOS build [fixes](https://github.com/CloudCompare/CloudCompare/pull/2005).

dgirardeau:
> @dgirardeau : macOS build [fixes](https://github.com/CloudCompare/CloudCompare/pull/2005).

Thanks! I've merged it.

dgirardeau (approved):

libs/qCC_db/src/ccScalarField.cpp
rjanvier: I think floor and cast are not equivalent when the value is negative. is this value could be negative before being converted to unsigned?

dgirardeau: Well, that was bugging me as well, but I don't see how it would have worked either with floor if the value could be smaller than m_displayRange.min() 🤔 

This code is quite old, I don't remember much about it.

tpwrules: m_displayRange.min() is set to the scalar field's minimum value at the start of the function [here](https://github.com/CloudCompare/CloudCompare/blob/5aec741ad191bb88da267f6fd8157f99142c4807/libs/qCC_db/src/ccScalarField.cpp#L167), so that expression can't be negative. There can't be any values in the scalar field lower than its minimum value.

PR #2029, opened 2024-06-28
Macos packaging based on Pixi

rjanvier:
Rework the macOS packaging of CC:

- Use[ Pixi](https://github.com/prefix-dev/pixi/) as a package manager. It manages package from both conda (mainly for C++ deps and build tools) and PyPI. Pixi come with a strong reproductible build focus and an internal task launcher.  As a result any macOS user should be able replicate the build on their computer with minimal fuss (Pixi is the only requirement).
- Update and optimize previous macOS post install script (for bundling and signing), this is more for internal purposes (cc @prascle)


dgirardeau (approved):
I have absolutely no idea what this does, but I trust you 😆

rjanvier:
Thanks Daniel. it seems ok on my side but I'm waiting for @prascle's feedback.

PR #2030, opened 2024-06-28
Update libE57format to v3.2.0

rjanvier:


dgirardeau:
Just to double check: is the E57 plugin compiled by the CI on all platforms? 👼

rjanvier:
Yes it is.

dgirardeau (approved):

PR #2038, opened 2024-07-20
LAS: add option to not decompose classification

tmontaigu:
This adds (back) the option not to decompose the classification field into "classification", "synthethic flag", "key point flag", "withheld flag".

This only is only shown for point format that it can be applied to (fmt <= 5).

### DIFF
--- a/plugins/core/IO/qLASIO/include/LasScalarFieldLoader.h
+++ b/plugins/core/IO/qLASIO/include/LasScalarFieldLoader.h
@@ -67,7 +67,8 @@ class LasScalarFieldLoader
 	/// the classification, synthetic flag, key_point flag, withheld flag.
 	///
 	/// Only applies to point format <= 5 (ie field Classification, not ExtendedClassification)
-	inline void setDecomposeClassification(bool state) {
+	inline void setDecomposeClassification(bool state)
+	{
 		m_decomposeClassification = state;
 	}
 
--- a/plugins/core/IO/qLASIO/include/LasScalarFieldLoader.h
+++ b/plugins/core/IO/qLASIO/include/LasScalarFieldLoader.h
@@ -63,6 +63,14 @@ class LasScalarFieldLoader
 		m_force8bitRgbMode = state;
 	}
 
+	/// Sets whether the classification field should be decomposed into
+	/// the classification, synthetic flag, key_point flag, withheld flag.
+	///
+	/// Only applies to point format <= 5 (ie field Classification, not ExtendedClassification)
+	inline void setDecomposeClassification(bool state) {
+		m_decomposeClassification = state;
+	}
+
 	/// If nan, this value will be ignored and the time shift
 	/// will be taken using the first value encountered.
 	inline void setManualTimeShift(double timeShift)
@@ -116,6 +124,7 @@ class LasScalarFieldLoader
 
   private:
 	bool                              m_force8bitRgbMode{false};
+	bool                              m_decomposeClassification{true};
 	bool                              m_ignoreFieldsWithDefaultValues{true};
 	double                            m_manualTimeShiftValue{std::numeric_limits<double>::quiet_NaN()};
 	unsigned char                     m_colorCompShift{0};
--- a/plugins/core/IO/qLASIO/include/LasScalarFieldSaver.h
+++ b/plugins/core/IO/qLASIO/include/LasScalarFieldSaver.h
@@ -42,10 +42,12 @@ class LasScalarFieldSaver
 		// a field with name that is one of the decomposed flags.
 		// However check by looking at the max value may be better to handle
 		// clouds not coming from LAS at all.
-		for (const LasScalarField& field: m_standardFields) {
+		for (const LasScalarField& field : m_standardFields)
+		{
 			if (strcmp(field.name(), LasNames::Classification) == 0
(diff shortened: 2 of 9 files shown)


plugins/core/IO/qLASIO/include/LasScalarFieldLoader.h
```diff
@@ -63,6 +63,14 @@ class LasScalarFieldLoader
 		m_force8bitRgbMode = state;
 	}
 
+	/// Sets whether the classification field should be decomposed into
+	/// the classification, synthetic flag, key_point flag, withheld flag.
+	///
+	/// Only applies to point format <= 5 (ie field Classification, not ExtendedClassification)
+	inline void setDecomposeClassification(bool state) {
```
dgirardeau, line 70: '{' on the next line (I'll make a commit as for the other PR anyway)

plugins/core/IO/qLASIO/include/LasScalarFieldSaver.h
```diff
@@ -38,6 +38,17 @@ class LasScalarFieldSaver
 	inline void setStandarFields(std::vector<LasScalarField>&& standardFields)
 	{
 		m_standardFields = standardFields;
+		// Another way to check this could have been to check for the existence of
+		// a field with name that is one of the decomposed flags.
+		// However check by looking at the max value may be better to handle
+		// clouds not coming from LAS at all.
+		for (const LasScalarField& field: m_standardFields) {
```
dgirardeau, line 45: same

plugins/core/IO/qLASIO/include/LasScalarFieldSaver.h
```diff
@@ -77,4 +88,8 @@ class LasScalarFieldSaver
   private:
 	std::vector<LasScalarField>      m_standardFields;
 	std::vector<LasExtraScalarField> m_extraFields;
+	/// Whether the classification flags from point format <= 5 were
+	/// decomposed into individual scalar field (synthetic, keypoint, withheld)
+	/// are if their values was kept packed into the classification field.
```
dgirardeau, line 93:
```suggestion
	/// or if their values were kept packed into the classification field.
```

plugins/core/IO/qLASIO/include/LasScalarFieldSaver.h
```diff
@@ -77,4 +88,8 @@ class LasScalarFieldSaver
   private:
 	std::vector<LasScalarField>      m_standardFields;
 	std::vector<LasExtraScalarField> m_extraFields;
+	/// Whether the classification flags from point format <= 5 were
+	/// decomposed into individual scalar field (synthetic, keypoint, withheld)
```
dgirardeau, line 92:
```suggestion
	/// decomposed into individual scalar fields (synthetic, keypoint, withheld)
```

plugins/core/IO/qLASIO/src/LasOpenDialog.cpp
```diff
@@ -268,6 +272,10 @@ bool LasOpenDialog::shouldForce8bitColors() const
 	return force8bitColorsCheckBox->isChecked();
 }
 
+bool LasOpenDialog::shouldDecomposeClassification() const {
```
dgirardeau, line 275: same as above

plugins/core/IO/qLASIO/src/LasOpenDialog.cpp
```diff
@@ -401,3 +409,21 @@ void LasOpenDialog::onCurrentTabChanged(int index)
 		applyAllButton->setText(APPLY_ALL_TEXT);
 	}
 }
+
+void LasOpenDialog::onDecomposeClassificationToggled(bool checked) {
```
dgirardeau, line 413: and same about '{ ' 😆

plugins/core/IO/qLASIO/src/LasOpenDialog.cpp
```diff
@@ -401,3 +409,21 @@ void LasOpenDialog::onCurrentTabChanged(int index)
...
+	static constexpr std::array<const char *, 3>flagNames = {
```
dgirardeau, line 414:
```suggestion
	static constexpr std::array<const char *, 3>flagNames {
```

plugins/core/IO/qLASIO/src/LasOpenDialog.cpp
```diff
@@ -401,3 +409,21 @@ void LasOpenDialog::onCurrentTabChanged(int index)
...
+			LasNames::SyntheticFlag, LasNames::KeypointFlag, LasNames::WithheldFlag};
+
+	for (const char *flagName: flagNames) {
```
dgirardeau, line 417: Actually, it's as if the clang format was not applied?

tmontaigu: No i forgot to run it

Potentially we could add formatting checks in the CI

It would also be cool to have a clang format for the whole project

dgirardeau (approved):

plugins/core/IO/qLASIO/src/LasOpenDialog.cpp
dgirardeau: Yes, we should apply it to the whole project and make a huge commit

PR #2067, opened 2024-10-02
Add qVoxFall to the CI

rjanvier:
- Update qVoxFall to point to the latest commit (it makes qVoxFall cross platform and leaks less memory :D)
- add qVoxFall to the CI for all supported platforms.


dgirardeau:
Thanks!

PR #2083, opened 2024-10-31
correct bug in command line call for the setting of the computation mode

p-leroy:
When calling M3C2 multiple times in the same command line call, depending on the configuration (cloud1 with normals), the second call may fail if the vertical mode is requested. In case the vertical mode is requested, force it.

### DIFF
--- a/plugins/core/Standard/qM3C2/src/qM3C2Dialog.cpp
+++ b/plugins/core/Standard/qM3C2/src/qM3C2Dialog.cpp
@@ -377,8 +377,8 @@ qM3C2Normals::ComputationMode qM3C2Dialog::getNormalsComputationMode() const
 	{
 		if (getRequestedComputationMode() == qM3C2Normals::VERT_MODE)
 		{
-			return qM3C2Normals::VERT_MODE;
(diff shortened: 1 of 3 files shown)


plugins/core/Standard/qM3C2/src/qM3C2Dialog.cpp
```diff
@@ -372,6 +372,16 @@ void qM3C2Dialog::setCloud2Visibility(bool state)
 
 qM3C2Normals::ComputationMode qM3C2Dialog::getNormalsComputationMode() const
 {
+	// force vertical mode if needed in command line mode
+	if (!m_app)
+	{
+		if (getRequestedComputationMode() == qM3C2Normals::VERT_MODE)
+		{
+			return qM3C2Normals::VERT_MODE;
+			ccLog::Print("[M3C2] force vertical mode as requested in the parameter file");
```
dgirardeau, line 381: This will never be logged (as it's after 'return')

dgirardeau (changes requested):

dgirardeau:
Thnaks. And can you add something to the CHANGELOG file? (in the 'Bugs' section of the latest release)

dgirardeau (approved):

PR #2113, opened 2025-02-04
qLASIO : Initial support of COPC format (File Reading)

rjanvier:
This work was funded thanks to [Hobu Inc](https://github.com/hobuinc).

This is a first PR of a series to add support of [COPC](https://copc.io) file format in CloudCompare.

This PR cover only the reader part, the LOD/rendering part will come hopefully shortly after. For now, for end user, only very basic filtering operation on COPC data structure (level and spatial) filtering are available in a dedicated tab of the LAS windows. (LOD filling code is available and commented in the PR.)

The decision was made to minimize the impact on the CC Codebase, ensuring a fast adoption of COPC in CC. Therefore, we rely on LASZip's point-seeking capability (instead of a more natural "byte seeking"), similar to the approach used in the LAStools implementation.

As a result, this implementation face two limitations. First, LASZip accepts int64 for seeking but it internally narrows it to an uint32, thus we are limited to file of approx. 4,2B points. Addressing this would require a fix upstream or to vendor our own LASZip version with a fix. This is unfortunately a known issue of [LASZip](https://github.com/LASzip/LASzip/issues/76). Second, due to the use of kind global state in LASZip, it lacks parallel processing capabilities, preventing us from leveraging parallel chunk processing. 

Laz-perf would have been an ideal candidate to replace LASZip in qLASIO, but it would have introduced additional complexity in maintenance, both on the CC side (requiring another change of the LAS/LAZ library and the IO plugin) and within laz-perf itself, as it lacks support for FWF (point formats 4, 5, 9, and 10), which some CC users requires. 

In the current state of CC these limitations have no impact. Point clouds data structure in CC is also limited to unsigned int and CC point cloud internal representation is by now 100% _In core_, so it does not allows to handle very big point clouds.

[Copc samples.](https://copc.io/#example-data)

### DIFF
--- a/plugins/core/IO/qLASIO/include/CopcLoader.h
+++ b/plugins/core/IO/qLASIO/include/CopcLoader.h
@@ -222,8 +222,8 @@ namespace copc
 		CCVector3d     m_globalShift;
 		UnscaledExtent m_extent;
 
-		bool m_hasClippingConstraint;
-		bool m_hasMaxLevelConstraint;
+		bool m_hasClippingConstraint{false};
+		bool m_hasMaxLevelConstraint{false};
 
 		int32_t        m_maxLevelConstraint{0};
 		UnscaledExtent m_ClippingConstraint;
--- a/plugins/core/IO/qLASIO/include/CopcLoader.h
+++ b/plugins/core/IO/qLASIO/include/CopcLoader.h
@@ -21,7 +21,6 @@
 #include "CopcVlrs.h"
 #include "LasDetails.h"
 
-// CC
 // Qt
 #include <QFile>
 
@@ -43,8 +42,6 @@ using namespace LasDetails;
 
 namespace copc
 {
-	using MyBox = CCCoreLib::BoundingBoxTpl<double>;
-
 	class CopcLoader
 	{
 	  public: // methods
--- a/plugins/core/IO/qLASIO/include/CopcLoader.h
+++ b/plugins/core/IO/qLASIO/include/CopcLoader.h
@@ -1,0 +1,235 @@
+#pragma once
+
+// ##########################################################################
+// #                                                                        #
+// #                CLOUDCOMPARE PLUGIN: LAS-IO Plugin                      #
+// #                            COPCLoader                                  #
+// #                                                                        #
+// #  This program is free software; you can redistribute it and/or modify  #
+// #  it under the terms of the GNU General Public License as published by  #
+// #  the Free Software Foundation; version 2 of the License.               #
+// #                                                                        #
+// #  This program is distributed in the hope that it will be useful,       #
+// #  but WITHOUT ANY WARRANTY; without even the implied warranty of        #
+// #  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         #
+// #  GNU General Public License for more details.                          #
+// #                                                                        #
+// #                   COPYRIGHT: Hobu, Inc.                           	    #
+// #                                                                        #
+// ##########################################################################
+
+#include "CopcVlrs.h"
+#include "LasDetails.h"
+
+// CC
+// TODO LOD
+//#include "ccPointCloudLOD.h"
+
+// Qt
+#include <QFile>
+
+// Laszip
+#include <laszip/laszip_api.h>
+
+// System
+#include <cstdint>
+#include <functional>
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
+
+// CCCoreLib
+#include <CCGeom.h>
+#include <ccLog.h>
+
+using namespace LasDetails;
+
+namespace copc
+{
+	using MyBox = CCCoreLib::BoundingBoxTpl<double>;
+
+	class CopcLoader
+	{
+	  public: // methods
+		/// Constructor
+		///
+		/// because the constructor could fail, the user should use isValid() check before
+		/// using the instance.
+		explicit CopcLoader(const laszip_header* laszipHeader, const QString& fileName);
+
+		~CopcLoader() = default;
+
+		/// We do not want copy constructor and assigment for now.
+		CopcLoader(CopcLoader const&)            = delete;
+		CopcLoader& operator=(CopcLoader const&) = delete;
+
+		/// Enable and set the max level constraint
+		void setMaxLevelConstraint(uint32_t maxLevelConstraint)
+		{
+			m_hasMaxLevelConstraint = true;
+			m_maxLevelConstraint    = maxLevelConstraint;
+		}
+
+		/// Release the max level constraint and set the max level to the max level of the octree
+		void releaseMaxLevelConstraint()
+		{
+			m_hasMaxLevelConstraint = false;
+			m_maxLevelConstraint    = m_maxLevel;
+		}
+
+		/// Enable and set the clipping box constraint
+		void setClippingBoxConstraint(const LasDetails::UnscaledExtent& extent)
+		{
+			m_hasClippingConstraint = true;
+			m_ClippingConstraint    = extent;
+		}
+
+		/// Release the clipping Constraint and reset the clipping box to the root extent
+		void releaseClippingBoxConstraint()
+		{
+			m_hasClippingConstraint = false;
+			m_ClippingConstraint    = m_extent;
+		}
+
+		/// Returns an array containing the number of point at each level
+		/// (sequantially and from the root to the deepest depth)
+		const std::vector<uint64_t>& levelPointCounts() const
+		{
+			return m_levelPointCounts;
+		}
+
+		/// Returns the current full (non clipped) extent
+		const LasDetails::UnscaledExtent& extent() const
+		{
+			return m_extent;
+		}
+
+		/// Returns the current clippingExtent
+		const LasDetails::UnscaledExtent& clippingExtent() const
+		{
+			return m_ClippingConstraint;
+		}
+
+		/// Returns the Maximal number of Level (i.e; depth)
+		/// of the current COPC octree.
+		const int32_t maxLevel() const
+		{
+			return m_maxLevel;
+		}
+
+		/// Returns the validity of the viewer.
+		///
+		/// this is set by the constructor.
+		/// User of this class have the responsibility to check this getter
+		/// before any usage of the instance of the CopcLoader.
+		bool isValid() const
+		{
+			return m_isValid;
+		}
+
+		/// Set global shift. This is used by
+		void setGlobalShift(const CCVector3d& globalShift)
+		{
+			m_globalShift = globalShift;
+		}
+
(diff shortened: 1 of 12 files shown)


plugins/core/IO/qLASIO/src/LasIOFilter.cpp
```diff
@@ -44,9 +46,7 @@
 
 // System
 #include <memory>
-#include <numeric>
 #include <utility>
-
```
dgirardeau, line 49: no space/empty line?

plugins/core/IO/qLASIO/src/LasIOFilter.cpp
```diff
@@ -88,10 +88,10 @@ static CCVector3d GetGlobalShift(FileIOFilter::LoadParameters& parameters,
 
 LasIOFilter::LasIOFilter()
     : FileIOFilter({"LAS IO Filter",
-                    3.0f, // priority (same as the old PDAL-based plugin)
+                    3.0f, /// priority (same as the old PDAL-based plugin)
```
dgirardeau, line 91: to be reverted?

plugins/core/IO/qLASIO/src/LasIOFilter.cpp
```diff
@@ -197,11 +216,35 @@ CC_FILE_ERROR LasIOFilter::loadFile(const QString&  fileName,
...
+	// Tiling takes precedence over COPC
 	if (m_openDialog.action() == LasOpenDialog::Action::Tile)
 	{
 		return TileLasReader(laszipReader, fileName, m_openDialog.tilingOptions());
 	}
 
+	// Update chunksToReads according to the COPCLoader if needed
+	if (copcLoader)
+	{
+		const uint32_t copcUserDefinedMaxLevel = m_openDialog.copcDepthComboBox->currentData().toUInt();
```
dgirardeau, line 228: ideally the dialog class will return the value via a function ;)

plugins/core/IO/qLASIO/src/LasVlr.cpp
```diff
@@ -25,13 +26,13 @@
...
 // System
 #include <algorithm>
-#include <cstring>
 
 LasVlr::LasVlr(const laszip_header& header)
 {
 	const auto vlrShouldBeCopied = [](const laszip_vlr_struct& vlr)
 	{
-		return !LasDetails::IsLaszipVlr(vlr) && !LasDetails::IsExtraBytesVlr(vlr);
+		// Currently, we avoid copying back the COPC VLR (due to potential modifications to the file)
```
dgirardeau, line 34: Not sure to understand the comment (or at least, why it's here)

dgirardeau (commented):
Here is a first round! I haven't tried to compile/test it on my side yet...

plugins/core/IO/qLASIO/include/CopcLoader.h
```diff
@@ -0,0 +1,235 @@
...
+		///
+		/// The entire ChunkInterval vector/set is sorted by pointOffset to minimize seeking.
+		/// It is a vector of references because file reading could have side effect on ChunkIntervals.
+		/// Moreover an estimatedPointCount is returned in order to resize the CC point cloud.
+		/// estimatedPointCount is only an estimation if a Clipping constraint is set, because points in intersected node
+		/// can only be filtered during file reading. if no clipping constraint is set, the estimatedPointCount variable represent the true number of points.
+		void getChunkIntervalsSet(std::vector<std::reference_wrapper<ChunkInterval>>& sortedChunkIntervalSet, uint64_t& estimatedPointCount);
+
+		/// Recurse COPC octree and create a CC LOD matching this Hierachy
+		// TODO LOD
```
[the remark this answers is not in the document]
rjanvier: LOD rendering will be included in a second PR. I’m still working on it, and for now, everything related to LOD in the CopcLoader is commented out in this commit (as mentioned in the PR message). The LOD approach differs between CC and Layered/Nested Octree-based LOD (i.e the ones in Potree and Entwine (including COPC)).

For now, I tried to retain as much a possible of the current LOD we have in CC because I'm not adressing the LOD creation inside CC (just loading existing COPC structure). My goal is to be sure that users who still want to use LOD outside the ones provides by COPC file can continue to use our current LOD. Regarding COPC LOD rendering : for now, I came up with two rendering solutions: one that reuses much of our existing LOD or another that aligns better with the Layered LOD structure of COPC using VBO. VBO approach offer the best visual quality but is more complex to implement because I try to have reach genericity with both ccPointCloudLOD and the VBO manager.

plugins/core/IO/qLASIO/include/CopcLoader.h
```diff
@@ -0,0 +1,235 @@
...
+// #                                                                        #
+// #                   COPYRIGHT: Hobu, Inc.                           	    #
+// #                                                                        #
+// ##########################################################################
+
+#include "CopcVlrs.h"
+#include "LasDetails.h"
+
+// CC
+// TODO LOD
```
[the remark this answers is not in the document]
rjanvier: Thanks, finally removed (but please see the comment on the other commented part of the LOD and PR message for the "why" ;))

plugins/core/IO/qLASIO/src/LasVlr.cpp
rjanvier: We exclude COPC's VRL to prevent inconsistencies when rewriting the file, as the user might have altered the file within CC.

plugins/core/IO/qLASIO/include/CopcVlrs.h
```diff
@@ -0,0 +1,271 @@
...
+	// Hash function taken from PDAL.
+	// see https://github.com/PDAL/PDAL/blob/0c8e84521c040aa8a03ca848d32fb30e75a6e1ec/io/private/copc/Entry.hpp#L79-L89
+	template <>
+	struct hash<copc::VoxelKey>
+	{
+		std::size_t operator()(copc::VoxelKey const& key) const noexcept
+		{
+			std::hash<uint64_t> h;
+
+			uint64_t k1 = ((uint64_t)key.level << 32) | key.x;
```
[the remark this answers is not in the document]
rjanvier: thanks, done. Initially I now we prefer this kind of cast but initially I conserved the PDAL coding style.

plugins/core/IO/qLASIO/src/CopcLoader.cpp
```diff
@@ -0,0 +1,399 @@
...
+		}
+	}
+
+	void CopcLoader::generateChunktableIntervalsHierarchy(std::vector<Entry>& entries)
+	{
+		// Sort entries by offset to be able to get the first point of each chunk.
+		std::sort(std::begin(entries), std::end(entries), [](const Entry& a, const Entry& b)
+		          { return a.offset < b.offset; });
+		uint64_t starting_point = 0;
+		m_chunkIntervalsHierarchy.reserve(entries.size());
```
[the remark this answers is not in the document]
rjanvier: In the constructor?

plugins/core/IO/qLASIO/include/CopcLoader.h
```diff
@@ -0,0 +1,235 @@
...
+
+		/// Consistancy check: check if the root node exists
+		bool hasRoot()
+		{
+			return m_chunkIntervalsHierarchy.count(VoxelKey::Root());
+		}
+
+	  private: // static members
+		/// static members used to check if a file is a COPC
+		static const uint8_t         COPC_LAS_VERSION_MINOR{4};
```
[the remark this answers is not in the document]
rjanvier: yes, sorry, Initially CopcLoader.h was header only.

plugins/core/IO/qLASIO/src/LasVlr.cpp
dgirardeau: And it's hard to create a new one I guess?

plugins/core/IO/qLASIO/include/CopcLoader.h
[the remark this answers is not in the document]
dgirardeau: Yes, but this can stay in a separate commit / branch before you eventually push it ;)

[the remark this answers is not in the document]
rjanvier: yes, code was removed an will be part of the next PR/commit.

plugins/core/IO/qLASIO/include/CopcVlrs.h
```diff
@@ -0,0 +1,270 @@
...
+		double spacing{0.0};
+
+		/// Byte offeset in the LAS file
+		/// the root node is not necessary the node comming at first
+		/// in the COPC EVRL, so it's not the start of the COPC
+		/// EVRL Reccord.
+		uint64_t root_hier_offset{0};
+		uint64_t root_hier_size{0};
+
+		/// TODO
```
dgirardeau, line 95: TODO?

plugins/core/IO/qLASIO/src/CopcLoader.cpp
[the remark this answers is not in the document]
dgirardeau: Above then?

Or don't do this in a constructor (if I understood you correctly?)

plugins/core/IO/qLASIO/src/LasIOFilter.cpp
```diff
@@ -470,12 +578,31 @@ CC_FILE_ERROR LasIOFilter::loadFile(const QString&  fileName,
...
+	// With the copcLoader, we shrink the point cloud to take into account the point that could be filtered in the loading process
+	// We have to do that before the LOD construction because shrinkTofit clears the LOD.
+	if (copcLoader)
+	{
+		pointCloud->shrinkToFit();
+	}
+
 	LasMetadata::SaveMetadataInto(*laszipHeader, *pointCloud, availableExtraScalarFields);
 
+	// TODO LOD
```
dgirardeau, line 590: to be moved to the next PR?

plugins/core/IO/qLASIO/src/LasVlr.cpp
rjanvier: if we modify the number of points within CC (and the BBox in some extent), we have to (re)create a Nested Octree like data structure. it's basically what PotreeConverter and Entwine/Untwine generate. It should not be too hard to implement a "basic" In core indexer in CC but it is out of my scope in the short term.

plugins/core/IO/qLASIO/include/CopcVlrs.h
rjanvier: values are unpacked from the file but it's not used anywhere. I wonder if there is a specific use cases for these values since If I understand correctly we compute our own shift in CC.

dgirardeau (approved):

dgirardeau (approved):

PR #2122, opened 2025-02-18
Remove deprecated QGLBuffer in favor of QOpenGLBuffer

rjanvier:
- QT documentation state that `QGLBuffer` is deprecated and `QOPenGLBuffer` has better performance (I'm pretty sure we won't notice any improvement thought :D). 
While we already use `QOpenGLBuffer` in some parts of our code, it has not yet been used the ccPointCloud class. This PR addresses this by making the necessary changes.

### DIFF
--- a/libs/qCC_db/include/ccPointCloud.h
+++ b/libs/qCC_db/include/ccPointCloud.h
@@ -31,7 +31,7 @@
 #include "ccWaveform.h"
 
 //Qt
-#include <QGLBuffer>
+#include <qopenglbuffer.h>
 
 class ccScalarField;
 class ccPolyline;
@@ -876,7 +876,7 @@ protected: // VBO
 	//! Init/updates VBOs
 	bool updateVBOs(const CC_DRAW_CONTEXT& context, const glDrawParams& glParams);
 
-	class VBO : public QGLBuffer
+	class VBO : public QOpenGLBuffer
 	{
 	public:
 		int rgbShift;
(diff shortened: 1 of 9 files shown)


libs/qCC_db/include/ccPointCloud.h
```diff
@@ -31,12 +31,12 @@
 #include "ccWaveform.h"
 
 //Qt
-#include <QGLBuffer>
+#include <qopenglbuffer.h>
```
dgirardeau, line 34: would `#include <QOpenGLBuffer>` work?

libs/qCC_db/src/ccPointCloud.cpp
```diff
@@ -5585,15 +5585,17 @@ bool ccPointCloud::updateVBOs(const CC_DRAW_CONTEXT& context, const glDrawParams
 		{
 			int chunkSize = static_cast<int>(ccChunk::Size(chunkIndex, m_points));
 
+			VBO * currentVBO = m_vboManager.vbos[chunkIndex];
```
dgirardeau, line 5588:
```suggestion
			VBO* currentVBO = m_vboManager.vbos[chunkIndex];
```

libs/qCC_db/include/ccPointCloud.h
rjanvier: yes, I think so!

libs/qCC_db/src/ccPointCloud.cpp
rjanvier: will be changed, thanks.

dgirardeau (approved):

PR #2130, opened 2025-03-01
Refactor sub sample action

tmontaigu:


qCC/ccCommandLineCommands.cpp
```diff
@@ -1324,70 +1324,50 @@ bool CommandSubsample::process(ccCommandLineInterface& cmd)
...
 		{
 			cmd.print(QObject::tr("\tProcessing cloud %1").arg(!desc.pc->getName().isEmpty() ? desc.pc->getName() : "no name"));
 
-			if (isPercent)
-			{
-				size_t nrOfPoints = desc.pc->size();
-				count = static_cast<unsigned>(ceil(nrOfPoints * percent / 100));
-				cmd.print(QObject::tr("\tOutput points: %1 * %2% = %3").arg(nrOfPoints).arg(percent).arg(count));
+			CCCoreLib::ReferenceCloud* refCloud = nullptr;
+			try {
```
dgirardeau, line 1353: can you move the { symbols to the next lines? (as per CC's style).

(by the way, I'm not sure in which state your attempt to activate a global .clangformat file is?)

qCC/ccCommandLineCommands.cpp
```diff
@@ -1324,70 +1324,50 @@ bool CommandSubsample::process(ccCommandLineInterface& cmd)
...
+				refCloud = DoSubSample(params, desc.pc, cmd.progressDialog());
+			} catch (const std::exception& e) {
```
dgirardeau, line 1355: same here and below

qCC/ccCommandLineCommands.cpp
```diff
@@ -1415,30 +1396,34 @@ bool CommandSubsample::process(ccCommandLineInterface& cmd)
...
 				{
-					bool validMin = false;
-					sfMinSpacing = cmd.arguments().takeFirst().toDouble(&validMin);
-					bool validMax = false;
-					sfMaxSpacing = cmd.arguments().takeFirst().toDouble(&validMax);
-					if (!validMin || !validMax || sfMinSpacing < 0 || sfMaxSpacing < 0)
-					{
-						return cmd.error(QObject::tr("Invalid parameters: Two positive decimal number required after '%1'").arg(OPTION_USE_ACTIVE_SF));
-					}
+					return cmd.error(QObject::tr("Missing parameters: Two positive decimal number required after '%1'").arg(OPTION_USE_ACTIVE_SF));
```
dgirardeau, line 1401:
```suggestion
					return cmd.error(QObject::tr("Missing parameters: Two positive decimal numbers required after '%1'").arg(OPTION_USE_ACTIVE_SF));
```

qCC/ccCommandLineCommands.cpp
```diff
@@ -1415,30 +1396,34 @@ bool CommandSubsample::process(ccCommandLineInterface& cmd)
...
-				else
+
+				bool validMin = false;
+				sfMinSpacing = cmd.arguments().takeFirst().toDouble(&validMin);
+				bool validMax = false;
+				sfMaxSpacing = cmd.arguments().takeFirst().toDouble(&validMax);
+				if (!validMin || !validMax || sfMinSpacing < 0 || sfMaxSpacing < 0)
 				{
-					return cmd.error(QObject::tr("Missing parameters: Two positive decimal number required after '%1'").arg(OPTION_USE_ACTIVE_SF));
+					return cmd.error(QObject::tr("Invalid parameters: Two positive decimal number required after '%1'").arg(OPTION_USE_ACTIVE_SF));
```
dgirardeau, line 1410:
```suggestion
					return cmd.error(QObject::tr("Invalid parameters: Two positive decimal numbers required after '%1'").arg(OPTION_USE_ACTIVE_SF));
```

qCC/ccCommandLineCommands.cpp
```diff
@@ -1451,93 +1436,41 @@ bool CommandSubsample::process(ccCommandLineInterface& cmd)
...
 				cmd.print(QObject::tr("\t'Use active SF' disabled. Falling back to constant spacing."));
 			}
 
-			CCCoreLib::ReferenceCloud* refCloud = CCCoreLib::CloudSamplingTools::resampleCloudSpatially(desc.pc, static_cast<PointCoordinateType>(step), modParams, nullptr, cmd.progressDialog());
-			if (!refCloud)
-			{
-				return cmd.error("Subsampling process failed!");
+			CCCoreLib::ReferenceCloud* refCloud = nullptr;
+			try {
+				refCloud = DoSubSample(params, desc.pc, cmd.progressDialog());
```
dgirardeau, line 1458: Maybe you could catch the exception inside the DoSubSample and return nullptr. Because here it's not obvious by design that you'll get a non-null 'refCloud' pointer if no exception is raised. Which means I would still test refCloud against nullptr afterwards...

qCC/ccCommandLineCommands.cpp
```diff
@@ -1767,6 +1680,45 @@ bool CommandSubsample::process(ccCommandLineInterface& cmd)
 	return true;
 }
 
+bool CommandSubsample::ReplaceAndSaveCloud(ccCommandLineInterface& cmd, CLCloudDesc& orignalDesc, ccPointCloud *newPointCloud, const SubsamplingParams& params) {
```
dgirardeau, line 1683:
```suggestion
bool CommandSubsample::ReplaceAndSaveCloud(ccCommandLineInterface& cmd, CLCloudDesc& orignalDesc, ccPointCloud* newPointCloud, const SubsamplingParams& params) {
```

dgirardeau: And maybe you could pass a reference to a 'newPointCloud' pointer  (ccPointCloud*& newPointCloud) so as to set it to nullptr after you delete it. So that the function caller has an updated (null) pointer.

qCC/ccSubsamplingDlg.cpp
```diff
@@ -66,109 +69,153 @@ ccSubsamplingDlg::~ccSubsamplingDlg()
 	delete m_ui;
 }
 
-CCCoreLib::ReferenceCloud* ccSubsamplingDlg::getSampledCloud(ccGenericPointCloud* cloud, CCCoreLib::GenericProgressCallback* progressCb/*=nullptr*/)
-{
+CCCoreLib::ReferenceCloud* ccSubsamplingDlg::getSampledCloud(ccGenericPointCloud* cloud, CCCoreLib::GenericProgressCallback* progressCb/*=nullptr*/) const {
```
dgirardeau, line 72: just as a reminder: can you move the { symbols to the next line 👼

qCC/ccSubsamplingDlg.cpp
```diff
@@ -66,109 +69,153 @@ ccSubsamplingDlg::~ccSubsamplingDlg()
...
+	return modulationEnabled;
+}
+
+
+CCCoreLib::ReferenceCloud* DoSubSample(const SubsamplingParams &params, ccGenericPointCloud *cloud, CCCoreLib::GenericProgressCallback *progressCb /*=nullptr*/) {
 	if (!cloud || cloud->size() == 0)
 	{
-		ccLog::Warning("[ccSubsamplingDlg::getSampledCloud] Invalid input cloud!");
-		return nullptr;
+		throw std::runtime_error("Invalid input cloud!");
```
dgirardeau, line 107: In CloudCompare, I prefer not too propagate exceptions, and manage errors as soon as they happen.

qCC/ccSubsamplingDlg.h
```diff
@@ -39,6 +43,98 @@ namespace Ui
...
+		double randomPercent;
+		//! Method::Octree
+		unsigned char octreeLevel;
+		//! Method::spatial
+		Spatial spatial;
+	};
+	Method method;
+
+
+	//explicit SubsamplingParams(const Spatial &params_) : spatial(params_), method(Method::Spatial) {}
```
dgirardeau, line 110: to be removed?

qCC/ccCommandLineCommands.cpp
tmontaigu: More or less abandoned for now, I should try it again

I think I had some problems between different clang version and the fact that some dependencies are vendorer, but since its been a few months I don't actually remember the specifics

qCC/ccCommandLineCommands.cpp
tmontaigu: If we keep exceptions from DoSubsample then the function should return a non pointer ReferenceCloud i think, otherwise if we remove exceptions then it should return a ptr that is null on error

qCC/ccSubsamplingDlg.cpp
tmontaigu: I expected a comment like that :grin: 

I'm starting to  think that to do this we may need to introduce some kind "UserFeedback" interface (and have CmdUserFeeback and GUIUserFeeback implementations), as probably, for warning ccWarning is fine, but for errors we probably want to display differently between GUI and cmd (and maybe later python scripts)

qCC/ccSubsamplingDlg.h
tmontaigu: likely

tmontaigu:
Also I'm not entirely satisfied with how the refactor is in this PR as the GUI params are not as capable as the cmd version, in my ideal world I'd like the cmd line to be (as much as it can be) just about parsing args from cmdline into a params struct and then call a function, and the GUI to get params from a dialog, populate som params struct and call the same function as the cmdline

qCC/ccCommandLineCommands.cpp
dgirardeau: Yes, that's what most functions would do in CC, returning a nullptr in case of error

qCC/ccSubsamplingDlg.cpp
dgirardeau: Or maybe simply add some static parameters/methods to change the behavior of ccLog::Error in command line mode? (or Python)

tmontaigu: I'm confused with how logging is handled between regular GUI mode, cmdline non-silent, cmdline silent

I see that commandline interface has its own methods for logging, that the commands use as `cmd.print`, `cmd.error`. 

To me only `ccLog::` should exists and be used, and then the actual instance  would handle GUI vs CMD mode non-SILENT vs CMD mode SILENT, either by having different subclasses for ccLog or ccConsole handling those via some settings

Looking at the code, the cmd.error and co seems to forward to ccConsole, so it seems things are correctly made (except I could not find where the message box for Error level is disabled in silent mode) but the cmd.print family of functions do not really provide that much value and at least for me are creating some confusion

tmontaigu: Is this something we should try to adress little by little ?

dgirardeau: Ah yes, don't hesitate!

PR #2136, opened 2025-03-13
Don't ignore empty csv fields

rotu:
Previously, an empty CSV field would cause columns in a CSV to get misaligned. E.g.
```csv
x,y
,42
```
would be the same as:
```csv
x,y
42
```

That is, "42" would be treated as if it were in the "x" column, not the "y" column. Now, empty fields are not discarded.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -204,6 +204,7 @@ Bug fixes:
 	- The circular cursor of the 'Cloud layers' and 'Compass' plugins was not displayed at the right position on high DPI screens
 	- The Compass plugin was not transferring the Global Shift & Scale information from the cloud to the generated planes or polylines
 	- UHD screens were not properly supported (rotation center picking with double click, entity selection with a rectangle, etc.)
+	- ASCII cloud file import will now respect empty fields instead of shifting all following columns left
 
 v2.13.2 (Kharkiv) - (06/30/2024)
 ----------------------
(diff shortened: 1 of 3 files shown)


dgirardeau:
Thanks,

2 questions:
- what happens if the separator is a whitespace (space char.) - I don't remember if this is something that can be used or not with this particular filter
- can you update the CHANGELOG file as well?

rotu:
@dgirardeau Unfortunately, I can't build this locally to check. I'm getting:

```
cmake -DCMAKE_PREFIX_PATH=D:\Qt\5.15.2\msvc2019_64 ..
CMake Error at CMakeLists.txt:3 (project):
  Generator

    Visual Studio 17 2022

  could not find specified instance of Visual Studio:

    C:/Program Files (x86)/Microsoft Visual Studio/2022/BuildTools


-- Configuring incomplete, errors occurred!
```

rotu:
I was a little confused.

The changes I meant to make belong in `AsciiFilter.cpp` and `AsciiOpenDlg.cpp`. I'm not sure what `CSVMatrixFilter.cpp` is even for, and I'll revert my change to that file.

> what happens if the separator is a whitespace (space char.) - I don't remember if this is something that can be used or not with this particular filter

As for `AsciiFilter`, we call [`QString::simplified()`](https://doc.qt.io/qt-5/qstring.html#simplified) to coalesce internal whitespace before splitting, so the import result is unaffected with repeated whitespace delimiters (and leading/trailing whitespace is ignored).

> can you update the CHANGELOG file as well?

Done.

dgirardeau:
Okay, seems reasonable! I'll merge it.

dgirardeau (approved):

PR #2149, opened 2025-04-02
Fix duplicate assignment of LAS extra scalar fields

### LINKED ISSUE #2144: Las save options dialog duplicates all extra fields from las file
**Describe the bug**
Las save options dialog duplicates all extra fields loaded from las file when "Save all remaining scalar fields as Extra fields / EB-VLRs" is enabled

**How to reproduce**
- Import a las file with extra fields, enable all extra fields (I tested a las version 1.4 point format 3 file with three extra fields)
- click on imported point cloud and click "Save"
- Set output format to "*.las" then bring up "LAS Save Options" dialog
- Goto "Standard LAS fields" tab, "Save all remaining scalar fields as Extra fields / EB-VLRs" is checked by default
- Goto "Extra fields (Extra Bytes VLRs)" tab and check extra fields to be exported
- Every extra field from las file is duplicated, resulting in the exported file having 2x number of extra fields

Image

**Expected behaviour**
Extra fields from las file won't be duplicated

**Your environment**
 - CC Version (or git hash/tag): v2.14.alpha (Feb 23 2025)
 - OS & Version: Windows 11 x64

tmontaigu:
When saving a LAS file that originally had extra scalar fields we could end up creating duplicates of them when saving.

This bug was due to the dialog auto assigning non standard field name to new scalar extra scalar field with some default options. This was done every time the point format changed (but here it was properly handled).

What was missing is that when we tell the dialog, which extra scalar field we know of thanks to the meta data attached to the cloud, we did not clear the default created extra scalar fields, so we ended up with duplicates.

Fixes #2144

### DIFF
--- a/plugins/core/IO/qLASIO/src/LasSaveDialog.cpp
+++ b/plugins/core/IO/qLASIO/src/LasSaveDialog.cpp
@@ -508,11 +508,21 @@ void LasSaveDialog::setExtraScalarFields(const std::vector<LasExtraScalarField>&
 		return;
 	}
 
+	// We received extra scalar fields spec saved from the file's metadata
+	// But we may have already default-assigned some fields as extra scalar fields
+	// so we have to clear them
+	unassignDefaultFields();
+
 	for (const LasExtraScalarField& field : extraScalarFields)
 	{
 		auto* card = createCard();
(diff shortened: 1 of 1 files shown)


dgirardeau (approved):

dgirardeau:
Thanks for the quick fix!

Issue #2110, opened 2025-01-28
Cursor is way too small


themightyoarfish:
**Describe the feature you would like**

The cursor to set the pivot is way too small on a retina display, oftentimes completely invisible

**Describe alternatives you've considered**
I compile myself, so I could change it, if only i knew where to look

**Is your feature request related to a problem? Please describe.**
Ideally, a setting to modify the size would be great

dgirardeau:
Hi,

We use Qt to manage the curosr. See https://doc.qt.io/qt-6/qcursor.html and https://doc.qt.io/qt-6/qwidget.html#cursor-prop

(in effect it's called here https://github.com/CloudCompare/CloudCompare/blob/master/libs/qCC_glWindow/include/ccGLWindow.h#L44 and here https://github.com/CloudCompare/CloudCompare/blob/master/libs/qCC_glWindow/include/ccGLWindowStereo.h#L48 depending on whether you use Stereo display or not, but it's the same mechanism offered by Qt)

I would expect Qt to use by default the system default cursor size. But if you want to override that, it seems that it's possible to define a 48x48 or 64x64 cursor (assuming the default is 32x32). But I don't know how portable this is.

themightyoarfish:
The default size on my machine is like this: 

<img width="1608" alt="Image" src="https://github.com/user-attachments/assets/075d7061-d24b-4f68-aa88-3739148cc036" />

especially when the RGB cloud is grey in areas, it's invisible

dgirardeau:
Oh, are you referring to the 'middle screen cross'? 😅  In this case, it's not the mouse cursor, nor something managed by the OS, but a simple cross displayed in 2D over the OpenGL context.

It's very simple to change:

see  https://github.com/CloudCompare/CloudCompare/blob/master/libs/qCC_glWindow/src/ccGLWindowInterface.cpp#L4120

Instead of using a hardcoded width (currently 10px), you could multiply the default size by the result of the `getDevicePixelRatio()` method:

```
const int retinaScale = getDevicePixelRatio();
float centerCrossLength = retinaScale * CC_DISPLAYED_CENTER_CROSS_LENGTH;

glFunc->glBegin(GL_LINES);
glFunc->glVertex3f(0.0f, -centerCrossLength , 0.0f);
glFunc->glVertex3f(0.0f, centerCrossLength , 0.0f);
glFunc->glVertex3f(-centerCrossLength , 0.0f, 0.0f);
[...]

PR #2162, opened 2025-05-07
improve crosshair visibility

themightyoarfish:
addresses https://github.com/CloudCompare/CloudCompare/issues/2110

### DIFF
--- a/libs/qCC_glWindow/src/ccGLWindowInterface.cpp
+++ b/libs/qCC_glWindow/src/ccGLWindowInterface.cpp
@@ -4110,17 +4110,20 @@ void ccGLWindowInterface::drawCross()
 	ccQOpenGLFunctions* glFunc = functions();
 	assert(glFunc);
 
+	const float pixel_ratio=getDevicePixelRatio();
+
 	//force line width
 	glFunc->glPushAttrib(GL_LINE_BIT);
-	glFunc->glLineWidth(1.0f);
+	glFunc->glLineWidth(2.0f * pixel_ratio);
 
 	//cross OpenGL drawing
 	glColor4ubv_safe<ccQOpenGLFunctions>(glFunc, ccColor::lightGrey);
 	glFunc->glBegin(GL_LINES);
-	glFunc->glVertex3f(0.0f, -CC_DISPLAYED_CENTER_CROSS_LENGTH, 0.0f);
-	glFunc->glVertex3f(0.0f, CC_DISPLAYED_CENTER_CROSS_LENGTH, 0.0f);
-	glFunc->glVertex3f(-CC_DISPLAYED_CENTER_CROSS_LENGTH, 0.0f, 0.0f);
-	glFunc->glVertex3f(CC_DISPLAYED_CENTER_CROSS_LENGTH, 0.0f, 0.0f);
+	const float cross_length = pixel_ratio * CC_DISPLAYED_CENTER_CROSS_LENGTH;
+	glFunc->glVertex3f(0.0f, -cross_length, 0.0f);
+	glFunc->glVertex3f(0.0f, cross_length, 0.0f);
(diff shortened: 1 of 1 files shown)


libs/qCC_glWindow/src/ccGLWindowInterface.cpp
```diff
@@ -4110,17 +4110,20 @@ void ccGLWindowInterface::drawCross()
 	ccQOpenGLFunctions* glFunc = functions();
 	assert(glFunc);
 
+	const float pixel_ratio=getDevicePixelRatio();
+
 	//force line width
 	glFunc->glPushAttrib(GL_LINE_BIT);
-	glFunc->glLineWidth(1.0f);
+	glFunc->glLineWidth(2.0f * pixel_ratio);
```
dgirardeau, line 4117: While I would agree with 1 tiles the pixel ration here you also make the cross 'fatter' for everyone, even for people without a high DPI screen.

libs/qCC_glWindow/src/ccGLWindowInterface.cpp
```diff
@@ -4110,17 +4110,20 @@ void ccGLWindowInterface::drawCross()
 	ccQOpenGLFunctions* glFunc = functions();
 	assert(glFunc);
 
+	const float pixel_ratio=getDevicePixelRatio();
```
dgirardeau, line 4113: Mind that we use camel case (pixelRatio)

libs/qCC_glWindow/src/ccGLWindowInterface.cpp
```diff
@@ -4110,17 +4110,20 @@ void ccGLWindowInterface::drawCross()
...
+	glFunc->glLineWidth(2.0f * pixel_ratio);
 
 	//cross OpenGL drawing
 	glColor4ubv_safe<ccQOpenGLFunctions>(glFunc, ccColor::lightGrey);
 	glFunc->glBegin(GL_LINES);
-	glFunc->glVertex3f(0.0f, -CC_DISPLAYED_CENTER_CROSS_LENGTH, 0.0f);
-	glFunc->glVertex3f(0.0f, CC_DISPLAYED_CENTER_CROSS_LENGTH, 0.0f);
-	glFunc->glVertex3f(-CC_DISPLAYED_CENTER_CROSS_LENGTH, 0.0f, 0.0f);
-	glFunc->glVertex3f(CC_DISPLAYED_CENTER_CROSS_LENGTH, 0.0f, 0.0f);
+	const float cross_length = pixel_ratio * CC_DISPLAYED_CENTER_CROSS_LENGTH;
```
dgirardeau, line 4122:
```suggestion
	const float crossLength = pixelRatio * CC_DISPLAYED_CENTER_CROSS_LENGTH;
```

libs/qCC_glWindow/src/ccGLWindowInterface.cpp
```diff
@@ -4994,7 +4997,10 @@ void ccGLWindowInterface::draw3D(CC_DRAW_CONTEXT& CONTEXT, RenderingParams& rend
 		&& (!renderingParams.useFBO || !m_activeGLFilter))
 	{
 		setStandardOrthoCenter();
+		glFunc->glPushAttrib(GL_DEPTH_BUFFER_BIT);
```
dgirardeau, line 5000: And here again you are changing the behavior for everyone. This cross has always been displayed this way for more than 15 years... Maybe add an option to control this behavior? (that's heavier, but you can see how the 'show middle screen cross option' display option is managed currently, and duplicate this behavior).

libs/qCC_glWindow/src/ccGLWindowInterface.cpp
jameelch7788: Pk95BPUN5010228028700011

dgirardeau:
Without any reaction, I'm going to close this PR.

PR #2168, opened 2025-05-16
Add shortcut settings dialog

tmontaigu:
This adds a new entry to the file menu which opens a dialog that allows users to edit shortcuts for actions.

The actions available are the ones of the main window, but it could be extended to also include other actions like the ones coming from plugins.

When a shortcut is edited, it is saved to QSettings so it is restored next time CloudCompare is opened.

Although QAction can support multiple shortcut, the UI only allows one, which should be enough for a starting point.

The dialog tries to prevent the user from setting shortcut which could conflict

Improvements for later could be:
- Add more actions (especially plugins)
- Add import/export shortcuts to/from a text file to be able to share / save
- Have default from which users could reset their state and start clean
- Add a search box

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,8 @@ New features:
 		- choice is now persistent, and will be reactivated when running CC again, or creating a new 3D view
 		- currently ignored by 3D mice and controllers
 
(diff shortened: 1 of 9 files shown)


dgirardeau:
Very nice!

Can you update the CHANGELOG.md file?

PR #2209, opened 2025-09-01
Fix native dialogs not showing up on macOS

### LINKED ISSUE #2152: Open and Save features completely not working in CloudCompare
**Describe the bug**

After upgrading to macos 15.4 I am suddenly unable to open or save anything. The open, save, and save as icons are not working (clicking on them does nothing). As well as the keyboard shortcuts (cmd+o or cmd+s or cmd+shift+s). I tried the binaries [here](https://www.simulation.openfields.fr/index.php/cloudcompare-downloads/7-cloudcompare-macos-binaries) and nothing is working. I also tried building the 2.14.alpha from source and still it did not work. 

**How to reproduce**

Steps to reproduce the behaviour:
1. Go to 'CloudCompare 2.13.0 and higher'
2. Click on 'Open icon' 
3. See error (i.e. nothing happens)

**Expected behaviour**

The icons are supposed to open the dialog boxes which then can be used to select to load / save files.

**Additional context**

Everything was working perfectly before the update.

**Your environment**
 - CC Version (or git hash/tag): 2.13.2 (mainly) but as stated tested before
 - OS & Version: macos 15.4
 - Graphics card: M4 pro
 - Qt Version (if compiling): tried with qt5 from brew, not sure which version exactly

tmontaigu:
Adding a GUI identifier to the bundle seems to fix it

Fixes #2152

### DIFF
--- a/qCC/Mac/CMakeLists.txt
+++ b/qCC/Mac/CMakeLists.txt
@@ -2,6 +2,7 @@
 if( APPLE )
 	set_target_properties( ${PROJECT_NAME} PROPERTIES
 		INSTALL_RPATH "@executable_path/../Frameworks"
+		MACOSX_BUNDLE_GUI_IDENTIFIER "org.cloudcompare.CloudCompare"
 		MACOSX_BUNDLE_INFO_PLIST ${CMAKE_CURRENT_SOURCE_DIR}/CloudCompare.plist
 		MACOSX_BUNDLE_ICON_FILE cc_icon.icns
 		MACOSX_BUNDLE_SHORT_VERSION_STRING "${PROJECT_VERSION}"


dgirardeau:
Nice! I'll let @prascle review and maybe create a new package?

prascle:
I think I've already tried a similar approach without success, but I'll try again. Maybe I made a mistake during my previous attempt.
It won't be possible until next week, as I don't have my Mac with me...

tmontaigu:
At least it worked great on the mac I tested

SynSab:
Hi, so sorry but how do I use this fix? Downloaded a version of CloudCompare earlier today and am having this bug. Would LOVE to be able to save stuff on my Mac.

prascle:
The fix works, see my latest version [here](https://www.simulation.openfields.fr/index.php/cloudcompare-downloads/7-cloudcompare-macos-binaries/123-cloudcompare-20250913-with-native-dialogs-without-python-plugin)!

SynSab:
Having some trouble with GCAL library support in this build when creating meshes. Not sure how that would've been lost. Can others reproduce?

PR #2210, opened 2025-09-02
Update ccPointCloud.cpp

p-leroy:
if any scalar field of the entity has a shift, set the file version properly


dgirardeau (approved):
Good catch!

PR #2213, opened 2025-09-10
Delete confirmaton dialog

farmakis:
I added a confirmation dialog in the DeleteSelectedEntities method in the simplest possible way, to avoid accidental deletions.
In case it's useful for the project, let me know if it should also have its own deleteDlg cpp and ui files like the other dialogs.


qCC/db_tree/ccDBRoot.cpp
```diff
@@ -32,6 +32,7 @@
 #include <QSettings>
 #include <QStandardItemModel>
 #include <QTreeView>
+#include <qcheckbox.h>
```
dgirardeau, line 35:
```suggestion
#include <QCheckbox>
```

qCC/db_tree/ccDBRoot.cpp
```diff
@@ -544,6 +546,35 @@ void ccDBRoot::deleteSelectedEntities()
...
 
+	// confirmation
+	ccOptions options = ccOptions::Instance();
+
+	if (options.confirmDelete)
+	{
+		QMessageBox confDial(MainWindow::TheInstance());
+		confDial.setIcon(QMessageBox::Question);
+		confDial.setWindowTitle(tr("Delete Selected Entities"));
+		confDial.setText(tr("Are you sure ?"));
```
dgirardeau, line 557:
```suggestion
		confDial.setText(tr("Are you sure?"));
```

dgirardeau (commented):
You'll have to fix this branch history also, it's very weird. Maybe you can squash all the commits?

PR #2219, opened 2025-09-23
Compile with Qt6 (and Visual Studio 2022)

dgirardeau:
# TODO
- Windows does not seeem to exit cleanly
- Gamepad support (QtGamepad does not exist in qt6, but someone did a port)
- qPCV plugin
- all plugins relying on flann (on Windows at least)


dgirardeau:
Replaced by PR #2237

PR #2222, opened 2025-09-29
MacOS Compat with QT6

rjanvier:
- Use a workaround for **hopefully** "portable" custom FBOs. It's ok both at compile and runtime on macOS but should be tested at runtime for other systems.1
- Update libE57 format to the latest version (v3.2).
- Update the pixi file to the current state of the macOS port (remove qMasonry and 3DFin for now).


rjanvier:
Validated on both Linux and Windows. It seems there is an issue on Linux on some configurations (proprietary NVIDIA driver + Wayland), but it's independent of this PR.

PR #2231, opened 2025-10-13
Update mainwindow.cpp

p-leroy:
ccRoot is automatically deleted by his parent as created with: m_ccRoot = new ccDBRoot(m_UI->dbTreeView, m_UI->propertiesTreeView, this);


p-leroy:
In debug mode, this was the source of an error+crash of QtCreator on my side, just an error with Visual Studio.

dgirardeau:
Can you either add a comment in the code to explain why these lines are commented? Or simply remove the code.

p-leroy:
Oh yes, I will remove the lines, sorry.

PR #2246, opened 2025-11-10
Remove oculus support

rjanvier:


libs/qCC_db/extern/CCCoreLib
dgirardeau: Is it on purpose? (updating the CCCoreLib submodule)

rjanvier: It seems ok now, thanks.

PR #2252, opened 2025-11-21
Temporary hack to allow versions compiled with Qt 6 to read BIN files saved with a previous version of CC compiled with Qt 5

dgirardeau:


dgirardeau:
Any better solution is welcome 😓

PR #2262, opened 2025-12-09
The user can now see and edit the angular ranges when editing or creating a TLS sensor

dgirardeau:


tpwrules (approved):
Seems to work well. I fooled around with it a little bit and it does what it says.

One oddity I found is that the behavior of auto-detecting the size when using 0 for all four params only works in the create dialog. On the edit dialog it just gives an error that the depth buffer is void. But I'll leave that up to you if you want to change it.

dgirardeau:
Good catch. By design it's only supported upon creation of the sensor. I'll hide the message in the other case.

PR #2265, opened 2025-12-20
Fix cloud sampling failure when the density for a single object is too low

omarelladen:
When sampling points on a mesh using the CLI, the execution should not terminate when ccGenericMesh::samplePoints returns nullptr. This is specially important when an object of the mesh is too small to have its cloud generated (in this case sampledCloud->size() == 0), so the sampling of rest of the objects should continue, as it does in the GUI.


omarelladen:
I added the same error message as in the GUI

dgirardeau (approved):
Looks good, thanks for the update!

PR #2268, opened 2025-12-31
Improve speed of cloud merge operation on CLI

omarelladen:
I noticed that the cloud merge operation performance was very bad, so I applied the improvement that was done for the GUI operation in this commit: https://github.com/CloudCompare/CloudCompare/commit/711e9e6f8fb565e467ac63263c956ee623fc1744 (reserve enough space for final cloud at start)

To be consistent with the GUI implementation, I also changed some variable names
and used the "append" method (I was not sure if it needed to have the "ignoreChildren" as false so I kept it like this).


dgirardeau (approved):

dgirardeau:
Good catch! Thanks for the contribution (and Happy new year!)

Issue #2278, opened 2026-02-04
Can't install on Windows ARM64


sammo2828:
**Describe the bug**

The installer is restricted with ArchitecturesAllowed=x64.
Please rebuild with x64compatible to support Windows ARM64 x64 emulation.

**How to reproduce**

1. Attempt to run the provided Inno Setup installer on a Windows ARM64 device.
2. Launch the setup executable.
3. Observe that the installer exits with a message indicating the system architecture is unsupported due to ArchitecturesAllowed=x64.

**Expected behaviour**

The installer should allow execution on ARM64 versions of Windows because they support x64 emulation, meaning systems that are x64-compatible should not be blocked by the ArchitecturesAllowed setting.

**Your environment**

 - OS & Version: Windows 11 ARM64

PR #2283, opened 2026-02-16
Update CloudCompare_x64_InnoSetup.iss

sammo2828:
x64compatible should be used instead of x64os, because x64compatible allows x64 apps to be installed on Arm64 Windows 11 systems as well

#2278


dgirardeau:
Thanks!

PR #2284, opened 2026-02-18
Fixed bug on deletion of the lattice with control points selected

gfr-misha:


dgirardeau:
Wow, looks very interesting!

Before I review/test it, would it be possible to move the plugin files under the 'plugins/core/Standard' folder? ('plugins/example' is really meant to store the examples only).

gfr-misha:
Hi! This PR was created by mistake, actually, but it would be cool if you found it interesting and it can be merged into main. I moved it into the standard folder; however, it might be better to include it as a submodule.

Basically, what it does is that it tries to replicate a behaviour found in Blender where a lattice modifier can be applied to a cloud to apply a non-rigid transformation. This is useful for situations where the point cloud from SLAM comes a bit off, but there are control clouds (from GPS) available, which can be used for alignment. 

Also, this PR has changes to CI and workflows, which are probably not needed here. 

Anyway, thanks for your attention and lmk what you think :)

dgirardeau:
Ok, I confirm that it compiles and the lattice is visible. I'll let you work on the next step quietly (don't hesitate to ask questions though).

And indeed, eventually this will have to be moved to a submodule.

PR #2289, opened 2026-03-06
fix: correctly render rect labels when rendering viewport to image

Atari2:
Currently when rendering the viewport to a file using the "Display > Render To File" tool with "Zoom" != 1 (for example 2x or 3x, which I commonly need), the rectangular labels get misplaced and wrongly sized. This is fixed by applying the context.renderZoom to the relativeZoom of the label and by increasing the line width by the same factor. I don't know if this is a complete fix but it worked for me.

For example, previously, when rendering with zoom 2x, a label like the following (this is a just a screenshot, to show how it should be displayed):
screenshot
Would come out like this in the render (misplaced):
render_old
Now, after the change, it comes out like this:
render_new

Please, if anything is not correct, let me know, it's the first time for me touching this codebase so I don't know if my fix is complete. Thanks.

### DIFF
--- a/libs/qCC_db/src/cc2DViewportLabel.cpp
+++ b/libs/qCC_db/src/cc2DViewportLabel.cpp
@@ -126,13 +126,13 @@ void cc2DViewportLabel::drawMeOnly(CC_DRAW_CONTEXT& context)
 	glFunc->glPushAttrib(GL_LINE_BIT);
 
 	// focal distance change compensation
-	double relativeZoom = m_params.getFocalDistance() / params.getFocalDistance();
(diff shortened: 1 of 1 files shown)


dgirardeau:
Good catch! Thanks for the fix.

I'll wait for the build to go, and you can merge the branch after it's finished.

Atari2:
Sorry, but I don't get the 
> and you can merge the branch after it's finished.

part of your message, how am I supposed to merge it? I don't have permissions for this repo.
Is it a typo?

dgirardeau:
Ah, sorry, I assumed you would be able to do it after I allowed the checking build to start. I'll merge it.

Mind that it will take some time before I can release a new version of the 2.14.beta version with this fix.

PR #2303, opened 2026-04-07
Introduce a ccArgumentParser to shorten arg parsing

tmontaigu:
The goal is to make the parsing of arguments in
the commandline mode less verbose.

### DIFF
--- a/qCC/ccArgumentParser.cpp
+++ b/qCC/ccArgumentParser.cpp
@@ -1,0 +1,100 @@
+// ##########################################################################
+// #                                                                        #
+// #                            CLOUDCOMPARE                                #
+// #                                                                        #
+// #  This program is free software; you can redistribute it and/or modify  #
+// #  it under the terms of the GNU General Public License as published by  #
+// #  the Free Software Foundation; version 2 of the License.               #
+// #                                                                        #
+// #  This program is distributed in the hope that it will be useful,       #
+// #  but WITHOUT ANY WARRANTY; without even the implied warranty of        #
+// #  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         #
+// #  GNU General Public License for more details.                          #
+// #                                                                        #
+// #                   COPYRIGHT: CloudCompare project                      #
+// #                                                                        #
+// ##########################################################################
+#include "ccArgumentParser.h"
+
+#include <cassert>
+#include <ccLog.h>
+
+ccArgumentParser::ccArgumentParser(QStringList& arguments)
+    : m_arguments(arguments)
+{
+}
+
+const QString ccArgumentParser::peek() const
+{
+	if (m_arguments.isEmpty())
+	{
+		return QString();
+	}
+	return m_arguments.first();
+}
+
+void ccArgumentParser::skip()
+{
+	assert(!m_arguments.isEmpty());
+	if (!m_arguments.isEmpty())
+	{
+		m_arguments.removeFirst();
+	}
+}
+
+bool ccArgumentParser::isEmpty() const
+{
+	return m_arguments.isEmpty();
+}
+
+std::optional<QString> ccArgumentParser::takeNext()
+{
+	if (m_arguments.isEmpty())
+	{
+		return std::nullopt;
+	}
+	return m_arguments.takeFirst();
+}
+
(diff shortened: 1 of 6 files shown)


qCC/ccArgumentParser.h
```diff
@@ -0,0 +1,74 @@
+#pragma once
```
dgirardeau, line 17: Do you want to add a copyright notice maybe?

qCC/ccArgumentParser.cpp
```diff
@@ -0,0 +1,84 @@
+#include "ccArgumentParser.h"
```
dgirardeau, line 17: copyright notice?

qCC/ccArgumentParser.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+#include <cassert>
+#include <ccLog.h>
+
+ccArgumentParser::ccArgumentParser(QStringList& arguments)
+    : m_arguments(arguments)
+{
+}
+
+const QString* ccArgumentParser::peek() const
```
dgirardeau: maybe use a simple "QString" (not reference) using the shallow copy capability of QString? (with QString::isNull or isEmpty to check if it's valid or not)

dgirardeau: or an optional for consistency with takeNext, but that seems also a bit overkill to me.

qCC/ccArgumentParser.h
```diff
@@ -0,0 +1,74 @@
...
+	// Checks if the next argument is `-OPTION` (case insensitive)
+	// - if yes: consumes it and returns true
+	// - if not: returns false without consuming
+	bool tryConsumeOption(const QString& option);
+
+	// Returns the next argument parsed as an enum
+	// Takes a list that matches strings to enum values
+	// strings should be `UPPER_CASE`, the user value will not be case sensitive
+	template <typename T>
+	std::optional<T> takeEnum(std::initializer_list<std::pair<const char*, T>> mapping, const QString& context)
```
dgirardeau: Could 'mapping' be a reference?

qCC/ccArgumentParser.h
```diff
@@ -0,0 +1,74 @@
...
+	{
+		if (m_arguments.isEmpty())
+		{
+			ccLog::Error(QObject::tr("Missing parameter: %1").arg(context));
+			return std::nullopt;
+		}
+
+		const QString arg = m_arguments.takeFirst().toUpper();
+
+		for (const auto& [key, value] : mapping)
```
dgirardeau, line 66: Can't you use a map (or even QMap) and use the direct 'key-based' search?

qCC/ccArgumentParser.h
```diff
@@ -0,0 +1,74 @@
...
+
+		ccLog::Error(QObject::tr("Invalid %1: '%2' (expected one of: %3)")
+		                 .arg(context, arg, valid.join(", ")));
+
+		return std::nullopt;
+	}
+
+	// Parses a float from a string
+	// Logs an error on failure
+	static std::optional<float> parseFloat(const QString& arg, const QString& name);
```
dgirardeau:
```suggestion
	static std::optional<float> ParseFloat(const QString& arg, const QString& name);
```

for consistency with CC's syntax: static methods start with a capital letter)

qCC/ccArgumentParser.h
tmontaigu: Initializer_list is lighter than map or QMap (it can be stored on the stack or even in readonly memory of the program), map/QMap would do allocations.

So with a init_list we just go through the array, but they will be small so its going to be very fast due to caching and data proximity, whereas with map/QMap since their constructor is not constexpr we would first iterate to buid a map, then go through the lookup algorithm

So I would prefer to stick with initializer_list as this code is mean to handle small (<50-100) number of elements

qCC/ccArgumentParser.h
tmontaigu: Yes it can

qCC/ccArgumentParser.h
dgirardeau: Ok, makes sense

qCC/ccArgumentParser.cpp
tmontaigu: added

dgirardeau (approved):

PR #2305, opened 2026-04-10
Fix macOS bundle identifier for ccViewer

Jacky1n7:
## Summary
- add a non-empty `MACOSX_BUNDLE_GUI_IDENTIFIER` for the macOS `ccViewer` app bundle
- add a lightweight CI regression check that verifies macOS app bundles don't leave `MACOSX_BUNDLE_GUI_IDENTIFIER` unset when their plist templates reference it

## Context
`qCC` gained a macOS bundle identifier in #2209, but `ccViewer` still uses a plist template that references `${MACOSX_BUNDLE_GUI_IDENTIFIER}` without setting that property in `ccViewer/Mac/CMakeLists.txt`.

That leaves the generated `ccViewer.app/Contents/Info.plist` with an empty `CFBundleIdentifier`, which is the same failure mode behind the macOS native open/save dialog regression discussed in #2152.

## Verification
- `python3 .ci/verify_macos_bundle_identifiers.py .`

### DIFF
--- a/.ci/verify_macos_bundle_identifiers.py
+++ b/.ci/verify_macos_bundle_identifiers.py
@@ -1,0 +1,59 @@
+#!/usr/bin/env python3
+
+from __future__ import annotations
+
+import re
+import sys
+from pathlib import Path
+
+
+BUNDLES = (
+    (
+        "CloudCompare",
+        Path("qCC/Mac/CloudCompare.plist"),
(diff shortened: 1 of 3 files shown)


dgirardeau:
Thanks! I'll need the support of the macOS experts to review this one.

prascle (approved):
OK for me!

PR #2312, opened 2026-04-16
Fix build failure with GDAL 3.13.0.

sebastic:
```
error: invalid conversion from 'CSLConstList' {aka 'const char* const*'} to 'char**' [-fpermissive]
```
From [GDAL 3.13.0. NEWS](https://github.com/OSGeo/gdal/blob/v3.13.0beta1/NEWS.md):
> * GDALMajorObject: Use CSLConstList for GetMetadata, SetMetadata (API breakage)


dgirardeau (dismissed):
Thanks for the fix!

dgirardeau (approved):

PR #2328, opened 2026-05-04
Use DXF layer names for Line/Arc/Polyline primitives

rjanvier:
Points are more tricky because they could have a name **per instance** and we group them in a single point cloud on our side, so I do not handle them in the PR. I also exclude meshes (Faces) from the mechanism.

### DIFF
--- a/libs/qCC_io/src/DxfFilter.cpp
+++ b/libs/qCC_io/src/DxfFilter.cpp
@@ -656,7 +656,7 @@ class DxfImporter : public DL_CreationAdapter
 	}
 
 	//! Returns the name or the current layer or a user defined default value
-	QString getLayerNameOr(const QString& defaultName) const
+	QString getLayerNameOr(const QString& defaultName)
 	{
 		const auto layerName = QString::fromStdString(getAttributes().getLayer());
 		return layerName.isEmpty() ? defaultName : layerName;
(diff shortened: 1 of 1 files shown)


libs/qCC_io/src/DxfFilter.cpp
```diff
@@ -655,6 +655,13 @@ class DxfImporter : public DL_CreationAdapter
 		return true;
 	}
 
+	//! Returns the name or the current layer or a user defined default value
+	QString getLayerNameOr(const QString& defaultName)
```
dgirardeau:
```suggestion
	QString getLayerNameOr(const QString& defaultName) const
```

rjanvier:
Thanks for the review. @dgirardeau do you think it would be okay to do the same for point clouds? i.e. for example we could maintain a `QMap<QString, ccPointCloud*>` (here the `QString` key is the "LayerName") instead of a unique `ccPointCloud*` member in the `DxfImporter` class.  `DxfImporter::addVertex` would check in which ccPointCloud to insert the current point based on its LayerName (we could cache the latest used point cloud as points with the same Layer name should appear consecutively in the DXF file)

dgirardeau:
Looks better indeed (I have never had a particular need for DXF files myself, so I'll trust you with what can be interesting or not 😉 )

rjanvier:
Thanks, maybe we can merge this first if it suits you ;).

dgirardeau (approved):

PR #2339, opened 2026-06-30
Add 3DConnexion SpaceMouse support on macOS via hidapi

braunsi23:
## Summary

The 3D mouse support was previously Windows-only (via the proprietary 3DxWare SDK). On macOS, CloudCompare could not use a SpaceMouse at all unless the 3Dconnexion driver injected fake mouse/keyboard events.

This PR adds a platform-independent HID path that talks to 3DConnexion devices directly via [hidapi](https://github.com/libusb/hidapi), without the 3DxWare SDK.

## Changes

- **New `HIDWorker` thread** (`Mouse3DInput_mac.{h,cpp}`): a `QThread` that polls the device with `hid_read_timeout` and emits the existing `sigMove3d` / `sigReleased` / `sigOn3dmouseKey*` signals, so `cc3DMouseManager` stays unchanged.
- **Motion reports** (report ID `0x01`, 13 bytes) are parsed as 6 int16 axes; **button reports** (report ID `0x03`) are mapped to `VirtualKey` values via a static table. Battery/status reports (`0x17`) are silently ignored.
- **Axis mapping** follows the Blender NDOF default (Y/Z swapped relative to the raw HID report) with a progressive (non-linear) speed curve for fine control at small deflections and fast navigation at large deflections.
- **`GetMatrix()`** gains a platform-neutral Rodrigues rotation fallback so it no longer depends on `SPW_ArbitraryAxisToMatrix`.
- **hidapi** is pulled in as a Git submodule under `extern/hidapi` and built via `add_subdirectory`; no Homebrew/system dependency required.
- The **Windows 3DxWare path** is left untouched behind `#ifndef CC_MAC_HID`.

## Files

| File | Action |
|---|---|
| `libs/CCAppCommon/devices/3dConnexion/CMakeLists.txt` | modified – macOS build gate via hidapi submodule |
| `libs/CCAppCommon/devices/3dConnexion/Mouse3DInput.h` | modified – `HIDWorker*` member behind `#ifdef CC_MAC_HID` |
| `libs/CCAppCommon/devices/3dConnexion/Mouse3DInput.cpp` | modified – `#ifdef CC_MAC_HID` / `#else` split for `connect()`, `disconnectDriver()`, `onSiEvent()`, `GetMatrix()` |
| `libs/CCAppCommon/devices/3dConnexion/Mouse3DInput_mac.h` | new – `HIDWorker` class definition |
| `libs/CCAppCommon/devices/3dConnexion/Mouse3DInput_mac.cpp` | new – HID report parsing + worker thread |
| `extern/hidapi` | new – Git submodule (libusb/hidapi) |
| `.gitmodules` | modified – submodule entry |

`cc3DMouseManager.*` and `mainwindow.cpp` are unchanged.

## Build

```bash
cmake -S . -B build -DCMAKE_BUILD_TYPE=Release \
    -DCMAKE_PREFIX_PATH="$(brew --prefix qt@6)/lib/cmake" \
    -DOPTION_SUPPORT_3DCONNEXION_DEVICES=ON
cmake --build build -j$(sysctl -n hw.ncpu)
```

`OPTION_SUPPORT_3DCONNEXION_DEVICES` defaults to `ON` on macOS when the hidapi submodule is present.

## Test plan

- [x] Tested with a **SpaceMouse Wireless** on macOS 15 (Apple Silicon)
- [x] Cap translation: pan X/Y and zoom behave like Blender/Fusion360 (object-centric)
- [x] Cap rotation: orbit X/Y/Z
- [x] Buttons: `FIT`, `MENU`, `TOP`, `LEFT`, `RIGHT`, `FRONT`, `BOTTOM`, `BACK` mapped
- [x] Thread shuts down cleanly on exit (no crash, no `wait()` deadlock)
- [ ] Tested on older 7-byte (int8) report devices — code path included but not verified with hardware
- [ ] Tested on SpaceMouse Pro / SpaceMouse Compact — would appreciate feedback from users with those models

## Notes

- The 3Dconnexion driver (`3DconnexionUtilities.app`) may hold the HID device exclusively; users should quit it if `hid_open_path` fails.
- macOS may prompt for Input Monitoring permission on first launch.

Assisted-by: opencode:glm-5.2

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -156,6 +156,7 @@ jobs:
           libopencv-dev
           libboost-dev libboost-iostreams-dev libboost-system-dev libpcap-dev libflann-dev
           libglu1-mesa-dev
+          udev libudev-dev libhidapi-dev libusb-1.0-0-dev
           xvfb
 
       - name: Setup GCC
@@ -237,6 +238,7 @@ jobs:
           -DPLUGIN_STANDARD_QCLOUDLAYERS=ON
           -DPLUGIN_STANDARD_QVOXFALL=ON
           -DBUILD_TESTING=ON
+          -DOPTION_SUPPORT_3DCONNEXION_DEVICES=ON
 
       - name: Build
         run: cmake --build build --parallel
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -56,6 +56,7 @@ jobs:
           libopencv-dev
           libboost-dev libboost-iostreams-dev libboost-system-dev libpcap-dev libflann-dev
           libglu1-mesa-dev
+          udev libudev-dev libhidapi-dev libusb-1.0-0-dev
           xvfb
 
       - name: Configure cmake
--- a/.gitmodules
+++ b/.gitmodules
@@ -50,3 +50,6 @@
 [submodule "plugins/core/Standard/cc3DFin"]
 	path = plugins/core/Standard/cc3DFin
 	url = https://github.com/3DFin/cc3DFin.git
+[submodule "extern/hidapi"]
+	path = extern/hidapi
+	url = https://github.com/libusb/hidapi.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -50,6 +50,6 @@
 [submodule "plugins/core/Standard/cc3DFin"]
 	path = plugins/core/Standard/cc3DFin
 	url = https://github.com/3DFin/cc3DFin.git
-[submodule "extern/hidapi"]
-	path = extern/hidapi
+[submodule "libs/CCAppCommon/devices/3dConnexion/extern/hidapi"]
+	path = libs/CCAppCommon/devices/3dConnexion/extern/hidapi
 	url = https://github.com/libusb/hidapi.git
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -322,6 +322,10 @@ Improvements:
 			now apply to all selected DB tree item
 		- the 'Search by name and/or type' entry can now be applied with multiple entities selected at once
 
+	- 3D mouse support
+		- 3D mouse support on macOS, Linux and Windows (thanks to https://github.com/braunsi23 and Paul Rascle!)
+		- on Windows, option to compile with the 3DxWare SDK or the generic hidapi library
+
 	- Others:
 		- the Subsampling dialog won't allow the user to input sampling modulation parameters if all SF values are the same
 		- the shortcut to the 'Level' tool in the 'View' toolbar (left) has been removed. Contrarily to the other options in this toolbar,
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -108,7 +108,7 @@ file(REMOVE "${SOURCES_LIST_FILE}")
 
 append_sources_to_file(source_list qCC EXCLUDE_DIRS extern)
 append_sources_to_file(source_list ccViewer EXCLUDE_DIRS extern)
-append_sources_to_file(source_list libs EXCLUDE_DIRS CCAppCommon/QDarkStyleSheet qCC_db/extern qCC_io/extern)
+append_sources_to_file(source_list libs EXCLUDE_DIRS CCAppCommon/QDarkStyleSheet qCC_db/extern qCC_io/extern CCAppCommon/devices/3dConnexion/extern/hidapi)
 append_sources_to_file(source_list plugins/core/GL EXCLUDE_DIRS qSSAO/extern)
 append_sources_to_file(source_list plugins/core/IO EXCLUDE_DIRS qE57IO/extern qPhotoscanIO/extern MeshIO/extern)
 
--- a/ccViewer/ccviewer.cpp
+++ b/ccViewer/ccviewer.cpp
@@ -44,7 +44,7 @@
 #include "ccPluginManager.h"
 
 // 3D mouse handler
-#ifdef CC_3DXWARE_SUPPORT
+#ifdef CC_3DMOUSE_SUPPORT
 #include "Mouse3DInput.h"
 #endif
 
@@ -94,7 +94,7 @@ ccViewer::ccViewer(QWidget* parent, Qt::WindowFlags flags)
 	reflectPerspectiveState();
 	reflectPivotVisibilityState();
 
-#ifdef CC_3DXWARE_SUPPORT
+#ifdef CC_3DMOUSE_SUPPORT
 	enable3DMouse(true);
 #else
 	ui.actionEnable3DMouse->setEnabled(false);
@@ -1098,7 +1098,7 @@ void ccViewer::doActionAbout()
 
 void ccViewer::release3DMouse()
 {
-#ifdef CC_3DXWARE_SUPPORT
+#ifdef CC_3DMOUSE_SUPPORT
 	if (m_3dMouseInput)
 	{
 		m_3dMouseInput->disconnect(); // disconnect from the driver
@@ -1112,7 +1112,7 @@ void ccViewer::release3DMouse()
 
 void ccViewer::enable3DMouse(bool state)
 {
-#ifdef CC_3DXWARE_SUPPORT
+#ifdef CC_3DMOUSE_SUPPORT
 	if (m_3dMouseInput)
 		release3DMouse();
 
@@ -1158,7 +1158,7 @@ void ccViewer::on3DMouseKeyUp(int)
 // ANY CHANGE/BUG FIX SHOULD BE REFLECTED TO THE EQUIVALENT METHODS IN QCC "MainWindow.cpp" FILE!
 void ccViewer::on3DMouseKeyDown(int key)
 {
-#ifdef CC_3DXWARE_SUPPORT
+#ifdef CC_3DMOUSE_SUPPORT
 
 	switch (key)
 	{
@@ -1248,7 +1248,7 @@ void ccViewer::on3DMouseCMDKeyUp(int cmd)
 
 void ccViewer::on3DMouseCMDKeyDown(int cmd)
 {
-#ifdef CC_3DXWARE_SUPPORT
+#ifdef CC_3DMOUSE_SUPPORT
 	switch (cmd)
 	{
 		// ccLog::Print(QString("on3DMouseCMDKeyDown Cmd = %1").arg(cmd));
@@ -1344,7 +1344,7 @@ void ccViewer::on3DMouseCMDKeyDown(int cmd)
 
 void ccViewer::on3DMouseMove(std::vector<float>& vec)
 {
-#ifdef CC_3DXWARE_SUPPORT
+#ifdef CC_3DMOUSE_SUPPORT
 	if (m_glWindow)
 		Mouse3DInput::Apply(vec, m_glWindow);
 #endif
--- a/libs/CCAppCommon/CMakeLists.txt
+++ b/libs/CCAppCommon/CMakeLists.txt
@@ -27,8 +27,8 @@ target_sources(${PROJECT_NAME}
 		${CMAKE_CURRENT_LIST_DIR}/QDarkStyleSheet/qdarkstyle/light/lightstyle.qrc
 )
 
-if( OPTION_SUPPORT_3DCONNEXION_DEVICES )
-	target_link_3DXWARE( ${PROJECT_NAME} )
+if( OPTION_SUPPORT_3DMOUSE_WITH_3DxWARE )
+	target_link_3DxWARE( ${PROJECT_NAME} )
 endif()
 
 InstallSharedLibrary( TARGET ${PROJECT_NAME} )
--- a/libs/CCAppCommon/devices/3dConnexion/CMakeLists.txt
+++ b/libs/CCAppCommon/devices/3dConnexion/CMakeLists.txt
@@ -1,18 +1,18 @@
 if( WIN32 )
 	option( OPTION_SUPPORT_3DCONNEXION_DEVICES "Build with 3dConnexion libraries (3D mouse support)" OFF )
-	
+
 	if( OPTION_SUPPORT_3DCONNEXION_DEVICES )
 		set( 3DXWARE_INCLUDE_DIR "" CACHE PATH "3DxWare include directory" )
 		set( 3DXWARE_LIB_DIR "" CACHE PATH "3DxWare libraries directory" )
-		
+
 		if( NOT 3DXWARE_INCLUDE_DIR )
 			message( SEND_ERROR "3DXWARE include path is not specified (3DXWARE_INCLUDE_DIR)" )
 		endif()
-		
+
 		if( NOT 3DXWARE_LIB_DIR )
 			message( SEND_ERROR "3DXWARE include path is not specified (3DXWARE_LIB_DIR)" )
 		endif()
-	
+
 		target_sources( ${PROJECT_NAME}
 			PRIVATE
 				${CMAKE_CURRENT_LIST_DIR}/cc3DMouseManager.h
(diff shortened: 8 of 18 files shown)


dgirardeau:
Very interesting.

Sadly I can't test it myself. @rjanvier or @prascle are you able to test that it compiles on your side at least?

And @braunsi23, do you know if this can work on Windows as well?

braunsi23:
@dgirardeau 

The current code is gated by CC_MAC_HID, which is only defined on APPLE (CMakeLists.txt:76).
On Windows, the 3DxWare SDK path already works (CloudCompare ships with it). 

The main pain point the **hidapi** path solves is **macOS**, where no SDK exists. The benefit on Windows is marginal — unless you want to drop the proprietary SDK dependency entirely (which would simplify the build and remove the 3DXWARE_INCLUDE_DIR/3DXWARE_LIB_DIR requirement).

But honestly, i haven't a Windows machine to test it if it would work with hidapi.

dgirardeau:
Yes, some people complained that the 3DxWare SDK the Windows code is based on is not available any more and a bit deprecated. Also, using only one 3rd party library and the same code is easier to maintain.

prascle:
I have a 3DxWare mouse, so I think I'll be able to test this pull request! I was actually planning to work on 3D mouse support for macOS, so this comes at just the right time!
Paul

braunsi23:
@dgirardeau , if you want i can try to change it for Windows as well. It would be a shot in to the blue but if someone else can test it 🤷🏻‍♂️ why not. Just let me know.

dgirardeau:
I tried yesterday, bypassing the CMake options, but I got a very strange compilation error with hidapi (the VERSION file was wrongly included - while it's not a source file - leading to errors). I'll try again later.

braunsi23:
Have you init the git submodule via `git submodule update --init --recursive`?
If not the extern/hidapi directory is empty. Maybe this is the issue.

prascle:
Hello, 
I got an error during my first compilation: 
```
In file included from /Users/paul/projets/CloudCompare/CloudCompare/libs/CCAppCommon/devices/3dConnexion/Mouse3DInput.cpp:25:
/Users/paul/projets/CloudCompare/CloudCompare/libs/CCAppCommon/devices/3dConnexion/Mouse3DInput_mac.h:36:10: fatal error: 'hidapi/hidapi.h' file not found
   36 | #include <hidapi/hidapi.h>
```
With `#include <hidapi.h>`, the compilation works fine. But my 3D mouse doesn't work in CloudCompare (nothing happens, no messages). It does work with the 3dXware demos, though. I'll look into this, but any suggestions are welcome!

My 3D mouse is a SpaceMouse Compact; I have a Mac with an Apple M1 Max processor running macOS Tahoe 26.5.1.

prascle:
Well, there is one message in the CloudCompare console: 

`[21:52:21] [3D Mouse] Device: 3Dconnexion Virtual Mouse (HID, fallback)`

corresponding to `Mouse3DInput_mac.cpp` line 115.
Is my mouse too old ? :-)

prascle:
I need to disable 3DConnexionHelper.app from running in the background to see any effect. The rotations aren't working yet.

dgirardeau:
> Have you init the git submodule via `git submodule update --init --recursive`? If not the extern/hidapi directory is empty. Maybe this is the issue.

That would have been too simple ;). No, it's really an issue during the compilation, with the VERSION file being included and the compiler complaining it's not valid code (which is true). I'll investigate later. Maybe a Visual Studio specific issue.

braunsi23:
Hi @prascle 

Thanks a lot for testing this on your SpaceMouse Compact — that's exactly the kind of real-world feedback I needed, since I only have a SpaceMouse Wireless to test with. I've just pushed a new commit (4c2765dc) that should address all three of the issues you reported. Here's a breakdown:

#1 — 'hidapi/hidapi.h' file not found
You were right, and sorry about that — it was a bug on my side. My build was accidentally picking up a system-wide hidapi install from /usr/local/include/hidapi/hidapi.h (Homebrew), which masked the problem. The in-tree submodule actually exposes the header as <hidapi.h> (the file lives at extern/hidapi/hidapi/hidapi.h, with extern/hidapi/hidapi/ on the include path), so <hidapi/hidapi.h> only resolves when a system-wide hidapi is present.
Fixed in the new commit:
- Mouse3DInput_mac.h now uses #include <hidapi.h>
- CMakeLists.txt now also adds extern/hidapi/hidapi to the include path explicitly, so we don't rely solely on transitive propagation from the hidapi::hidapi target
So you should be able to revert your local #include <hidapi.h> workaround and build cleanly without any system-wide hidapi installed.

#2 — "3Dconnexion Virtual Mouse (HID, fallback)"
This is expected, not a bug. 🙂 The SpaceMouse Wireless exposes several HID interfaces, and we prefer the one tagged as "Multi-axis Controller" (usage page 0x01, usage 0x08), which is the one that carries the 6-DOF motion reports. The wired SpaceMouse Compact (PID 0xc635) apparently doesn't expose that usage descriptor — it only has a single generic HID interface — so we fall through to the fallback path that opens the first openable interface. The "Virtual Mouse" string is just the product name the device reports; it's harmless. The fallback path parses the same 13-byte motion reports as the preferred path.
So: nothing to worry about for issue #2.

#3 — Rotations not working + 3DConnexionHelper.app
This is the important one, and it has two parts:
3a. 3DConnexionHelper.app must be quit. You already discovered this — thank you for confirming. The 3Dconnexion driver daemon holds an exclusive lock on the HID device and silently consumes all reports via its own event tap, so our hid_read returns 0 forever. This is a known and unavoidable limitation of coexisting with the proprietary driver on macOS. The new commit adds two things to help future users hit this less painfully:
- At openDevice() time, we now check whether 3DConnexionHelper.app is running (pgrep -x 3DConnexionHelper) and emit a ccLog::Warning telling the user to quit it.
- If the device opens successfully but no reports arrive within ~2 seconds of the worker thread starting, we emit a second warning suggesting the driver may be intercepting reports. This catches the silent case where the lock is held without the user realising.
- 
3b. Rotations still not working even after quitting the helper. This is the one I can't fully diagnose from here, because I don't have a SpaceMouse Compact to test with. My strong suspicion is that the Compact uses a different report ID than the SpaceMouse Wireless for motion reports — the Wireless uses report ID 0x01, and my original code only accepted motion reports with that exact ID. Any other 13-byte report would fall through to the "unknown report" branch and be silently dropped.
The new commit hardens the report dispatch:
- Button reports (ID 0x03) are still matched first.
- Any other 13-byte report is now treated as a potential motion report and passed to processMotion(), which skips the first byte as a report ID regardless of its value and parses the remaining 12 bytes as 6 int16 axes. If all axes are zero, it returns early (no move event) — so misidentified non-motion reports won't produce spurious motion.
- The 7-byte fallback for older devices is unchanged.
- 
This should make the Compact work, but I'd like to verify it with your help.

Could you test the updated branch?

Regards, Steffen

prascle:
Hi @braunsi23
Thank you for this very detailed and clear analysis! 
I'm going to test your updated branch tonight (French time), and maybe—if it doesn't work properly as is—by adding a few traces, I'll be able to find the correct motion IDs !
Regards,
Paul

prascle:
Hi @braunsi23 
My SpaceMouse Compact wasn't behaving the way you expected. There are two types of reports: separate translation and rotation (codes 1 and 2), 3 axes, 16 bits per axis. 
**It works now with my commit!**

That's strange—today, I didn't get the fallback interface…
I immediately get the warning “No HID reports received from the device…,” which is confusing.

```
[00:03:59] [3D Mouse] Device: SpaceMouse Compact (HID)
[00:04:01] [3D Mouse] No HID reports received from the device. If the 3DConnexion driver (3DConnexionHelper.app) is running, quit it - it intercepts the device and prevents the HID path from seeing reports.
[00:05:27] [3D Mouse] report (7 bytes): 02 20 00 00 00 17 00
[00:05:27] [3D Mouse] report (7 bytes): 01 12 00 00 00 f8 00
[00:05:27] [3D Mouse] report (7 bytes): 02 20 00 00 00 19 00
[00:05:27] [3D Mouse] report (7 bytes): 01 12 00 00 00 12 01
[00:05:27] [3D Mouse] report (7 bytes): 02 1d 00 00 00 1b 00
[00:05:27] [3D Mouse] report (7 bytes): 01 0d 00 00 00 25 01
```

Feel free to adapt my commit; there may be other devices that behave differently...

Regards,
Paul

prascle:
Hi @braunsi23 
I ran some more in-depth tests than I did last night with my SpaceMouse Compact.

The fallback interface seems to appear only when the 3DConnexionHelper process is active.

Translations work correctly, but there’s an issue with rotations: rotation around the Y-axis is correct, but rotation around the X-axis seems reversed (rotating in the opposite direction of the movement), and the same goes for rotation around the Z-axis.

The movements are jerky, whereas movements with a standard mouse are smooth.

I assume the behavior you’re seeing with the wireless mouse is better, and I think it’ll take a little time to fully understand how to use the SpaceMouse Compact. I’ve started looking for information, and Perplexity has some interesting tests to offer.
Any ideas or references you may have are welcome...

Regards,
Paul

braunsi23:
Hi @prascle,

Thanks a lot for your commit and the detailed testing with the SpaceMouse Compact — that was extremely helpful! I've incorporated your approach, fixed the axis mapping, and addressed the jerky movement issue. A new commit (78f806f3) is now on the branch.

What changed

1. Axis mapping (Y/Z swap for translation and rotation)
You were right that the rotation felt wrong. The root cause was that the Y/Z swap was previously only applied to translation, not to rotation. The device's HID coordinate system uses Y=forward/back, Z=up/down, but CloudCompare uses Y=up/down, Z=forward/back. Now Y and Z are swapped consistently for both — matching spacenavd's DF_SWAPYZ | DF_INVYZ flags.

2. Rotation sign conventions fixed
GetMatrix() was previously negating the X and Z axes (axis(-vec[3], vec[4], -vec[5])), which caused the inverted rotations you observed. That negation has been removed — the sign conventions are now handled solely in processMotion(), so the rotation vector passes through unchanged.

3. View-relative rotation (like Blender / Fusion 360)
The rotation is now relative to the current view direction, not to the world axes. This works automatically because the Rodrigues rotation axis stays in camera space and rotateBaseViewMat() pre-multiplies (viewMat = rotMat * viewMat), which applies the rotation relative to the current view — exactly like the regular mouse drag does. No manual view matrix transformation is needed.

4. Jerky movement fixed (LOD decimation)
The jerky movement you noticed with meshes was because the 3D mouse path didn't activate LOD/mesh decimation during interaction. The regular mouse sets an internal m_mouseMoved flag which enables decimateMeshOnMove, but the 3D mouse path bypassed that entirely. I added a set3DMouseActive(bool) method to ccGLWindowInterface that sets this flag, and Apply() now calls setLODEnabled(true) + set3DMouseActive(true) during motion. When the 3D mouse is released, set3DMouseActive(false) lets the LOD refinement cycle restore full quality. You should now see the same decimated preview during 3D mouse movement as you do when dragging with a regular mouse.

5. Your SpaceMouse Compact parsing — cleaned up
Your approach for handling the separate translation/rotation reports is kept, but I cleaned it up:
- Replaced the two bool flags (isTranslationReport / isRotationReport) with an enum ReportKind { Combined, Translation, Rotation } for readability
- Added an axis buffer (m_lastAxes[6]) in HIDWorker that merges separate translation and rotation reports into a single 6-axis motion event. This means a translation report updates only tx/ty/tz and keeps the last known rx/ry/rz, and vice versa. This should also help with smoothness on the Compact, since Apply() always receives a complete 6-axis vector.
- Removed the int8 fallback path (your version already replaced it with a warning log, which I kept)

6. "No HID reports" warning softened
The warning threshold was increased from 2s to 5s, and the message reworded to say "No HID reports received yet. If you have moved the cap and nothing happens..." — this avoids the confusing false alarm you saw when simply not touching the cap yet.

7. Fallback interface comment
Updated the comment to explain that the fallback interface typically only appears when 3DConnexionHelper.app is running (it creates virtual HID interfaces) or on devices that don't expose a Multi-axis Controller usage descriptor.

Could you test again?

It would be great if you could pull the latest from the branch and test with your SpaceMouse Compact. Specifically:

1. Rotation feel — does the rotation now feel natural and view-relative (like Blender/Fusion 360)? After rotating the viewport, does "tilt cap forward" still rotate around the screen's horizontal axis?

2. Smoothness — is the movement less jerky now with the LOD decimation and axis buffering?

3. Axis directions — could you verify:
- Tilt cap forward/back → object tilts forward/back correctly?
- Tilt cap left/right → object rolls correctly?
- Twist cap → object yaws correctly?
- 
If anything still feels off, please let me know and we'll fine-tune. Thanks again for the collaboration — having both the Wireless and Compact tested is really valuable for getting this right!

Cheers, Steffen

prascle:
Hi @braunsi23 
You did an excellent job! The movement looks natural and fluid. I tested several datasets (clouds and meshes) of different sizes and experimented with the LOD decimation threshold. The rotations are view-relative; all three rotations have the correct axis and orientation. The movement remains smooth (with LOD enabled on large datasets). 
As far as I’m concerned, it’s ready for integration, at least on macOS. @braunsi23, do you want to add anything? @dgirardeau, do you want me to test it on Windows? (Note: My Windows build process is the one I use for CloudComPy, which isn’t the same as yours; it might still require a few tweaks to work for you, even if I manage to compile and test it in my CloudComPy version.)

Cheers, 
Paul

braunsi23:
Hi @prascle, thanks for the thorough testing!

I'm happy with it as it is.  I'd be really glad if it makes it into the Mac version. 

One thing I had been considering was making it user-configurable (e.g. settings for axis speed or swapping axes), but if we go that route it would need to work consistently on Windows too. So maybe we keep it simple for now and revisit later if there's demand.

Cheers, Steffen

dgirardeau:
On my side, on Windows, I can now compile without any issue.

Then it was not detecting my Space Mouse, but I realized mine has an old 'VID': 046d (matching the list here: https://3dconnexion.com/uk/support/faq/how-can-i-check-if-my-usb-3d-mouse-is-recognized-by-windows/)

So I simply added this additional vendor ID and it works like a charm!

`static constexpr unsigned short c_Old3dconnexionVID = 0x046d;`

```
hid_device_info* devs = hid_enumerate(c_3dconnexionVID, 0x0);
if (!devs)
{
	devs = hid_enumerate(c_Old3dconnexionVID, 0x0);
	if (!devs)
	{
		ccLog::Warning("[3D Mouse] No 3DConnexion HID device found");
		return false;
	}
}
```

dgirardeau:
For now can you apply this change? Then after the merge I'll probably rework things a bit to give the option to use the 3Dconnexion SDK or this universal manager at compilation time (but later).

dgirardeau:
(you'll need to run the 'Format' project also - or manually call clang-format)

prascle:
On my computer, it also works on Windows, with just a minor adjustment to the CMakeLists.txt file to account for Windows in the same way as macOS. I also need to get the hidapi.dll file to include it in my package.
I'm also trying to get it to work on Linux: it should be possible—I have some Python scripts based on hidapi that work. In any case, it's a little more complicated for the user, who has to run certain commands with “sudo” to be allowed to use the mouse. But it doesn't work in CloudCompare yet: the way device detection via hidapi works—and how devices are used—isn't exactly the same on Linux as it is on other systems.
Paul

prascle:
It now works on Linux! The same adjustments as for Windows, plus all the necessary “sudo” commands to allow the user to use the 3D mouse. It was precisely this last part that was preventing the 3D mouse from working.
Paul

braunsi23:
Hi @dgirardeau, @prascle,

Great news that it works on Windows and Linux too! Thanks for testing.

**Old vendor ID (0x046d)**

@dgirardeau — done. I added the legacy Logitech vendor ID 0x046d as a fallback in openDevice(). If hid_enumerate(0x256f, ...) finds nothing, we now try hid_enumerate(0x046d, ...) before giving up. This should auto-detect older devices (SpaceNavigator, SpaceExplorer, SpacePilot, etc.) without any user configuration. Commit 8d6d26e.

**clang-format**

Also ran clang-format (Apple clang-format 21.0.0) on the changed files per your request. The .clang-format config in the repo was respected.

**Windows / Linux**

@prascle — thanks for confirming it works on Windows and Linux with the CMake adjustments. I'll leave the cross-platform CMake changes to you and @dgirardeau since you're both testing on those platforms and I can only verify on macOS. Happy to merge whatever you come up with.

Let me know if there's anything else you'd like adjusted!

Cheers, Steffen

prascle:
Here are the changes I made to compile CloudComPy on Windows and Linux. I think this should work as-is to compile the standalone version of CloudCompare, but I'm not entirely satisfied with how I obtained the hidapi library, especially on Linux (fixed version number). There's probably a cleaner way to do this.
Paul

dgirardeau:
@prascle It seems there's a CMake issue with Ubuntu?

prascle:
Hmm, yes, to compile on Ubuntu, hidapi needs the libudev package and probably the udev package at runtime (from the Ubuntu repository). I don't use GitHub for compiling, but I can try adding udev to Ubuntu's GitHub workflow to check.

prascle:
After a few tries and errors, I finally found the necessary dependencies!

dgirardeau (dismissed):

dgirardeau:
Ok, @braunsi23 thanks for the contribution, and @prascle thanks for all the support!

I'll merge this and make some more cleanup (as it's not a 'macOS' thing only anymore).

dgirardeau:
Oops, I'll have to rebase first, so I'll probably do the cleanup right away ;)

dgirardeau:
Ok, I've made some major changes:
- I've moved the hidapi library under `libs/CCAppCommon/devices/3dConnexion/extern/hidapi`
- I've renamed the CMake options to OPTION_SUPPORT_3DMOUSE_WITH_3DxWARE and OPTION_SUPPORT_3DMOUSE_WITH_HIDAPI  (and one of them at least has to be explicitly checked to get any 3D mouse support)
- almost anywhere there was a reference to macOS, I renamed that to 'hid' or something similar
- and there was no need to copy any dll, because hidapi is compiled as a static library by default on Windows (at least on my side)

dgirardeau:
@braunsi23 and @prascle are you maybe able to test this on your side and confirm that I didn't break everything?

braunsi23:
Hi @dgirardeau,

Thanks for the rework — I just synced to your HEAD (619924b1) and did a clean build on macOS (Apple Silicon, Qt from Homebrew). 

Everything builds cleanly:

- OPTION_SUPPORT_3DMOUSE_WITH_HIDAPI=ON → builds fine
- OPTION_HID_DEBUG=ON → builds fine
- No leftover CC_MAC_HID or Mouse3DInput_mac references after the rename
- The relocated submodule (libs/CCAppCommon/devices/3dConnexion/extern/hidapi) initializes correctly

Unfortunately I can't run the functional test with my SpaceMouse Wireless right now (I'm traveling with only my MacBook and the device is at home). @prascle — would you be able to do a quick runtime test on your side to confirm the device still works after Daniel's rework?

The changes look solid to me from a code-review perspective — the rename is consistent, the CMake option split between 3DxWare and HIDAPI is clean, and the static-library approach on Windows avoids the DLL packaging issue. From my side, +1 to merge once Paul confirms runtime.

Cheers, Steffen

prascle:
Hi @dgirardeau, @braunsi23 
I built CloudCompare and successfully tested the SpaceMouse Compact on Windows, Linux, and macOS. As far as I'm concerned, everything is ready for integration!

dgirardeau:
Thanks @prascle for the confirmation!

I've updated the CHANGELOG.md file, and I'll merge this PR.

Thanks again @braunsi23 for the very nice contribution!

dgirardeau:
Hi @braunsi23, it seems on Linux either hidapi or the plugin behaves a bit strangely. There are a lot of logs, and it even react to a standards Logitech mouse.

Do you know what may be the reason?

I'll let @rjanvier describe a bit more what he sees.

braunsi23:
> Hi @braunsi23, it seems on Linux either hidapi or the plugin behaves a bit strangely. There are a lot of logs, and it even react to a standards Logitech mouse.
> 
> Do you know what may be the reason?
> 
> I'll let @rjanvier describe a bit more what he sees.

Hi @dgirardeau, thanks for the report! I'm pretty sure I know exactly what this is.

For backward compatibility with older 3DConnexion devices (SpaceNavigator, SpaceExplorer, etc. — from the era when 3DConnexion was still part of Logitech), the device enumeration falls back to the legacy Logitech vendor ID 0x046d when no device with the newer 3DConnexion ID (0x256f) is found. Since 0x046d is Logitech's actual vendor ID, on a system with no (modern) SpaceMouse attached, the fallback picks up any Logitech device — including a standard mouse. And because a mouse doesn't expose the "Multi-axis Controller" HID usage we prefer, it goes through the "first openable interface" fallback, which opens it regardless.

The "reacts to the mouse" part follows directly: mouse reports overlap in length with the SpaceMouse report formats we parse (e.g. 7-byte reports get interpreted as SpaceMouse Compact translation/rotation reports), so ordinary mouse movement leaks into the 3D view.

@rjanvier: the logs would indeed help to confirm this (especially the [3D Mouse] Device: ... (HID, fallback) line showing which device we grabbed). An lsusb listing would also be useful.

The fix will be to filter the legacy-VID enumeration to known 3DConnexion product IDs (or to require the multi-axis usage even in the fallback). I'll prepare a PR once the logs confirm the diagnosis.

Cheers, Steffen

dgirardeau:
Ok, I believe @rjanvier already has a fix.

PR #2340, opened 2026-07-05
Load E57 timestamps as scalar fields, show scalar fields in the point picker

nh2:
CloudCompare can already show timetamps from LAS/LAZ files, but not E57.

This adds that, and also the ability to read them from the point picker tool (popup and Console).

Screenshots:

image

image

---

This work was sponsored by the company I co-founded, [benaco.com](https://benaco.com).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -214,6 +214,7 @@ Improvements:
 			preserving the image sensor definition
 		- CC will now properly handle the case when a reflective transformation has been applied to a cloud (see bug fixes)
 		- Empty scans will not trigger an error anymore (just a warning message)
+		- E57 timestamps are now loaded as scalar fields
 
 	- PLY files:
 		- loading dialog: new 'Add all' button to add all the unused standard properties to be loaded as scalar fields
@@ -279,6 +280,7 @@ Improvements:
 			- A: pick 3 points and display the (a)ngles
 			- R: draw a (r)ectangle with a subtitle
 			- S: (s)ave the current label
+		- All scalar fields of the picked point are now shown in the Console.
 
 	- Cross section tool
 		- the default bounding-box is now very slightly larger than the entities bounding-box so as
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -280,7 +280,7 @@ Improvements:
 			- A: pick 3 points and display the (a)ngles
 			- R: draw a (r)ectangle with a subtitle
 			- S: (s)ave the current label
-		- All scalar fields of the picked point are now shown in the Console.
+		- All scalar fields of the picked point are now shown in the resulting Label properties dialog.
 
 	- Cross section tool
 		- the default bounding-box is now very slightly larger than the entities bounding-box so as
--- a/libs/qCC_db/src/cc2DLabel.cpp
+++ b/libs/qCC_db/src/cc2DLabel.cpp
@@ -207,7 +207,7 @@ QString cc2DLabel::getTitle(int precision) const
 
 		// If available, we display the point's currently selected SF value.
 		// A point cloud may have very many scalar fields (e.g. multi-spectral data),
-		// so we show ALL of them in the Console only, not in this popup.
+		// so we show ALL of them in the Label properties dialog only, not in this popup.
 		LabelInfo1 info;
 		getLabelInfo1(info);
 		if (info.hasSF)
--- a/libs/qCC_db/src/cc2DLabel.cpp
+++ b/libs/qCC_db/src/cc2DLabel.cpp
@@ -205,7 +205,9 @@ QString cc2DLabel::getTitle(int precision) const
 		title = m_name;
 		title.replace(POINT_INDEX_0, m_pickedPoints[0].itemTitle());
 
-		// if available, we display the point SF value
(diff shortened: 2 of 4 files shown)


libs/qCC_db/src/cc2DLabel.cpp
```diff
@@ -883,8 +927,16 @@ QStringList cc2DLabel::getLabelContent(int precision) const
 			QString colorStr = QString("Color: (%1;%2;%3;%4)").arg(info.color.r).arg(info.color.g).arg(info.color.b).arg(info.color.a);
 			body << colorStr;
 		}
-		// scalar field
-		if (info.hasSF)
+		// scalar fields
+		if (!info.sfValues.empty())
+		{
+			for (const SFValue& sfVal : info.sfValues)
```
dgirardeau, line 935: This might become a problem for clouds with a lot of scalar fields (e.g. multi-spectral data, or simply LAS files)

libs/qCC_db/src/cc2DLabel.cpp
```diff
@@ -1466,6 +1543,19 @@ void cc2DLabel::drawMeOnly2D(CC_DRAW_CONTEXT& context)
 							tab.colContent[c] << "B";
 							tab.colContent[c + 1] << QString::number(info.color.b);
 						}
+
+						// next block: all scalar fields (one row per SF, in a single plain-background block
```
dgirardeau: same comment as above

libs/qCC_db/src/cc2DLabel.cpp
nh2: I don't have data with many scalar fields, so I don't know what's common here.

What's your recommendation here?

For me it only matters that there's _some_ way to read and copy the data out; the Console is also good enough for me for that.

Should I just remove it from that popup and keep it only in the console?

---

Side remark: One can still read the value in the popup if one first switches the `Properties > Scalar Fields > Active` dropdown to the desired scalar field, screenshot:

image

nh2:
I force-pushed some `-Wclang-format-violations` fixes.

libs/qCC_db/src/cc2DLabel.cpp
dgirardeau: So a LAS file can easily have 8 or more scalar fields. And some applications will generate 50 or 100...

So I would not display all of them in a 2D label.

nh2:
> I force-pushed some `-Wclang-format-violations` fixes.

@dgirardeau FYI it would be a good idea to add to `CONTRIBUTING.md` that one should run `cmake --target check-format`, otherwise one notices that the CI is not happy only after a very long delay.

libs/qCC_db/src/cc2DLabel.cpp
nh2: > Should I just remove it from that popup and keep it only in the console?

I force-pushed an implementation of that now.

dgirardeau:
Looks good to me!

Can you update the CHANGELOG.md file?

nh2:
> Can you update the CHANGELOG.md file?

Done

CHANGELOG.md
```diff
@@ -279,16 +280,12 @@ Improvements:
 			- A: pick 3 points and display the (a)ngles
 			- R: draw a (r)ectangle with a subtitle
 			- S: (s)ave the current label
+		- All scalar fields of the picked point are now shown in the Console.
```
dgirardeau: Isn't it more interesting to read this info in the Label properties dialog instead of the Console? (as it's dynamically updated)

dgirardeau:
```suggestion
		- All scalar fields of the picked point are now shown in the resulting Label properties dialog.
```

libs/qCC_db/src/cc2DLabel.cpp
```diff
@@ -205,7 +205,9 @@ QString cc2DLabel::getTitle(int precision) const
 		title = m_name;
 		title.replace(POINT_INDEX_0, m_pickedPoints[0].itemTitle());
 
-		// if available, we display the point SF value
+		// If available, we display the point's currently selected SF value.
+		// A point cloud may have very many scalar fields (e.g. multi-spectral data),
+		// so we show ALL of them in the Console only, not in this popup.
```
dgirardeau:
```suggestion
		// so we show ALL of them in the Label properties dialog only, not in this popup.
```

CHANGELOG.md
dgirardeau: I've updated the code to merge it.

libs/qCC_db/src/cc2DLabel.cpp
nh2: @dgirardeau What UI element is the "Label properties dialog"? Because the code certainly also prints all scalar values in the Console when you pick them.

dgirardeau: Ah yes, the description is sent to the console at the time of the label is created, but in fact it's also displayed in the 'properties' view at the bottom left part of the screen, when you highlight the label. And this is where you'll see all the values up-to-date if any SF is modified. The code would be much simpler if it was meant only to show the values in the Console.

dgirardeau (approved):

PR #2341, opened 2026-07-05
Fix failure to load symlink if it points to file without extension

nh2:
Until now, CloudCompare tried to open a symlink such as

    myfile.e57 -> someFileNotEndingInDotE57

it would fail with error

    [Load] Can't guess file format: no file extension

Such cases are common in content-addressed storage systems (where the underlying files have hashes as their file names).

This commit fixes it by taking the first file extension along the symlink chain.

---

This work was sponsored by the company I co-founded, [benaco.com](https://benaco.com/).

### DIFF
--- a/libs/qCC_io/src/FileIOFilter.cpp
+++ b/libs/qCC_io/src/FileIOFilter.cpp
@@ -62,6 +62,10 @@ QString FileIOFilter::GetRealFilename(QString filename)
 	{
 		// Resolve the whole symbolic link chain (not just a single level).
 		// canonicalFilePath() returns an empty string on broken links.
+		// It also resolves through Windows shortcut and macOS alias files;
+		// that is undocumented in Qt and will likely change in the future,
+		// given that `isSymlink()` is already documented to likely change, see:
+		//     https://github.com/CloudCompare/CloudCompare/pull/2341#issuecomment-4947138037
 		QString target = fi.canonicalFilePath();
 		if (!target.isEmpty())
 		{
@@ -435,8 +439,8 @@ ccHObject* FileIOFilter::LoadFromFile(const QString&  inputFilename,
 {
 	Shared filter;
 
-	// special case for symbolic link, shortcut or alias files
-	QString filename = GetRealFilename(inputFilename);
+	// special case for symbolic link, Windows shortcut or macOS alias files
+	QString resolvedFilename = GetRealFilename(inputFilename);
 
 	// if the right filter is specified by the caller
 	if (!fileFilter.isEmpty())
@@ -471,7 +475,7 @@ ccHObject* FileIOFilter::LoadFromFile(const QString&  inputFilename,
 		}
 	}
 
-	return LoadFromFile(filename, loadParameters, filter, result);
+	return LoadFromFile(resolvedFilename, loadParameters, filter, result);
 }
 
 CC_FILE_ERROR FileIOFilter::SaveToFile(ccHObject*            entities,
@@ -485,11 +489,11 @@ CC_FILE_ERROR FileIOFilter::SaveToFile(ccHObject*            entities,
 	}
 
 	// special case for symbolic link, shortcut or alias files
-	QString filename = GetRealFilename(inputFilename);
+	QString resolvedFilename = GetRealFilename(inputFilename);
 
 	// if the file name has no extension, we had a default one!
-	QString completeFileName(filename);
-	if (QFileInfo(filename).suffix().isEmpty())
+	QString completeFileName(resolvedFilename);
+	if (QFileInfo(resolvedFilename).suffix().isEmpty())
 	{
 		completeFileName += QString(".%1").arg(filter->getDefaultExtension());
 	}
@@ -501,17 +505,17 @@ CC_FILE_ERROR FileIOFilter::SaveToFile(ccHObject*            entities,
 	}
 	catch (...)
 	{
-		ccLog::Warning(QString("[I/O] CC has caught an unhandled exception while saving file '%1'").arg(filename));
+		ccLog::Warning(QString("[I/O] CC has caught an unhandled exception while saving file '%1'").arg(resolvedFilename));
 		result = CC_FERR_CONSOLE_ERROR;
 	}
 
 	if (result == CC_FERR_NO_ERROR)
 	{
-		ccLog::Print(QString("[I/O] File '%1' saved successfully").arg(filename));
+		ccLog::Print(QString("[I/O] File '%1' saved successfully").arg(resolvedFilename));
 	}
 	else
 	{
-		DisplayErrorMessage(result, "saving", filename);
+		DisplayErrorMessage(result, "saving", resolvedFilename);
 	}
 
 	return result;
--- a/libs/qCC_io/src/FileIOFilter.cpp
+++ b/libs/qCC_io/src/FileIOFilter.cpp
@@ -60,10 +60,49 @@ QString FileIOFilter::GetRealFilename(QString filename)
 	QFileInfo fi(filename);
 	if (fi.isSymLink())
 	{
-		return fi.symLinkTarget();
+		// Resolve the whole symbolic link chain (not just a single level).
+		// canonicalFilePath() returns an empty string on broken links.
+		QString target = fi.canonicalFilePath();
+		if (!target.isEmpty())
+		{
+			return target;
+		}
 	}
 
 	return filename;
+}
+
+QString FileIOFilter::GetExtensionForFormatGuessing(const QString& originalFilename)
+{
+	// Walk the symlink link chain, return the first extension found.
+	QString currentPath = originalFilename;
+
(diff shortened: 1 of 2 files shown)


nh2:
This one bugged me for years, finally got time to fix it.

nh2:
In the existing code, I don't really understand why in the `if (!fileFilter.isEmpty())` case it uses `filename = GetRealFilename(inputFilename)` instead of just `inputFilename`.

Is this necessary on Windows, e.g. can you not just open a Link/Shortcut there?
On Linux it is usually cleaner to just open the symlink and let the OS handle where it points to.

libs/qCC_io/src/FileIOFilter.cpp
```diff
@@ -60,12 +60,51 @@ QString FileIOFilter::GetRealFilename(QString filename)
...
 	}
 
 	return filename;
 }
 
+QString FileIOFilter::GetExtensionForFormatGuessing(const QString& originalFilename)
+{
+	// Walk the symlink link chain, return the first extension found.
+	QString   currentPath = originalFilename;
+	QFileInfo fi(currentPath);
```
dgirardeau: why not keeping that within the for loop scope below? (i.e. just before line 84 + you can remove line 102)

dgirardeau:
> In the existing code, I don't really understand why in the `if (!fileFilter.isEmpty())` case it uses `filename = GetRealFilename(inputFilename)` instead of just `inputFilename`.
> 
> Is this necessary on Windows, e.g. can you not just open a Link/Shortcut there? On Linux it is usually cleaner to just open the symlink and let the OS handle where it points to.

Isn't it to manage the file extensions?

libs/qCC_io/src/FileIOFilter.cpp
nh2: Ah yeah fair point, will fix.

nh2:
I force-pushed some `-Wclang-format-violations` fixes.

nh2:
> Isn't it to manage the file extensions?

@dgirardeau I mean why does CloudCompare care what the symlink target is at all if the user has picked a file filter in the Open dialogue?

I get that when it doesn't know the file extension, CloudCompare may want to use as much information possible to figure it out (e.g. walk symlinks), but if the user already chose the file type, why resolve the file at all instead of opening the symlink directly as the user-chosen file type?

dgirardeau:
I don't remember what the issue was, but it was answering a real issue 😅

libs/qCC_io/src/FileIOFilter.cpp
nh2: Should be done now.

nh2:
In any case you might want to test this on Windows because I build only on Linux.

libs/qCC_io/src/FileIOFilter.cpp
```diff
@@ -412,8 +451,7 @@ ccHObject* FileIOFilter::LoadFromFile(const QString&  inputFilename,
 	}
 	else // we need to guess the I/O filter based on the file format
 	{
-		// look for file extension (we trust Qt on this task)
-		QString extension = QFileInfo(filename).suffix();
+		QString extension = GetExtensionForFormatGuessing(inputFilename);
```
dgirardeau, line 458: So I had the time to test this on my side, and I believe we should read the extension from `filename` instead of `inputFilename`.

We should maybe rename `filename` to something like `realFilename` by the way.

dgirardeau, line 458:
```suggestion
		QString extension = GetExtensionForFormatGuessing(filename);
```

dgirardeau (changes requested):

nh2:
> > @dgirardeau I mean why does CloudCompare care what the symlink target is at all if the user has picked a file filter in the Open dialogue?
>
> I don't remember what the issue was, but it was answering a real issue 😅

Writing up here some more I found on the topic (I found this when thinking more about what correct logic should do):

I am looking now at https://github.com/CloudCompare/CloudCompare/commit/28070745aec73c82b06493cb77630ad04aebe411

```c++
	if (fi.isSymLink())
	{
		return fi.symLinkTarget();
```

I am now getting the opinion that this resolution of Windows shortcuts (`.lnk` files, with our without that file suffix), and macOS aliases -- both of which are separate from POSIX symlinks that both these OSs laso support -- is a bad idea as implemented via Qt, and at least controversial in general, for a couple reasons:

* Other Windows sofware does not necessarily that:
  * Windows's own Notepad, when asked to save a file to a `.lnk`, will override the actual `.lnk`, not save through it, thus corrupting it and turning it into a plain text file. It does, however, support loading through `.lnk` files. Tested on Windows Server 2025.
  * However, Firefox 152 supports loading and saving files through `.lnk` files (see also [FF #271732](https://bugzilla.mozilla.org/show_bug.cgi?id=271732) and [FF #283730](https://bugzilla.mozilla.org/show_bug.cgi?id=283730)).
* Qt docs on [`isSymLink`](https://doc.qt.io/qt-6/qfileinfo.html#isSymLink) (which is introduced by the above commit) say:

  > In addition, true will be returned for shortcuts (`*.lnk` files) on Windows, and aliases on macOS. This behavior is **deprecated and will likely change in a future version** of Qt. Opening a shortcut or alias will open the .lnk or alias file itself.

  So the logic used today will likely break in the future, and would have to be implemented manually with the new [`isAlias()`](https://doc.qt.io/qt-6/qfileinfo.html#isAlias) etc.
* It is completely undocumented in Qt whether the current CloudCompare code actually works.
  Only `isSymLink()` documents that it _currently_ follows shortcuts/aliases. But the [`symLinkTarget()`](https://doc.qt.io/qt-6/qfileinfo.html#symLinkTarget) docs don't mention this at all!
  It does currently work though (I read the Qt code and also tried it in CloudCompare).
  The [`canonicalFilePath()`](https://doc.qt.io/qt-6/qfileinfo.html#canonicalFilePath) I switched to in this PR so far also doesn't document it, but apparently it still works there as well.
  In both cases, CloudCompare is relying on undocumented Qt implementation details that might change any time.
  I would expect that as soon as Qt changes `isSymLink()`, the other ones will also switch to the saner behaviour, thus silently breaking CloudCompare's current behaviour, and requiring that all of that logic be re-implemented in CloudCompare.

libs/qCC_io/src/FileIOFilter.cpp
nh2: @dgirardeau How ould that work? Then we'd be starting the symlink chain walk from the end of the chain -- so there'd be nothing to walk.

nh2: Also, why did you think we should do that, did something not work?

dgirardeau:
I guess we will just have to change this when it breaks? So far it seems to work 😇.

Also, I don't use .lnk files on my side to test this feature, but aliases created with mklink. Are they different?

libs/qCC_io/src/FileIOFilter.cpp
dgirardeau: Yes, on Windows using the input filename (which is the symbolic link, the first stage) obviously doesn't give the right extension, thus leading to CC failing detecting the right extension.

You have to provide the real file with the real extension (`filename`) to the 'GetExtensionForFormatGuessing' method.

nh2:
> Also, I don't use .lnk files on my side to test this feature, but aliases created with mklink. Are they different?

@dgirardeau Yes, Windows (on NTFS) also supports real symlinks, that's what `mklink` does. Symlinks are resolved by the OS (e.g. when you do an `open` syscall, the OS follows the symlink), while Windows shortcuts are just small binary files that contain the file they point to, and not resolved by the OS (as far as I can tell).

libs/qCC_io/src/FileIOFilter.cpp
nh2: @dgirardeau I've pushed another commit that cleans up the variable naming a bit, thus making it more obvious why we should use `inputFilename`.

dgirardeau: Well, it doesn't work on my side 😅

dgirardeau: And maybe for the context: I drag and drop the symbolic link on the 3D view. So CC calls 'LoadFromFile' with `inputFilename` being the first stage. Then we reslove the real name (`resolvedFilename`) but we try to detect the extension with `inputFilename`:

image

dgirardeau: And apparently your magic doesn't work (also I'm not sure why we need it if we get `resolvedFilename` already)

dgirardeau: And last, this is a single level symlink (created with mklink).

dgirardeau: Yes, I think I'm now even more confused about what you are trying to do in `GetExtensionForFormatGuessing`. If the symbolic link happens to have an extension (.lnk?) then it returns `.lnk` isn't it?

Issue #1034, opened 2020-02-10
Match scales (PCA) via Command Line Interface


raphael2692:
Hi all, 

I am trying to use Cloud Compare via Command Line to automate the alignement of 2 clouds (obtained from sampling 2 meshes). 

With the GUI I am able to achieve this with these operations:

1. Tools -> Registration -> Match bounding box centers
2. Tools -> Registration  -> Match scales (PCA)
3. Tools -> Registration -> ICP Fine registration

I am working on replicate these operations via CLI but it seems the _Match scales_ filter is missing. Is it really so? Can you suggest me a work around? Thanks!

dgirardeau:
Indeed, the 'match scales' feature has probably not been ported to the command line yet ;)

Have you tried ICP with variable scale instead? (-ADJUST_SCALE)

raphael2692:
Hi @dgirardeau , thanks for your response. 

I tried ICP with that parameter only but the result is either not homomorphic (aka the the aligning cloud is deformed) or completely wrong (the cloud is caled down by a huge factor and fitted a small patch of the template). 

My ultimate goal is to use a template cloud to automatically scale different clouds of the same scene created via photogrammetry pipelines (structure-from-motion + multiview stereo, like openMVG). 

Do you have any other suggestion? 😊

dgirardeau:
Apart from instrumenting the command line mode to make this method available (you or someone else), no ;)

raphael2692:
Hi @dgirardeau , can you help me understanding where this function is located in the source code?

Also, it would be interesting to implement a scale range costraint, as it would dramatically reduce local minima pitfalls (resulting in small patch fitting) as described in [this paper](https://www.researchgate.net/publication/242506513_Scaling_iterative_closest_point_algorithm_for_registration_of_mD_point_sets).  Thanks!

dgirardeau:
The method itself is in ccLibAlgorithms::ApplyScaleMatchingAlgorithm (https://github.com/CloudCompare/CloudCompare/blob/8f4cc8abfb697a820f32feef02a41ea2c207623c/qCC/ccLibAlgorithms.cpp#L575). You can see how it is called in the mainwindow.cpp file: https://github.com/CloudCompare/CloudCompare/blob/42203ddcda49a169b10cfe8b6d340548f53ff582/qCC/mainwindow.cpp#L5233

[...]

PR #2347, opened 2026-07-22
Add -MATCH_SCALES command line option (#1034)

somtri:
## Summary

This ports the "Tools > Registration > Match scales" tool to the command line, requested in #1034. The tool exists in the GUI but was never exposed to the CLI.

New command:

```
-MATCH_SCALES {BB_MAX_DIM|BB_VOLUME|PCA_MAX_DIM|ICP} [-REFERENCE {index}] [-RMS_DIFF {value}] [-OVERLAP {percent}]
```

It rescales every loaded cloud/mesh to match the scale of a reference entity, wrapping the existing `ccLibAlgorithms::ApplyScaleMatchingAlgorithm` (the same function `MainWindow::doActionMatchScales` calls). `-REFERENCE` picks the reference entity by 0-based index into the loaded clouds-then-meshes list (default 0). `-RMS_DIFF` and `-OVERLAP` only affect the ICP algorithm and default to the GUI dialog's values (1e-5 and 100).

## Progress dialog change in ApplyScaleMatchingAlgorithm

`ApplyScaleMatchingAlgorithm` created a `ccProgressDialog` and called `start()` unconditionally, and `start()` always calls `show()`. Called from the CLI, that pops a dialog even under `-SILENT`. The other CLI-reachable algorithms in the same file create the dialog only when `parent` is non-null; `ComputeGeomCharacteristic` is one example. I made `ApplyScaleMatchingAlgorithm` follow the same pattern: the dialog is created only when `parent` is non-null, and `NormalizedProgress` already accepts a null callback. The GUI path always passes a non-null parent, so its behavior does not change.

## Scope

This covers the "match scales" CLI command only. The second request in #1034, specifying ICP axis rotation and translation on the command line, is a separate and larger change that I left out.

## Testing

I built this locally on Linux (Ubuntu 24.04, GCC 13, Qt 6.4) in a core configuration with plugins and ccViewer turned off. The full CloudCompare target compiles and links cleanly, including the three changed files. I have not exercised the command against sample data yet, so its runtime behavior rests on code review and on how closely it follows the existing commands (it mirrors `CommandMatchBBCenters` for gathering entities and `CommandICP` for parsing options). CI will build on Windows and macOS.

Addresses #1034.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,10 @@ New features:
 		- distances between a point cloud and a disc can be computed with 'Tools > Distances > Cloud/primitive dist'
 
 	- New Command line options
+		- New command -MATCH_SCALES {BB_MAX_DIM|BB_VOLUME|PCA_MAX_DIM|ICP} [-REFERENCE {index}] [-RMS_DIFF {value}] [-OVERLAP {percent}]
+			- ports the 'Tools > Registration > Match scales' tool to the command line
+			- rescales all loaded clouds/meshes to match the scale of the reference entity (0-based index, 0 by default)
+			- -RMS_DIFF and -OVERLAP only apply to the ICP algorithm (defaults: 1e-5 and 100 respectively)
 		- New command -DISTANCES_FROM_SENSOR [-SQUARED]
 			- to compute the distances from every point of the cloud to the associated sensor origin (if any)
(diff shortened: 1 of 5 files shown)


dgirardeau (approved):
This looks very clean, thanks!

dgirardeau:
And thanks for the contribution. I'll merge this PR.

PR #2354, opened 2026-07-29
scalar fields manager

farmakis:
This a table-like dialog for iterating through and managing the scalar fields of selected entities (only point clouds atm)


dgirardeau:
Thanks for the contribution. I'll need some time to test it on my side and do a thorough review.

dgirardeau:
I've rebased and tested this PR.

In dark mode, the background for the 'read only' cells prevent from reading the content, is it intentional?
image

farmakis:
No, it’s definitely not intentional. It’s because I didn’t check on dark mode.

dgirardeau:
If I set the foreground color to black, it's a little bit better:
image

Even though the highlighted cells are not super readable either, but that's ok.

farmakis:
Only the SF name column should be editable. Any SF statistic value shouldn’t be.
What is the color differentiation between rows?
Is the Scan Direction an editable row then? Or just selected?

dgirardeau:
I simply highlighted some rows here (number 4 and 5).

farmakis:
Looks ok to me!

dgirardeau:
I'll push my few fixes (mostly the comments were starting with //! instead of //, the .ui file was modified when I opened it with Qt Designer 6.8.3, and this foreground color for the text). And then I'll let you maybe test on your side with (Windows) dark mode enabled?

farmakis:
Cool. However I will not be able to check on windows for a week. Im away with only a macbook available.

dgirardeau:
Ok, then I guess we'll merge it and you'll test it later just in case.

farmakis:
Sure. Thanks.

dgirardeau:
I spent a little more time and I found a few glitches and some unused code.  I'll let your review the 2 last commits just in case.

dgirardeau (approved):

Issue #2043, opened 2024-07-27
PLY exporter renames fields


PhaseSpaceContinuum:
**Describe the bug**
Importing a PLY file with fields X, Y, Z, intensity and exporting it again, leads to the unexpected field names:
X, Y, Z, scalar_intensity.

**How to reproduce**

**Expected behaviour**
Import and export should not prepend "scalar_" to scalar field names and/or provide an option to skip this prepending.

**Your environment**
 - CC Version (or git hash/tag): f0df1c2691ab77e1400848077a092aca308de3f2
 - OS & Version: Ubuntu 22.04 LTS
 - Graphics card: -
 - Qt Version (if compiling): 5.15.13

dgirardeau:
Since CC can load many different formats or you can even create clouds from scratch, at the time CC saves a cloud to PLY, it applies its own format and can't 'remember' where the cloud comes from and what the original PLY fields name was...

So I guess the only solution would be to give the option to not prepend the field by 'scalar_' indeed. If anyone wants to do this change, he/she is welcome to do so 👼

tmontaigu:
Maybe we can do the same as we do for last extra bytes: when loading, attach to the cloud a mapping from scalar fiel name to how it was (type wise) in the original file

And try to use that when saving

PR #2356, opened 2026-08-02
Add -PLY_NO_SF_PREFIX command line option (#2043)

somtri:
Addresses #2043.

Master already keeps the original names of scalar fields that came from the input PLY file (06837ba8), so the round-trip case in the original report works today. What is left is the case you said you would welcome a patch for: a cloud that did not come from a PLY file, for example one loaded from LAS or E57, or one whose scalar fields were computed inside CC. Those still get `scalar_` prepended at save time, with no way to turn it off.

`-PLY_NO_SF_PREFIX` makes the PLY filter write the scalar field names without that prefix.

It follows the existing `PlyFilter::SetDefaultOutputFormat` accessor and the `-PLY_EXPORT_FMT` command that sets it.

Notes:
- Default behaviour is unchanged. The prefix is still added unless the option is passed.
- Fields already present in the source PLY keep their original names either way. That path is untouched.
- Spaces are still replaced with underscores when the prefix is skipped, since PLY property names cannot contain whitespace.
- The setter would also make a GUI checkbox easy to add. I left that out because I have no way to test the dialog.

Verified on Ubuntu 24.04 with GCC and Qt6 (core build, plugins off): no new warnings, and the existing ctest suite still passes. For the behaviour itself I used a 4 column ASCII cloud, so nothing that came from a PLY file. The default run writes `property float scalar_Scalar_field`. With `-PLY_NO_SF_PREFIX` the same cloud writes `property float Scalar_field`. That field is named "Scalar field" internally, so the second case also shows the space being replaced. I have not built this on Windows or macOS.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,9 @@ New features:
 			- ports the 'Tools > Registration > Match scales' tool to the command line
 			- rescales all loaded clouds/meshes to match the scale of the reference entity (0-based index, 0 by default)
 			- -RMS_DIFF and -OVERLAP only apply to the ICP algorithm (defaults: 1e-5 and 100 respectively)
+		- New command -PLY_NO_SF_PREFIX
+			- tells the PLY filter not to add the 'scalar_' prefix to the scalar field names when saving
+			- scalar fields coming from an input PLY file already keep their original name
 		- New command -DISTANCES_FROM_SENSOR [-SQUARED]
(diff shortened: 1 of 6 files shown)


dgirardeau (approved):

dgirardeau:
Sorry for the delay. I haven't been able to test it on my side, but this looks pretty straightforward. We can probably merge it.

Issue #2020, opened 2024-05-28
CLI-ICP: get the possibility get path name for transformation matrix


SchmidL:
**Describe the feature you would like**
When the ICP is calculated the transformation matrix is saved directly next to the initial point clouds. But i would like to save next to my output when I export the transformed point cloud. So instead of autosave the point cloud, it would be amazing to pass a file path for the transformation matrix along with the command so that the identification is also clearer

**Is your feature request related to a problem? Please describe.**
not a problem, and its hard to be 100% certainty to copy the right autogenerated transformation matrix file.

dgirardeau:
Ah yes, there was a similar request a few days ago regarding the fit plane description file... I'll add this to the TODO list (if not already there 😅)

SchmidL:
Thank you very much!

PR #2360, opened 2026-08-15
Add -OUTPUT_MATRIX_FILE and -OUTPUT_INFO_FILE command line options (#2020)

somtri:
Addresses #2020.

The `-ICP` command always writes its registration matrix next to the input cloud, as `{path}/{basename}_REGISTRATION_MATRIX[_timestamp].txt`. If the registered cloud is exported somewhere else, the matrix stays behind, and it gets hard to tell which autogenerated file came from which run.

In the issue thread you grouped this with a second file:

> Ah yes, there was a similar request a few days ago regarding the fit plane description file... I'll add this to the TODO list (if not already there)

So this PR covers both sites.

### `-ICP ... -OUTPUT_MATRIX_FILE {filename}`

Writes the registration matrix to `{filename}` instead of the autogenerated path.

### `-BEST_FIT_PLANE ... -OUTPUT_INFO_FILE {filename}`

Writes the plane information file to `{filename}` instead of the autogenerated path.

In both cases the filename is used exactly as given: no timestamp and no `.txt` extension is appended. When the option is absent, the existing naming is untouched.

The two commands are not symmetric, which shapes the second option. `-ICP` writes exactly one matrix file per run, but `-BEST_FIT_PLANE` loops over every loaded cloud and writes one info file per cloud. One explicit path cannot describe N files, so `-OUTPUT_INFO_FILE` refuses to run when more than one cloud is loaded:

```
Option 'OUTPUT_INFO_FILE' requires a single loaded cloud (2 are loaded)
```

I made that an error rather than a warning, since the alternative is each cloud silently overwriting the previous one. If you would rather it accept several clouds, a directory argument or a per-cloud naming scheme would both work and I am happy to switch.

Each command keeps its own existing option-parsing idiom, so neither parser loop is restructured: `-ICP` uses `ccArgumentParser` (same shape as `-ROT`), and `-BEST_FIT_PLANE` uses the raw argument deque (same shape as `-VOLUME -TO_FILE`). At both write sites the original default-path code is unchanged, just moved inside an `isEmpty()` branch.

### Verified

Ubuntu 24.04, Qt 6.4, Release: 398/398 targets built, no new compiler warnings (the 4 remaining are pre-existing, in `shapelib` and `icons.qrc`). `ctest` 2/2 passed. `clang-format 18 --dry-run --Werror` and `git diff --check` both exit 0.

Run against the built binary:

| Case | Result |
|---|---|
| `-ICP`, no option | still writes `data_REGISTRATION_MATRIX.txt` next to the input cloud |
| `-ICP -OUTPUT_MATRIX_FILE out/my_matrix.txt` | writes exactly that file, no autogenerated file, no timestamp or `.txt` appended |
| `-BEST_FIT_PLANE` 1 cloud, no option | still writes `plane_BEST_FIT_PLANE_INFO.txt` |
| `-BEST_FIT_PLANE -OUTPUT_INFO_FILE out/my_plane.txt` | writes exactly that file |
| same, with 2 clouds loaded | exits non-zero, and a pre-existing file at that path is left untouched |
| `-BEST_FIT_PLANE` 2 clouds, no option | still writes 2 info files, one per cloud |
| either option with no value after it | errors out, no crash |

The matrix produced through `-OUTPUT_MATRIX_FILE` is identical to the one produced at the default path, and on a synthetic pair offset by a known rigid transform (2 degrees about Z, translation 0.08/0.05/0.03) it recovers the inverse transform to six decimal places.

I have not built this on Windows or macOS locally, so those platforms rest on CI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,13 @@ New features:
 		- New command -PLY_NO_SF_PREFIX
 			- tells the PLY filter not to add the 'scalar_' prefix to the scalar field names when saving
 			- scalar fields coming from an input PLY file already keep their original name
+		- New option -OUTPUT_MATRIX_FILE {filename} for the -ICP command
+			- saves the registration matrix to this file instead of the automatically generated '{cloud path}/{cloud name}_REGISTRATION_MATRIX.txt'
+			- the filename is used as is: no timestamp and no '.txt' extension are appended
+		- New option -OUTPUT_INFO_FILE {filename} for the -BEST_FIT_PLANE command
+			- saves the plane information file to this file instead of the automatically generated '{cloud path}/{cloud name}_BEST_FIT_PLANE_INFO.txt'
+			- the filename is used as is: no timestamp and no '.txt' extension are appended
+			- as this command writes one information file per loaded cloud, this option requires that a single cloud is loaded
 		- New command -DISTANCES_FROM_SENSOR [-SQUARED]
 			- to compute the distances from every point of the cloud to the associated sensor origin (if any)
 		- New command -SCATTERING_ANGLES [-DEGREES]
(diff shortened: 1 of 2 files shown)


dgirardeau (approved):

dgirardeau:
Thanks for the contribution!

Issue #1885, opened 2023-11-21
Fit statistical model on an active scalar field exposed from command line mode


XRarach:
**Describe the feature you would like**
For me, it would be very useful if it would be possible to call fit statistical model on active/only one scalar field from the command line.

**Describe alternatives you've considered**
Local Statistical Test - but this does not seem to provide all in one evaluation of the previous comparison. 
My use case is that I do alignment and comparison of 2 meshes, filtering to remove outliers, but after this I would like to evaluate quality of signal to noise ratio - ideally by just one value, and this is exactly what this functionality provides. 
I also considered using just the output RMS value from the alignment, but this does not seem to be coherent with the RMS I get from Fit statistical model function in the GUI. I am getting 2 orders of magnitude different RMS from this function in comparison to Gaus Distribution fitting from the GUI.

**Is your feature request related to a problem? Please describe.**
The issue I am trying to solve is described above. 

If such a function is already exposed in some way, please let me know.
Thank you in advance for your time and for considering exposing this functionality.

dgirardeau:
Hum, maybe it's already accessible via CloudCompy? (see [https://github.com/CloudCompare/CloudCompy](url)).

Otherwise I can add that to the TODO list, however it's not clear when someone will have the motivation and time to do it, and it's also not clear what the output should be (a text file?).

XRarach:
Hi, a text file would be useful yes, for me the RMSE as a single measure of the precision of the measurement is quite useful. 
I would like to use this in automated measurement of how well a scan of an object went with one device in comparison to another one, which would serve as a ground truth. After removing outliers, within some precision range - how much of an error is still there ... 
So for example a whole data log into a file + RMSE as an output. 

Of course, I understand that you are most probably very busy.
Thank you in advance if you decide to find time for this.

PR #2361, opened 2026-08-15
Add -STAT_FIT command line option (#1885)

somtri:
Addresses #1885.

This adds `-STAT_FIT {GAUSS|WEIBULL}`, which fits a distribution to the active scalar field of each loaded cloud and prints the fitted parameters. It ports the GUI's 'Compute stat. params' to the command line.

`-STAT_TEST` already exists, but it is the complement of this: it takes the distribution parameters as input and runs a Chi2 test against them. The request in this issue was for the fitting itself, and the thread explicitly ruled out the Local Statistical Test.

You asked in the thread what the output should be:

> it's not clear what the output should be (a text file?)

I went with console output only, formatting the parameters exactly as the GUI does (the line also names the cloud and its scalar field), because the existing `-LOG_FILE` option already writes the console to a file. I tested that: with `-LOG_FILE run.log`, the fitted parameters appear in `run.log`. That covers the "whole data log into a file" the thread asked for without adding a new surface. If you would rather have a dedicated argument, say `-STAT_FIT GAUSS -OUTPUT_FILE {filename}`, that is a small follow-up.

I did not add the RMSE mentioned later in the thread. The GUI's fitting function does not compute one, so producing it here would invent a statistic CloudCompare does not report anywhere else. Worth noting for whoever needs it: for a Gaussian fit the reported std.dev. is the RMS of the residuals about the mean, and the figure that looked inconsistent in the thread was an ICP alignment RMS, which is a different quantity.

On the implementation: the command does not call `ccEntityAction::computeStatParams`. That function builds a modal `ccPickOneElementDlg` as its first statement and reads the chosen distribution back from it, so it cannot run headless without changing a shared GUI signature. `-STAT_FIT` instead mirrors `CommandStatTest`, which sits in the same file and already constructs `NormalDistribution` / `WeibullDistribution` directly. The diff is purely additive: 97 lines added across 4 files, no existing line modified or deleted.

### Verified

Ubuntu 24.04, Qt 6, Release: builds and links with 0 errors, and the files this PR touches compile with no warnings. `ctest` 2/2 passed. `clang-format 18 --dry-run --Werror` and `git diff --check` both exit 0 on all three source files.

The GAUSS path is checked against statistics computed outside CloudCompare. I generated a 20000 point cloud whose scalar field is drawn from a normal distribution, then computed the sample statistics directly: mean 4.994374, std.dev 1.983565. `-STAT_FIT GAUSS` reports:

```
[Distribution fitting] Cloud 'gauss - Cloud' (SF 'TestSF') - Gauss: mean = 4.994374413813 / std.dev. = 1.983565150657
```

Other cases, all run against the built binary:

| Case | Result |
|---|---|
| `-STAT_FIT WEIBULL` | reports `a`, `b`, `shift` plus the extra mode/skewness line, matching the GUI |
| lower-case `-STAT_FIT gauss` | accepted, same result as `GAUSS` |
| cloud with no active scalar field | warns and continues, exit 0, no crash |
| `-STAT_FIT` with no argument | errors |
| `-STAT_FIT BOGUS` | errors, naming the rejected value |
| no cloud loaded | errors |
| `-LOG_FILE run.log` | the fitted parameters appear in `run.log` |
| existing `-STAT_TEST` | unchanged, still prints its Chi2 result |

I have not built this on Windows or macOS locally, so those platforms rest on CI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,10 @@ New features:
 		- New command -PLY_NO_SF_PREFIX
 			- tells the PLY filter not to add the 'scalar_' prefix to the scalar field names when saving
 			- scalar fields coming from an input PLY file already keep their original name
+		- New command -STAT_FIT {GAUSS|WEIBULL}
+			- ports the 'Compute stat. params' tool (distribution fitting) to the command line
+			- fits the distribution on the active scalar field of each loaded cloud (see -SET_ACTIVE_SF)
+			- the fitted parameters are printed to the console, and therefore to the -LOG_FILE file if one is set
 		- New command -DISTANCES_FROM_SENSOR [-SQUARED]
 			- to compute the distances from every point of the cloud to the associated sensor origin (if any)
 		- New command -SCATTERING_ANGLES [-DEGREES]
--- a/qCC/ccCommandLineCommands.cpp
+++ b/qCC/ccCommandLineCommands.cpp
@@ -124,6 +124,7 @@ constexpr char COMMAND_C2C_LOCAL_MODEL[]                  = "MODEL";
 constexpr char COMMAND_C2X_MAX_DISTANCE[]                 = "MAX_DIST";
 constexpr char COMMAND_C2X_OCTREE_LEVEL[]                 = "OCTREE_LEVEL";
 constexpr char COMMAND_STAT_TEST[]                        = "STAT_TEST";
+constexpr char COMMAND_STAT_FIT[]                         = "STAT_FIT";
 constexpr char COMMAND_DELAUNAY[]                         = "DELAUNAY";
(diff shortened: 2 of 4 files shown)


dgirardeau (dismissed):

dgirardeau:
And thanks again!

dgirardeau (approved):

dgirardeau:
Are you able to rebase this branch on master? (I've merged a lot of PRs)

somtri:
Looks like you already merged master in yourself. The branch is green on 9e5ce80 (all six checks) and GitHub reports no conflicts, so nothing is blocking from my side.

If you would rather have a linear history than the merge commit, I can rebase and force-push. I held off because that would drop your merge commit and re-run the full CI matrix. Happy either way.

PR #2363, opened 2026-08-16
Add -MIN_SCALE and -MAX_SCALE options to -MATCH_SCALES (#1034)

somtri:
Addresses #1034.

This adds `-MIN_SCALE {value}` and `-MAX_SCALE {value}` to `-MATCH_SCALES`, constraining the scale factor the tool is allowed to apply. Both are optional and there is no limit by default, so existing command lines behave as before.

You described the semantics in the thread:

> I think the idea was to constrain the scales and clamp the values indeed.

So a factor below `-MIN_SCALE` is clamped up to it, a factor above `-MAX_SCALE` is clamped down to it, and either case logs a warning naming both the original factor and the bound applied. Passing `-MIN_SCALE` greater than `-MAX_SCALE` is rejected before any entity is touched.

On the CCCoreLib question from the same comment: no CCCoreLib change is needed for this. All four algorithms end up as a plain `double` in `ApplyScaleMatchingAlgorithm`, and the factor actually applied is computed in that function:

```cpp
double scaled = algo == ICP_SCALE ? scales[i] : scales[refEntityIndex] / scales[i];
```

The clamp sits immediately after that line, so one guard covers BB_MAX_DIM, BB_VOLUME, PCA_MAX_DIM and ICP. It mirrors the existing `scales[i] <= ZERO_TOLERANCE_D` guard a few lines above it.

One limit worth stating plainly: this bounds the factor that gets applied, it does not constrain the scale search inside ICP. `ICPRegistrationTools::Parameters` has no scale bounds, so steering ICP away from the small-patch local minimum reported in the issue would be a separate change in CCCoreLib.

`minScale` and `maxScale` are defaulted parameters on `ApplyScaleMatchingAlgorithm`, so the `mainwindow.cpp` caller is unchanged and the dialog work stays with you, as you offered.

### Verified

Ubuntu 24.04, Qt 6, Release: builds and links with 0 errors, and no warning names a file this PR touches. `ctest` 2/2 passed. `clang-format 18.1.8 --dry-run --Werror` exits 0 on all three source files.

Behaviour, run against the built binary with a reference cube of span 1 and a second cube of span 3, so the natural factor is 1/3:

| Case | Result |
|---|---|
| no bounds | span 3 rescaled to 1, reference untouched, no warning |
| `-MIN_SCALE 0.5` | 0.333 clamped up to 0.5, span becomes 1.5, warning names the minimum |
| `-MAX_SCALE 0.2` | 0.333 clamped down to 0.2, span becomes 0.6, warning names the maximum |
| `-MIN_SCALE 0.1 -MAX_SCALE 0.5` | factor in range, span becomes 1, no warning |
| `-MAX_SCALE 0.9` | bound above the factor, span becomes 1, no warning |
| `-MIN_SCALE 5 -MAX_SCALE 1` | rejected before any rescaling, non-zero exit |
| `-MIN_SCALE` with no value | errors |
| zero, negative or non-numeric value | rejected at parse time, non-zero exit |
| `-REFERENCE 0 -MIN_SCALE 0.1 -MAX_SCALE 0.5` | existing sub-options still parse |

28 assertions, all passing. I have not built this on Windows or macOS locally, so those platforms rest on CI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,10 +24,11 @@ New features:
 		- distances between a point cloud and a disc can be computed with 'Tools > Distances > Cloud/primitive dist'
 
 	- New Command line options
-		- New command -MATCH_SCALES {BB_MAX_DIM|BB_VOLUME|PCA_MAX_DIM|ICP} [-REFERENCE {index}] [-RMS_DIFF {value}] [-OVERLAP {percent}]
+		- New command -MATCH_SCALES {BB_MAX_DIM|BB_VOLUME|PCA_MAX_DIM|ICP} [-REFERENCE {index}] [-RMS_DIFF {value}] [-OVERLAP {percent}] [-MIN_SCALE {value}] [-MAX_SCALE {value}]
 			- ports the 'Tools > Registration > Match scales' tool to the command line
 			- rescales all loaded clouds/meshes to match the scale of the reference entity (0-based index, 0 by default)
 			- -RMS_DIFF and -OVERLAP only apply to the ICP algorithm (defaults: 1e-5 and 100 respectively)
+			- -MIN_SCALE and -MAX_SCALE constrain the scale factor: a factor falling outside the range is clamped to the nearest bound and a warning is issued (both are optional, and no limit is applied by default)
 		- New command -PLY_NO_SF_PREFIX
 			- tells the PLY filter not to add the 'scalar_' prefix to the scalar field names when saving
 			- scalar fields coming from an input PLY file already keep their original name
--- a/qCC/ccCommandLineCommands.cpp
+++ b/qCC/ccCommandLineCommands.cpp
@@ -3771,9 +3771,9 @@ bool CommandMatchScales::process(ccCommandLineInterface& cmd)
 	unsigned referenceIndex  = 0;
 	double   icpRmsDiff      = 1.0e-5;
 	int      icpFinalOverlap = 100;
-	// a negative bound means 'no limit'
-	double minScale = -1.0;
-	double maxScale = -1.0;
+	// a NaN value means 'no limit'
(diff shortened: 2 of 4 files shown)


dgirardeau:
So my issue with just clamping after ICP has finished running, is that the resulting position with the clamped scale might be suboptimal (compared to the final position that ICP would have reached with the clamped scale). But I now realize that I should probably also expose this scale 'range' parameter in the ICP tool first to be consistent. I'll modify ICP myself.

somtri:
Makes sense, and I won't touch ICP then.

You were right earlier about CCCoreLib: `ICPRegistrationTools::Parameters` only has `adjustScale` today, no bounds, so constraining the scale inside ICP does mean changing it.

The other three methods have no search to constrain, since BB_MAX_DIM, BB_VOLUME and PCA_MAX_DIM are closed-form ratios. Clamping is the whole answer there.

Either way works for this PR. I can hold it until your ICP range parameter lands and then pass -MIN_SCALE/-MAX_SCALE into the ICP parameters for ICP_SCALE, keeping the clamp for the other three. Or close it now and let your change cover both.

dgirardeau:
FYI I've added the option to CCCoreLib (https://github.com/CloudCompare/CCCoreLib/pull/134). I'll warn you when it's merged and CC's submodule pointer is updated.

dgirardeau:
Ok master has now been updated. The part that interest you the most is this one I guess: https://github.com/CloudCompare/CloudCompare/pull/2373/changes#diff-118975ad9a78954af9c80785ac1a7b5c5e2486f5466712dd322e8c01e5a6bf85R4042

Can you rebase this PR and pass the new range parameters to the ICP algorithm? Or do you prefer I do it?

dgirardeau (approved):

Issue #2178, opened 2025-06-02
CloudCompare only imports first 512 columns from wide .xyz files


jeanp-ys:
**Describe the bug**

When importing a .xyz file with more than 512 columns, CloudCompare only imports the first 512 columns. The remaining columns are ignored.

**How to reproduce**

1. Open CloudCompare.
2. Go to File > Open and select `example.xyz` file.

[example.zip](https://github.com/user-attachments/files/20553469/example.zip)

Observe that only the first 512 columns are loaded; the rest are missing. Eg: `verticality 10m` is missing.

**Expected behaviour**

All columns in the .xyz file should be imported, not just the first 512.
Or I should get a warning.

**Your environment**
 - CC Version (or git hash/tag): 2.13.2
 - OS & Version: Ubuntu 24.04
 - Graphics card: nvidia RTX 3500
 - Qt Version (if compiling): Qt 5.15.10

dgirardeau:
512 columns! Indeed, you are pushing this dialog to its limits 😅

jeanp-ys:
Yes, a dialog is not really appropriate for my use case indeed !

capCafu:
The warning part of this shipped in #2364: the dialog now says e.g. "This file has 600 columns, but only the first 512 can be loaded. The remaining 88 column(s) will be ignored.", and logs the same to the console. The 512 limit itself is unchanged.

PR #2364, opened 2026-08-17
Warn the user when an ASCII file has too many columns (#2178)

capCafu:
Addresses #2178.

The ASCII loading dialog stops at 512 columns (`MAX_COLUMNS`). `updateTable()` clamps the per-line column count to that with a `std::min`, and the extra columns are dropped without any message, so a wide file opens looking perfectly normal while some of its fields are simply missing. In the issue that was a `verticality 10m` column that disappeared with nothing to suggest anything had been left out.

You already said the dialog is not really meant to go beyond this:

> 512 columns! Indeed, you are pushing this dialog to its limits 😅

so this does not touch the limit, it only makes the truncation visible. The dialog now keeps track of how many columns the file actually has, and when that is more than `MAX_COLUMNS` it reports how many are being ignored, both in a label above the preview table and in the console. For the file attached to the issue:

> This file has 600 columns, but only the first 512 can be loaded. The remaining 88 column(s) will be ignored.

Notes:
- Default behaviour is unchanged. Files with 512 columns or fewer never show the label and log nothing.
- The label is hidden again at the start of every `updateTable()`, so it clears when the separator or the number of skipped lines changes.
- `MAX_COLUMNS` is untouched, and so is the set of columns that actually get loaded. This only reports the truncation, it does not change it.

Built on Ubuntu 24.04 with GCC 15.3 and Qt 6.9.3 (core build, most plugins off): the tree compiles with no new warnings. I then drove the dialog the way `AsciiFilter::loadStream` does and checked files with 3, 511, 512, 513 and 600 columns — the label only appears above 512, 512 itself stays quiet, and the counts in the message are correct. I have not built this on Windows or macOS.

dialog_warning

### DIFF
--- a/libs/qCC_io/src/AsciiOpenDlg.cpp
+++ b/libs/qCC_io/src/AsciiOpenDlg.cpp
@@ -597,7 +597,7 @@ void AsciiOpenDlg::updateTable()
 	// warn the user if some columns can't be handled (and would be silently ignored otherwise)
 	if (maxPartsCount > MAX_COLUMNS)
 	{
-		QString warning = QString("This file has %1 columns, but only the first %2 can be loaded. The remaining %3 column(s) will be ignored.")
+		QString warning = tr("This file has %1 columns, but only the first %2 can be loaded. The remaining %3 column(s) will be ignored.")
 		                      .arg(maxPartsCount)
 		                      .arg(MAX_COLUMNS)
 		                      .arg(maxPartsCount - MAX_COLUMNS);
--- a/libs/qCC_io/src/AsciiOpenDlg.cpp
+++ b/libs/qCC_io/src/AsciiOpenDlg.cpp
@@ -23,6 +23,7 @@
 #include "FileIOFilter.h"
 
 // qCC_db
+#include <ccLog.h>
 #include <ccPointCloud.h>
 
 // Qt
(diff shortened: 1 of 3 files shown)


libs/qCC_io/src/AsciiOpenDlg.cpp
```diff
@@ -585,6 +594,20 @@ void AsciiOpenDlg::updateTable()
 		m_ui->commentLinesSkippedLabel->setVisible(true);
 	}
 
+	// warn the user if some columns can't be handled (and would be silently ignored otherwise)
+	if (maxPartsCount > MAX_COLUMNS)
+	{
+		QString warning = QString("This file has %1 columns, but only the first %2 can be loaded. The remaining %3 column(s) will be ignored.")
```
dgirardeau:
```suggestion
		QString warning = tr("This file has %1 columns, but only the first %2 can be loaded. The remaining %3 column(s) will be ignored.")
```

dgirardeau: I don't remember if the other strings in this dialog are translatable (but they should probably be)

dgirardeau (commented):
Thanks for the contribution. I just have a small comment if you have the time to fix it.

dgirardeau (approved):

PR #2365, opened 2026-08-18
Make the ASCII open dialog strings translatable

capCafu:
Follow-up to the comment from @dgirardeau on #2364:

> I don't remember if the other strings in this dialog are translatable (but they should probably be)

They were not. The .ui strings are, since uic wraps them in `QCoreApplication::translate`, but everything set from C++ in `AsciiOpenDlg.cpp` was a plain `QString` — after #2364 the truncation warning was the only `tr()` in the file.

This wraps the remaining eight: the two separator labels, the header and comment-lines labels above the preview table, the three messages built in `CheckOpenSequence()`, and the "Error" title of the box that shows them.

Notes:
- `CheckOpenSequence()` is static, but `tr()` is a static member too, so it works there as is.
- The header label was `QString("Header: ") + displayHeader`, now `tr("Header: %1").arg(displayHeader)` so translators can move the placeholder. Output unchanged.
- I left the .ts files alone since lupdate regenerates all ten. Happy to add that here if you prefer.

Built on Ubuntu 24.04 with GCC 15.3 and Qt 6.9.3, no new warnings. Checked the affected labels still read as before ("Header: col0 col1 …", "+ 1 comment line(s) skipped", "(ASCII code: …)"). Not built on Windows or macOS.

### DIFF
--- a/libs/qCC_io/src/AsciiOpenDlg.cpp
+++ b/libs/qCC_io/src/AsciiOpenDlg.cpp
@@ -337,7 +337,7 @@ void AsciiOpenDlg::onSeparatorChange(const QString& separator)
 	assert(separator.size() == 1);
 	if (separator.length() < 1)
 	{
-		m_ui->asciiCodeLabel->setText("Enter a valid character!");
+		m_ui->asciiCodeLabel->setText(tr("Enter a valid character!"));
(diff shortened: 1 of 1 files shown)


dgirardeau (approved):
Great, thanks for the follow up!

Issue #801, opened 2018-10-21
Cloudcompare 2.9.1 for Mac crashes upon opening files


nick-yfeng:
## Current Behavior
Cloudcompare 2.9.1 crashes when I attempt to open files. I tried opening .csv and .txt files. 
I'm using OSX 10.14 and graphic specs are:
NVIDIA GeForce GT 650M 1024 MB
Intel HD Graphics 4000 1536 MB

The crash logs are:
Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000018
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Segmentation fault: 11
Termination Reason:    Namespace SIGNAL, Code 0xb
Terminating Process:   exc handler [64480]

VM Regions Near 0x18:
--> 
    __TEXT                 000000010c89d000-000000010cc0a000 [ 3508K] r-x/rwx SM=COW  /Applications/CloudCompare.app/Contents/MacOS/CloudCompare

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   org.qt-project.QtGui          	0x0000000111d34743 QMacCGContext::QMacCGContext(QPainter*) + 51
1   org.qt-project.QtWidgets      	0x000000011186430b 0x11149a000 + 3973899
2   org.qt-project.QtWidgets      	0x0000000111599fdd QFrame::drawFrame(QPainter*) + 77
3   org.qt-project.QtWidgets      	0x00000001115f70d5 QLabel::paintEvent(QPaintEvent*) + 69
4   org.qt-project.QtWidgets      	0x00000001114e3f1d QWidget::event(QEvent*) + 2141
5   org.qt-project.QtWidgets      	0x000000011159a05d QFrame::event(QEvent*) + 45
6   org.qt-project.QtWidgets      	0x00000001115f706f QLabel::event(QEvent*) + 255
7   org.qt-project.QtWidgets      	0x00000001114abbad QApplicationPrivate::notify_helper(QObject*, QEvent*) + 269
8   org.qt-project.QtWidgets      	0x00000001114ae47c QApplication::notify(QObject*, QEvent*) + 5852
9   org.qt-project.QtCore         	0x000000011215da24 QCoreApplication::notifyInternal2(QObject*, QEvent*) + 164
[...]

PR #2366, opened 2026-08-19
Fix reading of ASCII files with legacy Mac line endings (#801)

capCafu:
Addresses #801.

`QTextStream::readLine()` only breaks on LF, so an ASCII file written with legacy Mac line endings (a lone CR) is read as one very long line. The crash is gone — the 512 column cap takes care of that — but the file still loads wrongly, and silently. The same 12 point cloud, saved two ways:

    lf_only.xyz  -> error=0  points=12
    cr_only.xyz  -> error=0  points=1

@asmaloney suggested reading character by character in the thread. I went a different way: the four `readLine()` calls include the loader's main loop, which runs once per point of every ASCII file, and I didn't want to slow the common case down for a legacy format.

`loadStream()` hands the same `QTextStream` to the dialog and to the loading loop, so there is one place to intervene. `AsciiFilter::loadFile` now peeks at the first 64 KB and, only if it finds a CR and no LF, wraps the `QFile` in a small read-only `QIODevice` that turns CR into LF as the bytes are read. The `readLine()` calls are untouched.

Notes:
- Nothing changes for files that already load: the device isn't created unless the sniff matches, and CRLF files always contain a LF.
- Replacing every CR is safe because such files contain no LF — the same condition the detection tests for.
- `loadAsciiData()` already has its data in memory, so there it's a plain `replace()`.

Built on Ubuntu 24.04 with GCC 15.3 and Qt 6.9.3, no new warnings. Loaded the same cloud saved as LF, CR only and CRLF: before the change the CR file gave 1 point, after it all three give 12, with identical coordinates. Not built on Windows or macOS.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -372,6 +372,7 @@ Improvements:
 			(only the first 512 columns are loaded, the other ones were previously ignored silently)
 
 Bug fixes:
+	- ASCII files saved with legacy Mac line endings (a lone CR) were read as a single line, silently loading only one point
 	- the weights derived from normals comparison during ICP registration of 2 clouds could be wrong (the wrong normals were compared)
 	- editing the Global Shift & Scale information of a polyline would make CC crash
 	- segmenting a cloud with polylines depending on it but not directly present below the cloud entity in the DB tree could lead
--- a/libs/qCC_io/src/AsciiFilter.cpp
+++ b/libs/qCC_io/src/AsciiFilter.cpp
@@ -36,6 +36,7 @@
 #include <ccScalarField.h>
 
 // System
+#include <algorithm>
 #include <cassert>
 #include <cstring>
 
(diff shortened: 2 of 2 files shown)


capCafu:
The red jobs here aren't coming from this change — they all stop at `Install Dependencies`:

```
Error: getaddrinfo ENOTFOUND repo.anaconda.com
Error: Invalid miniconda version!
```

`Configure CMake`, `Check Format` and `Build` are all skipped in those jobs, so nothing in this PR actually got compiled. macOS fails fast on the DNS error, while the two Linux jobs and CodeQL keep retrying until the 6h timeout. Windows MSVC passes because it uses the pre-installed `C:\Miniconda\envs\CloudCompareDev` environment and never hits the network.

That also means formatting is fine now: `Check Format` runs inside the Windows job, and it's green. The earlier run on the same tree (64b8801, byte-identical content) had Ubuntu Clang, macOS Clang and Windows MSVC all passing.

Happy to push again once repo.anaconda.com is reachable, but I don't think there's anything to fix on my side.

dgirardeau (approved):

dgirardeau:
Would you be able to share such a file with me? (for testing purpose in the future)

capCafu:
Attached — lidar scan, 5335 points (X Y Z intensity), CR line endings only.
[ascii_cr_only.zip](https://github.com/user-attachments/files/31319485/ascii_cr_only.zip)

dgirardeau:
Thanks

PR #2372, opened 2026-08-21
Make the ASCII column type names translatable

capCafu:
Follow-up to #2365. That one made the strings in `AsciiOpenDlg.cpp` translatable but missed the ones people look at most: the column type names ("coord. X", "Scalar", and so on) that fill the combo box under every column of the preview table. They also get pasted into the `'%1' defined at least twice!` message, so that message was translated while the name inside it stayed in English.

They're a `const char[][20]` array in the header, so lupdate needs `QT_TRANSLATE_NOOP` on the declaration, plus a lookup where they're used.

Notes:
- `QT_TRANSLATE_NOOP` just expands to the literal, so the array keeps its type and size.
- I pinned the context to `AsciiOpenDlg` so it matches the `tr()` calls and stays with the rest of the dialog.
- With no translation available the labels read exactly as before.

Built on Ubuntu 24.04 with GCC 15.3 and Qt 6.9.3, no new warnings. lupdate now picks the labels up in the `AsciiOpenDlg` context (9 -> 33 messages for libs/qCC_io), and with a small .qm translating two of them the combo boxes show the translated text while the rest stay English. Not built on Windows or macOS.

### DIFF
--- a/libs/qCC_io/include/AsciiOpenDlg.h
+++ b/libs/qCC_io/include/AsciiOpenDlg.h
@@ -22,6 +22,7 @@
 #include "qCC_io.h"
 
 // Qt
+#include <QCoreApplication>
 #include <QDialog>
 #include <QString>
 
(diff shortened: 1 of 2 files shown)


dgirardeau (approved):
Nice catch

Issue #2358, opened 2026-08-04
SOR progress remains at 0% and CloudCompare becomes unresponsive during processing


pointknown:
Hi,

First, thank you to everyone who develops and maintains CloudCompare. We use it regularly in our existing-conditions and point-cloud workflow, and it is an extremely valuable tool.

I am using CloudCompare 2.14 beta on Windows 11 and noticed some behavior during the SOR filter that I wanted to report. I am not sure whether this is expected behavior, a regression, or an issue specific to my installation.

When I run SOR on a large cloud, the progress indicator remains at 0% for the entire calculation. The CloudCompare window then becomes gray, and Windows displays “Not Responding.”

The filter eventually completes successfully, and CloudCompare becomes responsive again. However, there is no visible indication of progress while it is running, so it is difficult to tell whether the calculation is continuing or whether the application has actually frozen.

I also normally work with several clouds from the same project loaded at once. My usual workflow would be to let SOR process one cloud in the background while reviewing, cropping, or working on another cloud. When I try to interact with another cloud during SOR processing, the entire CloudCompare interface becomes unresponsive, so no other work can be performed in that session until SOR finishes.

Test information:

- CloudCompare 2.14 beta, Windows 64-bit build dated July 22, 2026
- Windows 11
- Approximately 96 million points in the cloud
- Several clouds loaded in the same session
- CloudCompare used approximately 16% total CPU and 17 GB of memory during the calculation
- Overall workstation CPU and memory usage remained relatively low
- SOR completed successfully despite remaining at 0% and showing “Not Responding”

Would it be possible for the SOR progress indicator to update during processing and for the interface to remain responsive enough to display progress or allow cancellation?

It would also be very helpful if lengthy filters could run in the background while the user works with other loaded clouds. I understand that this may be a feature request rather than a bug.

I can provide screenshots or additional system information if useful.

Thank you.

dgirardeau:
Yes, sadly it seems the GUI doesn't have the time to refresh itself during the computation...

In the latest version (online today) I have added the option to set the number of threads used for the computation. It definitely helps to make Windows and the other applications running in parallel to CC more responsive. But it doesn't seem to make CC's progress bar more responsive. I'll try to think about something, but for now nothing obvious come to mind...

capCafu:
Had a look at this one. I don't think the progress bar is the problem. The value gets updated fine. The GUI just never gets a chance to redraw.

[...]

PR #2375, opened 2026-08-25
Run the SOR filter in a worker thread to keep the GUI responsive (#2358)

capCafu:
Addresses #2358, carrying on from the discussion there.

`ccBackgroundTask::Run()` takes a function, runs it on a worker thread, and spins a local event loop until it comes back. The call only returns once the function is done, so callers don't need rewriting. I put it in CCPluginAPI as its own header rather than in `ccQtHelpers.h`, so nothing that already includes that picks up a QtConcurrent dependency it didn't have.

The new file uses the `COPYRIGHT: CloudCompare project` banner, the same one `ccArgumentParser.h` has, rather than the older EDF R&D / TELECOM ParisTech one that most of the files in that folder still carry. Let me know if you'd rather it matched the neighbours instead.

The SOR call site ends up like this:

```cpp
CCCoreLib::ReferenceCloud* selection = ccBackgroundTask::Run(
    [&]()
    {
	    return CCCoreLib::CloudSamplingTools::sorFilter(cloud, s_sorFilterKnn, s_sorFilterNSigma,
	                                                    cloud->getOctree().data(), &pDlg, s_maxThreadCount);
    });
```

I had to touch `ccProgressDialog` as well. Algorithms call `start()` and `stop()` on the callback, and once the work happens on a worker thread those calls arrive there too, so `show()` and `hide()` were being called from the wrong thread. Qt complains about it ("Timers cannot be stopped from another thread"). They now hop over to the GUI thread the same way `update()` already did.

To see whether the event loop really keeps running, I put a timer on the GUI thread during a SOR over 3M points and counted the ticks:

| | heartbeats | took | points kept |
| --- | --- | --- | --- |
| before | 1 of a possible 19 | 973 ms | 2525966 |
| after | 19 of 19 | 960 ms | 2525966 |

Same output, same time.

I also tried it in the GUI on a 6M point cloud, and two things came up.

The bar now visibly resets once partway through. That's just the octree being built first and then the filter itself, two phases that each go from 0 to 100. It always did that, you simply couldn't see it before because the bar never moved.

There's also still a short freeze right at the end. That one is `cloud->partialClone(selection)` copying the surviving points, which runs after the call above and is still on the GUI thread. It has no progress callback so I left it alone, but it could go through the same helper if you'd rather the whole action stayed responsive.

Two things I'd like your view on.

The rest of the app is live now while SOR runs. That's the whole point, but it does mean someone could start something else halfway through. The progress dialog goes out of its way not to take focus (`WA_ShowWithoutActivating`), so making it modal felt like something to ask about rather than just do.

And I only did SOR here, like you suggested. The noise filter, C2C distances and M3C2 would suit the same thing, and the hand written wait loops in qPoissonRecon, qRANSAC_SD, qMeshBoolean, qCork and qPCL could fold into it as well. Happy to do those as follow ups once you're happy with how this looks.

### DIFF
--- a/libs/CCPluginAPI/include/CMakeLists.txt
+++ b/libs/CCPluginAPI/include/CMakeLists.txt
@@ -3,6 +3,7 @@ target_sources( ${PROJECT_NAME}
 	PRIVATE
 		${CMAKE_CURRENT_LIST_DIR}/CCPluginAPI.h
 		${CMAKE_CURRENT_LIST_DIR}/ccArgumentParser.h
+		${CMAKE_CURRENT_LIST_DIR}/ccBackgroundTask.h
 		${CMAKE_CURRENT_LIST_DIR}/ccColorScaleSelector.h
 		${CMAKE_CURRENT_LIST_DIR}/ccColorScaleEditorDlg.h
 		${CMAKE_CURRENT_LIST_DIR}/ccColorScaleEditorWidget.h
--- a/libs/CCPluginAPI/include/ccBackgroundTask.h
+++ b/libs/CCPluginAPI/include/ccBackgroundTask.h
@@ -1,0 +1,70 @@
+#pragma once
+// ##########################################################################
+// #                                                                        #
+// #                            CLOUDCOMPARE                                #
+// #                                                                        #
+// #  This program is free software; you can redistribute it and/or modify  #
+// #  it under the terms of the GNU General Public License as published by  #
+// #  the Free Software Foundation; version 2 of the License.               #
+// #                                                                        #
+// #  This program is distributed in the hope that it will be useful,       #
+// #  but WITHOUT ANY WARRANTY; without even the implied warranty of        #
+// #  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         #
+// #  GNU General Public License for more details.                          #
+// #                                                                        #
+// #                   COPYRIGHT: CloudCompare project                      #
+// #                                                                        #
+// ##########################################################################
+
+// Qt
+#include <QEventLoop>
+#include <QFutureWatcher>
+#include <QtConcurrentRun>
+
+// System
+#include <type_traits>
+#include <utility>
+
+//! Helper to run a long computation without freezing the GUI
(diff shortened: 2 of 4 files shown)


dgirardeau (dismissed):

dgirardeau:
This looks excellent. I'll try to quickly test it first on Windows before merging though.

However I guess you can expand it to other methods if you feel like it.

capCafu:
Thanks. I'll focus on the noise filter and C2C distances next.

dgirardeau:
So on Windows (Visual Studio, debug mode) I get this assertion:
image

I'll investigate.

dgirardeau:
Ok it was simple, basically all methods of ccProgressDialog should use a deferred signal when called outside of the main thread. I'll let you check on your side that things still work.

capCafu:
Ran into two things with that on Linux. Dragging a file in gives no progress bar at all, and Compute Normals just sits at 0%. Anything still running on the main thread, basically.

The main thread branch in `update()` is inside `#if defined(CC_WINDOWS)`, so on Linux everything goes down the deferred path, and the `processEvents()` that used to sit after the `singleShot` isn't there any more. The main thread is busy in the computation, so the queued refresh never gets a chance to run.

That bit is my fault. I called that `processEvents()` dead code in the issue, which is only true when a worker thread calls it. On the main thread it was the thing making the dialog repaint.

`setInfo()` lost its `setLabelText()` on the main thread branch too, so the text went as well.

I've pushed a commit that makes the main thread branch unconditional and puts `processEvents()` back into it, keeping your deferred path for the worker case. Reading the dialog back from both threads:

```
before:  main   label=''           bar=-1
after:   main   label='INFO_MAIN'  bar=80
```

The worker side was fine either way. SOR, the noise filter and C2C all still behave here.

dgirardeau:
Ok, makes sense. Thanks for the fixes.

I'll see if I can remove this Sleep(1) on Windows, might be a legacy thing.

dgirardeau (approved):

dgirardeau:
Thanks for the support, the result is great!

PR #2377, opened 2026-08-27
Run the noise filter and distance computations in a worker thread (#2358)

capCafu:
Same thing as #2375, applied to the other two.

The noise filter was easy, it's the same shape as SOR.

Distances took a bit more thought. I first wrapped the whole `switch (m_compType)`, then noticed the cases read widgets while setting their parameters up, like `lmKNNSpinBox->value()`. That would have meant touching Qt widgets from a worker thread, so now only the two computation calls are wrapped and the setup stays where it was. Cloud to mesh came along for free, it's in the same switch.

Checked both the same way as before. The GUI keeps redrawing during the computation and the results come out identical.

This sits on top of #2375, so it shows that commit too until it merges.

### DIFF
--- a/libs/qCC_db/include/ccPointCloud.h
+++ b/libs/qCC_db/include/ccPointCloud.h
@@ -707,7 +707,6 @@ class QCC_DB_LIB_API ccPointCloud : public CCCoreLib::PointCloudTpl<ccGenericPoi
 		bool             applyToSFduringRGB      = false;
 		RGB_FILTER_TYPES filterType              = RGB_FILTER_TYPES::NONE;
 		unsigned char    burntOutColorThreshold  = 0;
-		bool             commandLine             = false;
 		double           sigmaSF                 = -1;
 		double           spatialSigma            = -1;
 		bool             blendGrayscale          = false;
--- a/qCC/ccCommandLineCommands.cpp
+++ b/qCC/ccCommandLineCommands.cpp
@@ -6602,7 +6602,6 @@ bool CommandFilter::process(ccCommandLineInterface& cmd)
 	bool applyToSF  = false;
 	bool gaussian   = false;
 	ccPointCloud::RgbFilterOptions(filterParams);
-	filterParams.commandLine = true;
 	while (!cmd.arguments().empty())
 	{
 		QString argument = cmd.arguments().front();
@@ -6753,11 +6752,11 @@ bool CommandFilter::process(ccCommandLineInterface& cmd)
 	}
 	if (applyToSF)
 	{
-		return ccEntityAction::sfGaussianFilter(selectedEntities, filterParams, cmd.widgetParent());
+		return ccEntityAction::sfGaussianFilter(selectedEntities, filterParams, cmd.widgetParent(), true);
 	}
 	else if (applyToRGB)
 	{
-		return ccEntityAction::rgbGaussianFilter(selectedEntities, filterParams, cmd.widgetParent());
+		return ccEntityAction::rgbGaussianFilter(selectedEntities, filterParams, cmd.widgetParent(), true);
 	}
 
 	return true;
--- a/qCC/ccComparisonDlg.cpp
+++ b/qCC/ccComparisonDlg.cpp
@@ -29,6 +29,7 @@
 #include <ScalarFieldTools.h>
 
 // qCC_db
+#include <ccBackgroundTask.h>
 #include <ccGBLSensor.h>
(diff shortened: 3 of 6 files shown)


capCafu:
Moved this to draft while #2375 settles, I'll rebase it once that's merged.

fallenmi (changes requested):
`computeOctree()` now moves GUI and model operations into the worker thread along with the expensive calculation.

On exact head `70654a8a40080a640c6f2c0dd6764be07ae803aa`, `MainWindow::doActionComputeOctree()` passes `this`, and `ccBackgroundTask::Run()` executes the lambda with `QtConcurrent::run`. Inside that lambda, `coDlg.getMode()` and `getMinCellSize()` dereference `QRadioButton` / `QDoubleSpinBox` children of the GUI-thread `QDialog`. The same lambda also calls `removeObjectTemporarilyFromDBTree()` and `putObjectBackIntoDBTree()`; the real `MainWindow` implementations forward to `ccDBRoot::removeElement()` / `addElement()`. `ccDBRoot` is a `QAbstractItemModel`: those methods call `beginRemoveRows()` / `beginInsertRows()` and touch the attached views (`addElement(..., false)` still calls `m_dbTreeWidget->expand()`).

Qt documents that [`QWidget` subclasses can only be used from the main thread](https://doc.qt.io/qt-6/threads-qobject.html) and that [`QAbstractItemModel` APIs must run in the model's owning thread](https://doc.qt.io/qt-6/qabstractitemmodel.html#thread-safety); background changes must be queued back to the main thread. The nested GUI event loop remains active here, so these wrong-thread calls leave the shipped action in undefined behavior.

An exact-revision call-chain oracle passes base `27fdb3ccb7c67f8f7b0889f52ee7ba9613a6d693` (no worker lambda) and deterministically fails head and current synthetic merge `3a93e9ffb09d64ecf711dcf7db6405e120c58a9f`: two worker-side `getMode()` calls, one `getMinCellSize()`, one DB removal, and one DB insertion. The head result reproduced 10/10. Current exact-head CI is 6/6 green, but it does not execute this GUI action or check object affinity.

Please snapshot all dialog values on the GUI thread, keep DB-tree detach/reattach and model/view updates there, and restrict the worker lambda to the expensive non-GUI computation (queuing any resulting GUI/model updates back to the owning thread).

AI disclosure: This review was prepared with OpenAI Codex assistance for exact-diff and call-graph inspection, base/head/merge contract-oracle execution, live CI/head/interaction verification, primary Qt documentation checking, and drafting. The exact revisions and call chain above are included for independent verification.

capCafu:
Had a look through the converted call sites. Most of them are fine, but the octree one in `ccEntityAction.cpp` reads the dialog and moves things around the DB tree from inside the worker thread.

`coDlg.getMode()` and `coDlg.getMinCellSize()` both read widgets:

```cpp
if (cellSizeRadioButton->isChecked()) return MIN_CELL_SIZE;   // getMode()
return cellSizeDoubleSpinBox->value();                        // getMinCellSize()
```

and `removeObjectTemporarilyFromDBTree` / `putObjectBackIntoDBTree` end up calling `m_ccRoot->removeElement()`, so the tree model gets modified off the GUI thread while its views are attached.

I've pushed a commit that moves the loop out of the lambda: the dialog values are read once up front, the detach and reattach stay on the GUI thread, and only `computeOctree` / `octree->build` runs in the worker. Same behaviour, and it also drops a `switch` that was being evaluated twice per cloud.

The only other thing I noticed is `prepareDisplayForRefresh()` being called from the worker in three places, but that just sets flags on the GL window, so I left it.

dgirardeau:
Ok, that's probably better, but do you think it's really an issue to call the getters from a dialog? As long as no display happens, I guess we should be good? (at least I didn't get any warning from Qt).

capCafu:
Yeah, you're probably right about the getters. They're just reading members, and the dialog's closed by then so nothing's changing underneath. Qt won't warn either, it only really complains about timers, painting and creating widgets off thread. I wouldn't have touched that on its own.

The bit I'd keep is the DB tree. `removeObjectTemporarilyFromDBTree` ends up calling `m_ccRoot->removeElement()`, so the item model gets changed while its views are still attached, and Qt doesn't warn about that one either. That's the sort of thing that turns up later as a confused tree rather than a crash you can trace back to anything.

The getters just came along for the ride when I moved the loop out. Side benefit is the switch only gets evaluated once per cloud now instead of twice.

dgirardeau (approved):

Issue #2290, opened 2026-03-09
CANUPO Classification freezes CloudCompare


EZChe-EZDataMD:
**Describe the bug**

After running the classification (Plugins->CANUPO->classify), the mainwindow of CloudCompare is frozen (Not Responding). It appears that the classification process is completed with CANUPO.class and CANUPO.confidence scalar fields being added. Not sure if it is a function of the data size (I have ~388k points), it took a few minutes for it to come back alive. Then switching to any other scalar fields will freeze the program for a few minutes again. Everything works as normal before running CANUPO classification. 

**How to reproduce**

**Expected behaviour**

**Additional context**

**Your environment**
 - CC Version (or git hash/tag): v2.14 beta (Feb 20, 2026)
 - OS & Version: Windows 11
 - Graphics card: NVIDIA GeForce RTX4070 Laptop GPU
 - Qt Version (if compiling):

dgirardeau:
So it doesn't surprise me, considering how much CPU power you need to compute the descriptors. Especially if you used a lot of different scale values for instance.

Mind that you can reduce the number of threads used (at the bottom of the 'CANUPO Classification' dialog) to prevent the computer from freezing while computing things.

EZChe-EZDataMD:
Thanks. Just to clarify, the freezing took place after the classification as described, not during the classification process itself.

________________________________

dgirardeau:
Oh sorry, I did not understand that.

400k points is definitely not a big number.

Maybe you can play with the 'Enable shader for faster display...' option in the 'Display > Display settings':

<img width="420" height="198" alt="Image" src="https://github.com/user-attachments/assets/135e3b9a-cdad-4f99-a49c-95f42c6a1909" />

EZChe-EZDataMD:
No worries at all!. I’ll play around with settings and see.
But in general CC can handle many more points with many more attributes with no issues. Adding two more scalar fields shouldn’t add so much burden to the machine. I almost wonder if those scalar fields are dynamic and got some update when they become active, or some memory leak or extra memory allocation triggered by Canupo classification.

Best,
Ezra

________________________________

dgirardeau:
I looked at this on my side, and I saw that while the dialog disappears (and the display is updated), the computation is in fact not finished. So you can attempt to interact with the cloud, but everything is super slow. You have to wait until the message "MSC descriptors have been saved as meta-data (cloud 'XXX')" appears.

Is it the same thing on your side maybe?

[...]

PR #2380, opened 2026-08-31
Use ccBackgroundTask in the qPoissonRecon and qCanupo plugins (#2358)

capCafu:
Carrying on from #2375 and #2377, this time in the plugins. Two of the six that still have their own wait loop.

qPoissonRecon had this:

```cpp
QFuture<bool> future = QtConcurrent::run(doReconstruct);
while (!future.isFinished())
{
#if defined(CC_WINDOWS)
	::Sleep(500);
#else
	usleep(500 * 1000);
#endif
	pDlg.setValue(pDlg.value() + 1);
	QApplication::processEvents();
}
result = future.result();
```

which turns into `result = ccBackgroundTask::Run(doReconstruct);`. That also gets rid of the made up progress, since `pDlg.setValue(pDlg.value() + 1)` was just adding one to the bar every half second no matter what the computation was actually doing.

qCanupo is a bit different. It uses `QtConcurrent::map` rather than `run`, and there's already a comment in there saying a blocking map would block the UI. Wrapping it in `Run` would have meant blocking inside a thread from the same pool, which can deadlock if the user turns the thread count down, so I added a small `Wait()` alongside `Run()` for futures that have already been started:

```cpp
auto future = QtConcurrent::map(corePointsIndexes, ComputeCorePointDescriptor);
ccBackgroundTask::Wait(future);
```

I left the rest of both files alone. `plugins/core/Standard` isn't in the format target, so running clang-format over them would have added a few hundred lines of noise.

Only these two are switched on in my build, so the other four (qCork, qMeshBoolean, qPCL, qRANSAC_SD) are the same pattern but I can't compile them here. Happy to do those as well if you'd rather have them all together.

For what it's worth, #2290 was a CANUPO freeze report that got closed without a fix.

### DIFF
--- a/libs/CCPluginAPI/include/ccBackgroundTask.h
+++ b/libs/CCPluginAPI/include/ccBackgroundTask.h
@@ -67,4 +67,24 @@ class ccBackgroundTask
 			return watcher.result();
 		}
 	}
+
+	//! Waits for an already started task, while the GUI keeps refreshing
+	/** Same idea as Run, but for code that has already launched something with
+	    QtConcurrent (a map, for instance). Waiting on it directly would block the
+	    GUI thread, and wrapping it in Run would mean blocking inside a thread of
+	    the same pool, which can deadlock when few threads are allowed.
+
+	    \param future the task to wait for
+	**/
+	template <typename T>
+	static void Wait(QFuture<T>& future)
+	{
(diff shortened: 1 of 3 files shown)


dgirardeau (approved):

dgirardeau:
Thanks again!

PR #2382, opened 2026-09-04
Remove deprecated items /  Re-enable warnings on GCC/Clang

rjanvier:
Warnings with Unix compilers (clang and GCC) have been disabled for a long time because they produce overly verbose output in the CI.... But as a side effect, we miss warnings about deprecation (and other potentially useful warnings)

This PR re enables warnings and tries to fix as much as possible our deprecation warnings.


libs/qCC_glWindow/include/ccGLWindowInterface.h
```diff
@@ -199,6 +199,8 @@ class CCGLWINDOW_LIB_API ccGLWindowInterface : public ccGenericGLDisplay
 	{
 		return m_viewportParams;
 	}
+	// FIXME(RJ): temporary helper for migration
+	QPointF toCenteredGLCoordinates(const QPointF& coordinates) const;
```
dgirardeau: Is it really temporary? Or should we remove the other version?

dgirardeau (approved):

PR #2383, opened 2026-09-05
Use ccBackgroundTask in the qRANSAC_SD, qPCL and qM3C2 plugins (#2358)

capCafu:
Three more after #2380.

qRANSAC_SD and qPCL's `BaseFilter` both used `QtConcurrent::run` with a polling loop, so each one becomes a single `ccBackgroundTask::Run(...)`.

The qPCL one is worth a mention, because its loop had no `processEvents()` in it at all:

```cpp
QFuture<void> future = QtConcurrent::run(DoCompute);
int progress = 0;
while (!future.isFinished())
{
#if defined(CC_WINDOWS)
	::Sleep(500);
#else
	usleep(500 * 1000);
#endif
	if (m_showProgress)
	{
		pDlg.setValue(++progress);
	}
}
```

so while a PCL filter was running the window had no chance to repaint between the sleeps.

I also did qM3C2, which is a different case. It wasn't polling at all, it was calling `QtConcurrent::blockingMap` directly in three places, so the GUI thread was blocked outright with no progress of any kind. Same thing you and I looked at in DgmOctree for #2358. I used `Wait()` rather than wrapping the map in `Run()`, to avoid blocking inside a thread of the same pool.

While I was looking around I also checked qPDALIO, since it calls `waitForFinished()`. That one is already fine: it sets up a `QFutureWatcher`, then runs `pDlg->exec()`, so the dialog's own event loop keeps things alive and `waitForFinished()` has nothing left to wait for by the time it's reached. Same idea as `ccBackgroundTask`, just written out by hand. Left it alone.

That leaves qCork and qMeshBoolean. Same pattern as the others, but they want Cork and libIGL and I couldn't get hold of either here, so I haven't touched them. Looks like they're off in pixi.toml too.

Rest of the plugin files untouched, same as last time, since `plugins/core/Standard` isn't in the format target.

Built on Ubuntu 24.04 with GCC 15.3 and Qt 6.9.3, with QRANSAC_SD, QPCL and QM3C2 enabled, no new warnings. Not built on Windows or macOS.

### DIFF
--- a/plugins/core/Standard/qM3C2/src/qM3C2Process.cpp
+++ b/plugins/core/Standard/qM3C2/src/qM3C2Process.cpp
@@ -26,6 +26,7 @@
 
 //qCC_plugins
 #include <ccMainAppInterface.h>
+#include <ccBackgroundTask.h>
 #include <ccQtHelpers.h>
 
 //qCC_db
@@ -997,7 +998,8 @@ bool qM3C2Process::Compute(const qM3C2Dialog& dlg, QString& errorMessage, ccPoin
 				}
 				assert(maxThreadCount > 0 && maxThreadCount <= QThread::idealThreadCount());
 				QThreadPool::globalInstance()->setMaxThreadCount(maxThreadCount);
-				QtConcurrent::blockingMap(pointIndexes, ComputeM3C2DistForPoint);
+				auto future = QtConcurrent::map(pointIndexes, ComputeM3C2DistForPoint);
+				ccBackgroundTask::Wait(future);
 			}
 			else
 			{
--- a/plugins/core/Standard/qM3C2/src/qM3C2Tools.cpp
+++ b/plugins/core/Standard/qM3C2/src/qM3C2Tools.cpp
@@ -34,6 +34,7 @@
 
 //qCC
 #include <ccMainAppInterface.h>
+#include <ccBackgroundTask.h>
 #include <ccQtHelpers.h>
 
 //Qt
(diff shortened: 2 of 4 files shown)


dgirardeau (approved):

dgirardeau:
Thanks again!

Yes, the PCL filter is rather old (and also has a unique architecture among the plugins).

I'll try to patch the qCork and qMeshBoolean plugins myself when I'll find the time.

capCafu:
Thanks. Sounds good on qCork and qMeshBoolean, I'll leave those with you.

One thing I spotted while poking around: `BinFilter.cpp` has two of the same loops, in the BIN save and load. They're not as bad as the plugin ones, since they do call `processEvents()` every 500 ms, so it's more choppy than frozen, but it's the same duplicated `#ifdef` sleep.

The awkward bit is that `qCC_io` only links `QCC_DB_LIB`, not `CCPluginAPI`, so `ccBackgroundTask.h` isn't reachable from there. Moving it into `qCC_db` next to `ccProgressDialog` sorts that out and nothing else needs changing, since everything already has that include path. I tried it locally and all the plugins still build fine.

Happy to send a PR if you like the sound of it, or drop it if you'd rather not move the header around.

dgirardeau:
Ah yes, I stumbled on the same issue. I think it's ok to move this piece of code to qCC_db, even if it's a bit weird. It's probably worth the weirdness ;)

Go ahead, thanks.

Issue #2164, opened 2025-05-11
Progress dialog for some operations doesn't work


tpwrules:
**Describe the bug**
For some operations on some systems, e.g. number of neighbors computation on macOS or Linux, the progress dialog doesn't show up and the application freezes completely until the operation finishes. These operations work fine on Windows.

Other operations affected include the noise filters and probably many others

**How to reproduce**
1. Open a cloud with e.g. 100M points
2. Perform a number of neighbor computation (Tools -> Other -> Compute geometric features ->  Check "number of neighbors")
3. Observe that the program becomes completely frozen for several minutes after the octree computation is completed
4. Observe that the operation does eventually complete successfully

**Expected behaviour**
I expect the program to not freeze during step 3 and the progress dialog to appear and show an accurate progress percent.

**Additional context**
On my Linux machine, the progress dialog appears but is always just black. On macOS the progress dialog never appears. On Windows 10 the progress dialog appears and works as expected.

**Your environment**
 - CC Version (or git hash/tag): 2.13.2 / macOS build 20250417 (not sure which commit)
 - OS & Version: Windows 10 / Ubuntu 18.04 / macOS 15.2
 - Graphics card: Nvidia / Nvidia / Apple Silicon
 - Qt Version (if compiling):

dgirardeau:
Have you been able to test the 2.14.alpha versions on Linux or macOS?

tpwrules:
I tested [this macOS build](https://www.simulation.openfields.fr/index.php/cloudcompare-downloads/7-cloudcompare-macos-binaries/119-cloudcompare-20250417-2-14-alpha-for-macos-15-4-arm64-without-python-plugin). I also tested the 2.13.2 version from that same site. There is no change, both are broken.

I also built and tested [this commit](https://github.com/CloudCompare/CloudCompare/commit/ff7926cd9462a2570c8c9863bd4cb0f06358342f) on my Linux machine. Still no change from v2.13.2 I built myself.

I only tested 2.13.2 on Windows (same machine that I use Linux, just dual booted, so hardware is identical).

I tried some kind of random hacks in `ccProgressDialog.cpp` with sprinkling in `QCoreApplication::processEvents()` and `QProgressDialog:update()` calls, plus `setWindowModality(Qt::WindowModal)` but was unable to fix the problem. Not 100% clear to me if that is even the dialog I am looking at... The octree build dialog works fine and smoothly updates progress on all three operating systems on all versions.

[...]

PR #2388, opened 2026-09-08
Run the geometric feature and SF gradient computations in a worker thread (#2164)

capCafu:
Addresses #2164.

`ComputeCharactersitic` runs on the GUI thread and ends up in `blockingMap`, same as the SOR and noise filters were. It goes through `ccBackgroundTask::Run` now. I did the SF gradient too, same path.

I put a timer on the GUI thread ticking every 50 ms and ran it on 6 million points. 1 tick out of 34 before, 39 out of 39 after.

That explains the three symptoms in the report. The one `processEvents()` before the blocking map is inside `#ifndef __APPLE__`, so on Linux the dialog appears and then never repaints, and on macOS it never appears.

The octree stays on the GUI thread. `computeOctree` calls `addChild` and `removeChild`, and the DB tree is a Qt model. It already works anyway.

`ccAlignDlg` calls `ComputeLocalDensityApprox` the same way. Different tool and a small cloud, so I left it. Say if you want it.

Built on Ubuntu 24.04 with GCC 15.3 and Qt 6.9.3, no new warnings. Checked both in the GUI. Not built on Windows or macOS.

### DIFF
--- a/qCC/ccLibAlgorithms.cpp
+++ b/qCC/ccLibAlgorithms.cpp
@@ -21,6 +21,7 @@
 #include <ScalarFieldTools.h>
 
 // qCC_db
+#include <ccBackgroundTask.h>
 #include <ccOctree.h>
 #include <ccPointCloud.h>
 #include <ccScalarField.h>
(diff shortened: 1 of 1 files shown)


dgirardeau (approved):
Thanks again!

Issue #2166, opened 2025-05-11
Approximate density / nearest neighbor distance


tpwrules:
**Describe the feature you would like**
I would like to know the approximate density or nearest neighbor distance from a point to its closest neighbor for fast noise removal.

**Describe alternatives you've considered**
Older CloudCompare versions had an approximate mode for the density dialog but this seems to have been lost when it was folded into the "compute geometric features" dialog?

**Is your feature request related to a problem? Please describe.**
The SoR and noise removal functions are both very slow on clouds with >100M points. For a scanner which has a habit of outputting random points very far from everything; this feature would be a fast way to identify those.

dgirardeau:
I've added this to the TODO list.

PR #2396, opened 2026-09-13
Expose the approximate local density in the geometric features dialog

capCafu:
Addresses #2166.

The approximate density already exists, it just never made it into the dialog. There is an `-APPROX_DENSITY` command, so right now the command line is the only way to reach it. This adds the three checkboxes to the Density group.

It is worth having because it ignores the radius. On 6 million points:

| radius | exact | approx |
|---|---|---|
| 0.5 | 2.8 s | 1.1 s |
| 1.0 | 8.3 s | 1.0 s |
| 2.0 | 34.9 s | 1.2 s |

It only ever looks at the nearest neighbour, so it stays flat while the exact one grows.

The first commit fixes two names, both visible from the CLI already. The field name had `(r=0)` in it, for a radius this mode ignores. And the KNN one was called "Number of neighbors", but CCCoreLib says only one neighbour is extracted, so it is really the inverse distance to it, and the values are not integers. It is now "Inverse of the nearest neighbor distance", which is what the reporter asked for anyway.

I added all three variants since the backend has them. Say if you would rather keep only the nearest neighbour one.

Built on Ubuntu 24.04 with GCC 15.3 and Qt 6.9.3, no new warnings. Checked in the GUI, and the CLI field names before and after. Not built on Windows or macOS.


dgirardeau:
Looks good. Can you update the CHANGELOG.md file? I guess in the 'improvements' section.

dgirardeau (approved):

PR #2398, opened 2026-09-13
Use ccBackgroundTask in the qCork and qMeshBoolean plugins

dgirardeau:
@capCafu: as discussed in #2358


capCafu:
One small thing, qCork sets the progress dialog modal but qMeshBoolean doesn't. Was that deliberate?

The qCork one is also after `show()`, so it doesn't take effect. `isModal()` returns true either way, but `activeModalWidget()` stays null unless `setModal` runs before `show()`. The `s_inUse` guard covers the re-entry case anyway, so it's only the line itself that does nothing.

dgirardeau:
Sorry, good catch, that was leftover of some tests I did.

dgirardeau:
Thinking twice, I added the 'setModal(true)' for the 2 plugins and at the right place, as an additional security, to prevent the user from messing with the entities during the computation.

Issue #1705, opened 2022-11-13
Update/Extend command line mode for RANSAC plugin


Checktaschu:
**Describe the feature you would like**

Extend the command line mode for the RANSAC Plugin to include the functionalities introduced in Version 2.12.0, e.g. max radii for various shapes, export leftover points etc.

**Describe alternatives you've considered**

-using the plugin manually, which is the opposite of what i am trying to achieve
-use the plugin without the added functionality which means passing on sphere and cylinder as geometric shapes
-calculate leftover points "manually" by removing all points that are part of detected shapes

**Is your feature request related to a problem? Please describe.**

-the algorithm very often detects imperfect planes as cylinders or spheres with huge radii, which obviously has been resolved with the added options in the GUI (basically what has been mentioned in the patchnotes)

**Additional context**

I am very new to github and basically everything. I hope this is the right place, if not, feel free to point it out.

dgirardeau:
That's the right place!

PR #2399, opened 2026-09-14
Apply RANSAC radius limits to refitted shapes

somtri:
Found while working on #1705, which asks for the dialog's options, such as the radius limits, on the `-RANSAC` command. The new options reached the same shape constructors the dialog uses, but detected spheres and cylinders still ignored the limits. The cause is in the vendored RANSAC_SD library, so the dialog is affected too. This PR is only the library fix. The command line options will follow in their own PR.

### What was wrong

In `RansacShapeDetector::Detect`, the limit check runs on `clone2`, a copy of the candidate taken before the least-squares refit. The refit itself is installed without a check. The limits were also lost on the way:

- `CylinderPrimitiveShape::LSFit` passes `(fit, m_maxRadius, m_maxLength)` to the constructor `(cylinder, minRadius, maxRadius, maxLength)`.
- `SpherePrimitiveShape::LSFit` and `TorusPrimitiveShape::LSFit` pass no limits.
- `TorusPrimitiveShapeConstructor::Construct` passes `(torus, m_allowAppleShaped, m_maxMinorRadius, m_maxMajorRadius)`, so each torus candidate holds the maximum radii in its minimum slots.
- `CylinderPrimitiveShape::CheckGeneratedShapeWithinLimits` does not test the minimum radius.

Tori fail the other way. The check reads the shifted limits from the candidate copy, where `MinorRadius() < m_minMinorRadius` is always true (the default maximum is `FLT_MAX`). Every torus refit is rejected, so tori respect the limits but are never refined.

### The change

One line in each of six files: `clone2.Shape(shape);` so the check sees the refit, the limits passed in declaration order to the three `LSFit` results and to the torus candidate, and `m_cylinder.Radius() >= m_minRadius` in the cylinder check.

`clone2.Shape(shape);` is the line that 405fcb0 ("preventing a crash in the QRansacSD plugin") removed together with three `shape->Release()` calls. Only the line comes back. `Candidate::m_shape` is a `MiscLib::RefCountPtr`, so the assignment adds a reference that `clone2` drops when it goes out of scope. The existing leak of `shape` (the `//shape->Release();` line) is unchanged and out of scope here.

With the default settings, tori are now refined. Planes, spheres and cylinders give the same results as before.

### Verified

Ubuntu 24.04, Qt 6.4.2, GCC 13.3, Release, `-DPLUGIN_STANDARD_QRANSAC_SD=ON`: builds and links, no warning names a file this PR touches, and `ctest` 2/2 passed. No ctest test covers this plugin, so the checks below are command line runs on synthetic clouds, with the radii read from the saved shapes. The detector seeds with `time(NULL)`, so the runs are spaced more than a second apart. Every run in both tables exited 0.

Default settings, this commit (a8cf7bd) against master (943535a), 5 runs each:

| Cloud | master | this PR |
|---|---|---|
| noisy plane (plane, sphere and cylinder enabled) | detected as a sphere, r = 123.6 to 130.0 | detected as a sphere, r = 123.7 to 123.9 |
| half cylinder, r = 2 | r = 1.994 to 2.058 | r = 1.989 to 2.021 |
| hemisphere, r = 2 | r = 1.99990 | r = 1.99989 to 1.99990 |
| torus, R = 4, r = 1 | r = 0.956 to 1.054, split into 3 and 2 tori in 2 runs | r = 0.999995 to 0.999999, one torus in every run |

The radius options do not exist on master, so the limit cases need the command line change from the follow-up PR. They ran on 74b423c (this commit plus that change) and, for the master column, on 943535a plus the same change. 10 runs each, counting the runs that returned a shape outside the limit:

| Case | master | this PR |
|---|---|---|
| half cylinder r = 2, `MAX_CYLINDER_RADIUS 1` | 10 | 0 (largest r = 0.999) |
| half cylinder r = 2, `MIN_CYLINDER_RADIUS 3` | 3 | 0 (smallest r = 3.004) |
| hemisphere r = 2, `MIN_SPHERE_RADIUS 3` | 10 | 0 (smallest r = 3.002) |
| torus, `MIN` and `MAX` of the minor and major radius (4 cases) | 0 | 0 |

Not verified: the dialog itself, because the build is headless; it calls the same constructors (`qRANSAC_SD.cpp:467-473`). `MAX_SPHERE_RADIUS` has no case that shows the fix, because the hemisphere yields no sphere under a maximum of 1. Cones were not tested (their limits are hidden in the dialog). Windows and macOS were not built or run locally.


dgirardeau:
Thanks for the fixes. I tested it on my side, and found other issues (especially a lot of parameters were not properly saved and restored on the next run, but the dialog structure was also a bit weird, there were some warnings, etc.)

somtri:
Thanks for testing this and for fixing the dialog issues directly. I pulled your commits into my local branch, now at 00ed75b. The shape-refit files from this PR are untouched, so there's nothing on my end to update.

Issue #2167, opened 2025-05-15
Dotbim


Isaina:
Can you support dotbim file format https://dotbim.net/ , compare two or compare stl and dotbim , i good with usdz workflow with apple ecosystem

dgirardeau:
Ah! Yet another triangular mesh format... and sadly no C++ library or code is provided. So someone will have to take that in charge.

I'll add it to the TODO list for now.

paireks:
Hey, I found this ticket by browsing some dotbim related topics, but if C++ code is necessary, then this might be a reference: https://github.com/bimdots-dev/DotbimArchicadAddOn

PR #2402, opened 2026-09-14
Add a dotBIM (.bim) import filter

somtri:
Addresses #2167 (import only; export and optional fields are left for a follow-up, see below).

## What

Adds a new IO filter that loads dotBIM (.bim) files, a small JSON mesh format
(https://dotbim.net/, schema v1.1.0) with no C++ reference implementation. Each file's
`elements` are loaded as separate meshes.

## The change

CloudCompare has no shared-instance geometry, so dotBIM's mesh-sharing model (one
`meshes[]` definition, referenced by multiple `elements[]` via `mesh_id`) can't be
represented directly. Each element is baked into its own mesh instead: its mesh's raw
vertices are transformed by that element's quaternion and translation (rotate then
translate, matching the format's own documentation) and stored as a fresh point cloud.

The filter uses Qt's own QJsonDocument for parsing, so no new dependency is added.

Scope for this PR: import only. The reporter's actual request ("compare two, or compare
stl and dotbim") only needs loading. Export, the optional `face_colors` field (per-face
color, takes precedence over the element's mandatory `color` since schema 1.1.0), and
the optional `info` metadata dictionaries are left for a follow-up.

Known limitation: unlike CloudCompare's other raw-coordinate importers (STL, OBJ, PTX),
this filter does not call `HandleGlobalShift` on load. dotBIM is a building-scale format
and the two real-world fixtures I tested stay well within float precision, but a dotBIM
file exported with real-world geo-referenced coordinates would not get CloudCompare's
usual large-coordinate shift/warning.

## Verified

Built on Ubuntu 24.04 (Docker), g++ 13.3.0, Qt 6.4.2. 403/403 build targets, 0 warnings
naming a touched file. clang-format 18.1.8 (matching CI's pin): 0 diff on both new files.

The qCoreIO plugin has no CTest coverage for any of its filters (including the existing
STL/OBJ/PTX filters), so I wrote a manual CLI verification suite: 13 checks, all passing.

- Two real dotBIM files (from the format's own repo, github.com/paireks/dotbim):
  `Pyramid.bim` (1 mesh, 1 element) loads exactly 1 mesh with 5 vertices / 6 triangles;
  `Cubes.bim` (1 mesh, 3 elements at 3 different positions) loads exactly 3 separate
  meshes, each 8 vertices / 12 triangles, at 3 distinct bounding boxes matching the
  file's own translation vectors exactly.
- A synthetic fixture with a real 90-degree rotation (neither real fixture has one, both
  use the identity quaternion): the loaded vertices match a hand-computed
  rotate-then-translate result exactly. This is the one check that would catch a
  quaternion field-order bug (dotBIM stores qx,qy,qz,qw; CloudCompare's
  `ccGLMatrixTpl::FromQuaternion` expects w,x,y,z).
- A file with valid JSON but a missing `elements` array, and a file that isn't JSON at
  all: both are rejected with a specific error message rather than a crash or silent
  no-op.

## Not verified

Windows and macOS were not built or run locally.

- [x] less than half of the code, or with careful human review and edits


plugins/core/IO/qCoreIO/src/DotbimFilter.cpp
```diff
@@ -0,0 +1,209 @@
...
+		mesh->setName(!elementType.isEmpty() ? elementType : (!guid.isEmpty() ? guid : "dotBIM element"));
+
+		if (!mesh->reserve(static_cast<unsigned>(geom.triangles.size())))
+		{
+			delete mesh; // also deletes 'vertices', already added as its child
+			return CC_FERR_NOT_ENOUGH_MEMORY;
+		}
+		for (const std::array<unsigned, 3>& tri : geom.triangles)
+		{
+			mesh->addTriangle(tri[0], tri[1], tri[2]);
```
dgirardeau: Why adding the triangles to a std::vector first, and then transferring them to a ccMesh entity, instead of adding them directly to a mesh instance? (technically with a ccMesh, you don't need to call reserve, but you can also use a step-wise reserve strategy as it's done in some other filters).

 Just make sure the whole process is within a std::catch block to detect memory issues.

plugins/core/IO/qCoreIO/src/DotbimFilter.cpp
```diff
@@ -0,0 +1,209 @@
...
+		{
+			ccLog::Warning(QString("[dotBIM] Element '%1' references unknown or empty mesh_id %2, skipped").arg(guid).arg(meshId));
+			continue;
+		}
+		const MeshGeometry& geom = *geomIt;
+
+		// rotate then translate (dotBIM applies rotation first, see the format's DeveloperTips)
+		QJsonObject rotationObj = elemObj.value("rotation").toObject();
+		double      quaternion[4] // (w, x, y, z)
+		    = {rotationObj.value("qw").toDouble(1.0),
```
dgirardeau:
```suggestion
		    {rotationObj.value("qw").toDouble(1.0),
```

plugins/core/IO/qCoreIO/src/DotbimFilter.cpp
```diff
@@ -0,0 +1,209 @@
...
+			continue;
+		}
+
+		MeshGeometry geom;
+
+		QJsonArray coordinates = meshObj.value("coordinates").toArray();
+		geom.vertices.reserve(static_cast<size_t>(coordinates.size() / 3));
+		for (int i = 0; i + 2 < coordinates.size(); i += 3)
+		{
+			geom.vertices.emplace_back(coordinates[i].toDouble(),
```
dgirardeau: same comments as for the triangles, I guess you could directly push the vertices in a ccPointCloud instance?

dgirardeau (changes requested):

somtri:
Updated the branch as requested. Each element now appends transformed coordinates directly to `ccPointCloud` and triangles directly to `ccMesh`. The loader returns `CC_FERR_NOT_ENOUGH_MEMORY` if allocation or child attachment fails. The quaternion initializer change is included.

Verification: clang-format 18.1.8 passed, the incremental Docker build succeeded, and the dotBIM CLI suite passed 13 checks.

AI assistance: less than half of the code, with careful human review and edits.

dgirardeau:
Thanks for the update. I've rebased and updated the code a bit more myself after testing it on my side.

dgirardeau (approved):

PR #2403, opened 2026-09-14
Add radius limit options to the -RANSAC command

### LINKED ISSUE #1705: Update/Extend command line mode for RANSAC plugin

somtri:
Fixes #1705

## What

The RANSAC Shape Detection plugin's dialog can restrict the radius of detected spheres,
cylinders and tori, but the `-RANSAC` command line mode could not. Adds
`MIN_SPHERE_RADIUS`, `MAX_SPHERE_RADIUS`, `MIN_CYLINDER_RADIUS`, `MAX_CYLINDER_RADIUS`,
`MIN_TORUS_MINOR_RADIUS`, `MAX_TORUS_MINOR_RADIUS`, `MIN_TORUS_MAJOR_RADIUS` and
`MAX_TORUS_MAJOR_RADIUS`, each taking a value. All are optional, and the default
behaviour (no limit applied) is unchanged.

Cone radius/length limits and the fitting/simplification switches were left out: cone
limits are hidden in the current GUI dialog and cylinder length isn't wired to anything,
so neither actually works there either, and the fitting/simplification switches weren't
asked for.

## Depends on #2399

Stacked on the just-merged #2399: the vendored RANSAC library dropped these limits after
the least-squares refit step, so before that fix the options in this PR would have
parsed and echoed correctly but silently done nothing. That's what #1705 turned out to
actually be about. With #2399 in, the limits are enforced.

## Verified

Built and tested in the same Docker environment as #2399 and #2402 (Ubuntu 24.04,
g++ 13.3.0, Qt 6.4.2), at the rebased commit d160459 (stacked on #2399's merge commit).
0 build errors, ctest 2/2. RANSAC samples randomly (seeded from wall-clock time), so
these check limits, never shape counts.

23 checks, all passing:
- No new option given: output unchanged from the pre-existing baseline (no radius line
  printed, grouped output still saves leftovers).
- `MAX_SPHERE_RADIUS 50` and `MIN_SPHERE_RADIUS 200` against a synthetic noisy plane: both
  cases detected zero spheres, so neither bound was actually exercised here. MIN_SPHERE_RADIUS
  enforcement has been confirmed against a real detected sphere before, in #2399's own
  debugging (a hemisphere fixture, not this PR's suite); MAX_SPHERE_RADIUS enforcement has
  never been confirmed against an actual detection.
- `MAX_CYLINDER_RADIUS`/`MIN_CYLINDER_RADIUS` against a half-cylinder fixture (r=2):
  every detected cylinder respects the given bound.
- All 4 torus options: parsed and echoed correctly (no torus fixture in this suite, so
  not exercised geometrically here, but the underlying refit-and-check path is the exact
  one verified end-to-end in #2399).
- An invalid number and a missing value after an option: both produce a specific error
  naming the option, not a crash or a silent default.

## Not verified

Windows and macOS were not built or run locally.


dgirardeau (approved):
Looks good!

PR #2405, opened 2026-09-15
Better 3DMouse Handling

rjanvier:
This PR proposes a better integration of #2339.

Improvements are two fold.

- First, it tries to improve some "leaks" / false positives at detection time.
The current detection mechanism, handled by `openDevice()`, is a bit too "relaxed". The current fallback path could literally take anything as a 3DMouse ! On my computer, my dongle is even detected as a 3DMouse. So I tried to improve the detection by two means. First, I check if the device matches a PID in a whitelist (because VID is not enough, since Logitech VID = some 3DCon VIDs). Obviously, the whitelist has to be maintained somehow, and as of today, it's maybe not even complete (I must admit I generated it with an LLM, hello Mistral ;)). And second, the PR forces the fallback path to do what the comment suggests: only trigger when `is3dConnexionHelperRunning()`.

@braunsi23, you are free to run any of the LLMs you used to propose modifications to this part. But on my side, I think the whitelist is a reasonable choice.

- Second, it tries to improve the build system. The build system now detects if the system already embeds HIDAPI and takes it by default. In the other case, resorting to the bundled HIDAPI submodule, a static lib is build is forced, so no more linking/install with hardcoded `.so` files (anyway, HIDAPI uses standard install CMake mechanisms with `GNU_INSTALL_DIR` so it didn’t have to be like that in the first place...).

What is not handled and should be discussed:

- There are a lot of references to a macOS only `3dConnexionHelper.app`. I don't know how it translates to other systems...
- I'm strongly in favor of opting out of this feature (OFF by default) since someone that compiles from scratch does not have to "pay" for this feature.


libs/CCAppCommon/devices/3dConnexion/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 if(WIN32)
 	option( OPTION_SUPPORT_3DMOUSE_WITH_3DxWARE "Build with 3DxWare (3D mouse support)" OFF )
 endif()
+option( OPTION_SUPPORT_3DMOUSE_WITH_HIDAPI "Build with hidapi (3D mouse support)" ON )
```
dgirardeau, line 4: Ok I think we don't need everyone to compile with this option (doesn't make sense). Just for the official versions on the 3 platforms ;)

libs/CCAppCommon/devices/3dConnexion/Mouse3DInput_hid.cpp
```diff
@@ -106,6 +106,55 @@ static const int c_buttonMap[] = {
...
+    0xc631, // SpaceMouse Pro Wireless
+    0xc632, // SpaceMouse Pro Wireless Receiver
+    0xc633, // SpaceMouse Enterprise
+    0xc635, // SpaceMouse Compact
+    0xc636, // SpaceMouse Module
+    0xc638, // SpaceMouse Pro Wireless BT (USB)
+    0xc63a, // SpaceMouse Wireless (Bluetooth)
+};
+
+static bool isKnownSpaceMouse(unsigned short vid, unsigned short pid)
```
dgirardeau:
```suggestion
static bool IsKnownSpaceMouse(unsigned short vid, unsigned short pid)
```

libs/CCAppCommon/devices/3dConnexion/Mouse3DInput_hid.cpp
```diff
@@ -135,7 +135,7 @@ static const unsigned short c_3dconnexionPIDs[] = {
     0xc63a, // SpaceMouse Wireless (Bluetooth)
 };
 
-static bool IsKnownSpaceMouse(unsigned short vid, unsigned short pid)
+static bool isKnownSpaceMouse(unsigned short vid, unsigned short pid)
```
dgirardeau: No Copilot... if you looked at the discussion and previous comments, the aim was to add the capital letter at the beginning of the function name (since it's a static function). The solution was to change the case where this method is called! Bad copilot, bad.

dgirardeau (changes requested):

libs/CCAppCommon/devices/3dConnexion/CMakeLists.txt
dgirardeau: @rjanvier Can you remove this line and still have the official Linux package including it?

dgirardeau:
Hi @braunsi23,

By any chance, would you be able to test the latest iteration on macOS in case we've broken something?

braunsi23:
Hi @dgirardeau, 

I tested the latest iteration on macOS (arm64): the build works fine. 👍 (One unrelated note: I first hit a compile error in `ccGeomFeaturesDlg.cpp` because my `CCCoreLib` submodule was stale — `git submodule update` fixed it.)

However, while reviewing the changes I found a regression introduced in 6c1267a7 ("Make 3DMouse detection more robust"): the fallback loop in `HIDWorker::openDevice()` lost its `hid_open_path()` call. The inner `if (m_handle)` can never be true since the block is only entered when `m_handle` is null, so the fallback is now dead code on all platforms — devices that don't expose the Multi-axis Controller usage (e.g. the wired SpaceMouse Compact, which relies on this fallback) can no longer be opened at all.

On top of that, the new `&& Is3dConnexionHelperRunning()` gate means the fallback never runs on Linux (there is no such process) and on macOS only when the driver daemon is running — but the Compact scenario is precisely *without* the helper.

The good news: the PID whitelist is exactly the right fix for the Logitech mouse issue reported on Linux, and it also makes the helper gate unnecessary — with the whitelist in place, the fallback can no longer grab an unrelated device, so it can safely run unconditionally again:

```cpp
	if (!m_handle)
	{
		cur = devs;
		for (; cur; cur = cur->next)
		{
			if (IsKnownSpaceMouse(cur->vendor_id, cur->product_id))
			{
				m_handle = hid_open_path(cur->path);
				if (m_handle)
				{
					m_devicePath = cur->path;
					QString name = (cur->product_string ? QString::fromWCharArray(cur->product_string) : QStringLiteral("3DConnexion device"));
					ccLog::Print(QString("[3D Mouse] Device: %1 (HID, fallback)").arg(name));
					break;
				}
			}
		}
	}
```

Minor remark on the whitelist itself: newer/unknown product IDs are now rejected even if they expose the multi-axis usage — that's a safe default, but we might see reports from owners of very recent devices.

I couldn't do a runtime test (no SpaceMouse at hand right now) — but @rjanvier or Paul could verify quickly with a wired Compact: it should fail to connect with the current iteration. Build-wise everything is fine on my side.

Cheers, Steffen

dgirardeau:
Thanks for the careful review (especially the missing call to hid_open_path!).

I've fixed the code accordingly:
- I added back the call to hid_open_path
- a 3dConnexion device with an unknown PID will still be considered as valid (I just added a warning message to invite the user to declare this new model)
- I removed the Is3dConnexionHelperRunning() gatekeeper

dgirardeau:
Hey @prascle. Would you be able to test this PR on your side by any chance?

dgirardeau (dismissed):

dgirardeau (approved):

Issue #1695, opened 2022-10-11
NaN values in las files


tiedl25:
**Describe the bug**
Since version 2.12.1 NaN values in scalar fields are automatically converted to a default value when saving to a las file. This is problematic for applications/scripts that rely on those values. There should be at least an option to set the default value manually. 

**How to reproduce**

**Expected behaviour**
An option to set the default value manually (also in commandline mode)

**Additional context**

**Your environment**
 - CC Version (or git hash/tag): 2.12.4
 - OS & Version: Windows 10

dgirardeau:
Do you mean that some applications rely on NaN values?

Because it didn't seem that LAS specifications allowed NaN values (and some SW tools reject LAS files with NaN values). That's why we automatically convert them now. Do you know if there's an official way to manage NaN values with LAS files maybe?

And if not, I believe that to 'solve' your issue we would probably prefer a more generic way to change NaN values before saving to LAS or any other format. For instance a special operation of the SF Arithmetic tool.

tiedl25:
https://github.com/ASPRSorg/LAS/wiki/Standard-ExtraByte-Definitions#published-standard-extrabyte-definitions
> As a general guideline, the maximum possible value is a good default NODATA value, except in the cases when zero has the same meaning as null and is therefore preferable.

For example the M3C2-distances parameter sometimes contains nan values but also zeros, which do not have the same meaning and shouldn't be treated the same way.
So it would be great to have an option in commandline mode to change this default value or leave it as NaN.

dgirardeau:
Yes, but what would be the default value of NODATA in a general case? (as I see that it depends on the application, and quite often '0' in the examples...).

tiedl25:
Agreed, there is no one-for-all (applications) general case. Therefore, the NoData value should be determined based on the value range of each attribute, e.g. > maximum possible value, as documented here: https://github.com/ASPRSorg/LAS/wiki/Standard-ExtraByte-Definitions#published-standard-extrabyte-definitions.
Looking at the LAS standard, there is a NoData flag available for extra byte attributes, which should be used to store the NoData value (Table 25 in this document: http://www.asprs.org/wp-content/uploads/2019/03/LAS_1_4_r14.pdf).
When reading LAS files, the NoData flag must be checked and all values occurring in the corresponding attribute are to be considered NoData/nan.

PR #2407, opened 2026-09-19
Preserve LAS Extra Bytes NoData semantics

somtri:
## What does this PR do?

Addresses #1695.

For an Extra Bytes field that already declares NoData, saving a scalar-field NaN now writes that raw NoData value. Loading a raw value that matches declared NoData restores NaN, including for fields with scale metadata.

Fields without declared NoData metadata keep their current conversion. This PR does not choose a new sentinel value.

- [ ] less than half of the code, or with careful human review and edits
- [x] more than half of the code, with limited human review and edits

## Testing
- [x] I have tested this PR on Linux
- [ ] I have tested this PR on macOS


plugins/core/IO/qLASIO/src/LasScalarFieldLoader.cpp
```diff
@@ -367,19 +381,6 @@ void LasScalarFieldLoader::handleOptionsFor(const LasExtraScalarField& extraFiel
 	assert(extraField.numElements() <= 3);
 	for (unsigned dimIndex = 0; dimIndex < extraField.numElements(); ++dimIndex)
 	{
-		if (extraField.noDataIsRelevant())
-		{
-			auto noDataValue = ParseValueOfTypeAs<T, T>(static_cast<const uint8_t*>(extraField.noData[dimIndex]));
-			if (noDataValue == inputValues[dimIndex])
-			{
-				outputValues[dimIndex] = ccScalarField::NaN();
```
tmontaigu, line 375: Could have just put a continue here ?

somtri: Yes, that's simpler. The original block in `handleOptionsFor` is back, with a `continue` after the NaN assignment.

Going back to it also showed a bug in the previous version. The LAS spec stores `no_data` upcast to 8 bytes (a double for float fields). Comparing and copying only `elementSize()` bytes was therefore wrong for f32 fields. With a NoData of -9999, a NaN was written as 0.0 and a real 0.0 was read back as NaN. For f32 fields the saver now converts the double NoData value to float. Integer and f64 fields keep the byte copy.

Verification: clang-format 18.1.8 passed and the qLAS plugin built. A round-trip harness passed 54 checks covering all 10 Extra Bytes types (NaN for each type, plus a real 0.0 in an f32 field, with and without scale). The same harness fails 3 checks on the previous commit, all on f32.

AI assistance: more than half of the code, with careful human review and edits.

plugins/core/IO/qLASIO/src/LasScalarFieldSaver.cpp
```diff
@@ -213,6 +215,24 @@ void LasScalarFieldSaver::handleExtraFields(size_t pointIndex, laszip_point& poi
...
 		for (unsigned i = 0; i < extraField.numElements(); i++)
 		{
+			if (extraField.noDataIsRelevant() && std::isnan(values[i]))
+			{
+				// no_data is upcast to 8 bytes: a double for floating point types,
+				// and a 64-bit integer (whose first bytes hold the value) otherwise
+				if (extraField.type == LasExtraScalarField::f32)
+				{
+					double noDataValue;
+					std::memcpy(&noDataValue, extraField.noData[i], sizeof(noDataValue));
```
dgirardeau:
```suggestion
					const double& noDataValue = *reinterpret_cast<const double*>(extraField.noData[i]);
```

PR #2408, opened 2026-09-20
Don't abort when saving a cloud with more than 341 scalar fields to LAS

capCafu:
## What does this PR do?

Saving a cloud with more than 341 scalar fields to LAS aborts the application.

```
ASSERT: "byteArray.size() == vlr.record_length_after_header" in LasExtraScalarField.cpp, line 310
Aborted (core dumped)
```

`InitExtraBytesVlr` computes the Extra Bytes VLR size into a `laszip_U16`, so 342 fields gives 65664 which wraps to 128. The buffer is allocated at the wrapped size and the assert catches the mismatch. qLASIO sets `QT_FORCE_ASSERTS`, so this happens in release builds too.

You can hit it without a LAS file, just load an ASCII cloud with 345 columns and save it as LAS.

The save now stops with a message instead. It does not lift the limit, anything past 341 needs the descriptor in an EVLR and that is a bigger change.

- [x] less than half of the code, or with careful human review and edits

## Testing
- [x] I have tested this PR on Linux
- [ ] I have tested this PR on macOS

341 still saves and reads back. 342 and 500 are refused. GCC 15.3, Qt 6.9.3, Ubuntu 24.04, no new warnings.


plugins/core/IO/qLASIO/include/LasExtraScalarField.h
```diff
@@ -81,6 +81,12 @@ class LasExtraScalarField
 	friend QDataStream& operator>>(QDataStream& dataStream, LasExtraScalarField& extraScalarField);
 	friend QDataStream& operator<<(QDataStream& dataStream, const LasExtraScalarField& extraScalarField);
 
+  public:
+	/// The payload length of a VLR is stored on 16 bits, and each field is described
+	/// by 192 bytes, so an "Extra Bytes" VLR cannot describe more fields than this.
+	/// Files with more of them have to store the descriptor in an EVLR.
+	static constexpr size_t MAX_EXTRA_FIELDS_IN_VLR = 65535 / 192;
```
dgirardeau: Maybe it would be also clearer to add a definition like VLR_FIELD_SIZE_BYTES = 192?

plugins/core/IO/qLASIO/src/LasExtraScalarField.cpp
```diff
@@ -295,17 +296,28 @@ LasExtraScalarField::Kind LasExtraScalarField::kind() const
...
-	vlr.record_length_after_header = 192 * static_cast<laszip_U16>(extraFields.size());
+	vlr.record_id = 4;
+
+	// the length of a VLR payload is stored on 16 bits, so the descriptor cannot
+	// describe more fields than that. The caller is expected to have checked this
+	// already, we only make sure we never write past the end of the buffer.
+	const size_t fieldCount = std::min(extraFields.size(), MAX_EXTRA_FIELDS_IN_VLR);
+	if (fieldCount != extraFields.size())
+	{
+		ccLog::Warning("[LAS] Only the first %u extra scalar fields can be described (out of %u)",
```
dgirardeau:
```suggestion
		ccLog::Warning(QString("[LAS] Only the first %1 extra scalar fields can be described (out of %2)").arg(fieldCount).arg(extraFields.size()));
```

plugins/core/IO/qLASIO/src/LasIOFilter.cpp
```diff
@@ -913,6 +913,26 @@ CC_FILE_ERROR LasIOFilter::saveToFile(ccHObject* entity, const QString& filename
...
+		// the saver adds one field per normal component on top of the ones selected here
+		size_t extraFieldCount = params.extraFields.size();
+		if (params.shouldSaveNormalsAsExtraScalarField && pointCloud->hasNormals())
+		{
+			extraFieldCount += 3;
+		}
+
+		if (extraFieldCount > LasExtraScalarField::MAX_EXTRA_FIELDS_IN_VLR)
+		{
+			ccLog::Error("[LAS] Cannot save more than %u extra scalar fields (%u requested)",
```
dgirardeau:
```suggestion
			ccLog::Error(QString("[LAS] Cannot save more than %1 extra scalar fields (%2 requested)").arg(LasExtraScalarField::MAX_EXTRA_FIELDS_IN_VLR).arg(extraFieldCount));
```

plugins/core/IO/qLASIO/src/LasIOFilter.cpp
```diff
@@ -913,6 +913,26 @@ CC_FILE_ERROR LasIOFilter::saveToFile(ccHObject* entity, const QString& filename
...
+			             static_cast<unsigned>(LasExtraScalarField::MAX_EXTRA_FIELDS_IN_VLR),
+			             static_cast<unsigned>(extraFieldCount));
+			return CC_FERR_NOT_IMPLEMENTED;
```
dgirardeau: Won't the fields be truncated by the code above? Maybe a warning would be sufficient? (or a question to the user whether they want to proceed or not?)

capCafu: I tried that and the file comes out malformed. `point_data_record_length` counts all the fields but the descriptor only covers the first 341, so with 342 fields the header says 1380 extra bytes per point while the descriptors cover 1372.

It would work if the surplus was dropped from `parameters.extraFields` before the size is computed, not just inside `InitExtraBytesVlr`. Then the file is valid and the warning says what was lost.

Either is fine by me, refuse or truncate and warn. Tell me which you prefer.

Taking the other three.

dgirardeau: Maybe truncate and warn, because there's no solution to do it manually with CloudCompare. Well, technically you can delete SFs one by one, but I assume in this case the user would potentially have a lot more SFs, and removing them one by one until one reaches 341 may be tedious.

capCafu: Done. The surplus is dropped from `params.extraFields` before the saver works out the point record length, so the file stays consistent. 500 fields now saves and reads back with 341.

One thing worth knowing, the warning says 338 rather than 341 when the cloud has normals, because the saver adds Nx Ny Nz as extra fields afterwards and I leave room for them. Say if you would rather the message mentioned 341.

dgirardeau: No that's clearer for the user (as only the scalar fields are dropped).

Thanks!

dgirardeau (approved):

Issue #2321, opened 2026-04-23
Opening a .PLY file by double-clicking fails on MacOS (regression in CC 2.14)


zestyping:
**Describe the bug**
The bug described in https://github.com/CloudCompare/CloudCompare/issues/1883 has reappeared in CloudCompare 2.14 beta (Jan 28 2026) on MacOS 15.7.3.  I see exactly the behaviour described in #1883.

**How to reproduce**
1. Ensure CloudCompare is not already open.
2. Double-click on a .PLY file.
3. CloudCompare launches and shows the message:
> [Load] Can't guess file format: unhandled file extension 'ply'

<img width="372" height="309" alt="Image" src="https://github.com/user-attachments/assets/086117f4-119b-4dba-bf65-94b4c086c9cf" />

**Expected behaviour**
The .PLY file opens in CloudCompare.

**Additional context**
If CloudCompare is already open, double-clicking on a .PLY file opens properly in CloudCompare.

**Your environment**
CloudCompare: 2.14 beta (Jan 28 2026) [MacOS 64-bit]
Operating system: MacOS 15.7.3
QT version: Compiled with Qt 6.9.3

<img width="617" height="411" alt="Image" src="https://github.com/user-attachments/assets/5e58f089-f761-453b-b113-01711b3d43d1" />

PR #2409, opened 2026-09-22
Open files double-clicked in the macOS Finder when CC is not running yet

capCafu:
## What does this PR do?

- When CC is launched by opening a file on macOS, the FileOpen event arrives before the I/O filters are registered and the file fails to load with "Can't guess file format" (#2321). 
- The files are now kept until the main window is ready and opened from main.cpp. Opening a file while CC is already running works as before.

- [x] no AI, or only a minor contribution
- [ ] less than half of the code, or with careful human review and edits

## Testing

- [ ] I have tested this PR on Linux
- [x] I have tested this PR on macOS


rjanvier (approved):
It's ok for me, thanks. Maybe for the sake of semantics, you could have used `const filename` where possible, but anyway the current codebase is not consistent in this matter.

dgirardeau:
Oops, I didn't have the time to review this PR. Shouldn't this be done for ccViewer as well? @capCafu @rjanvier

capCafu:
> Oops, I didn't have the time to review this PR. Shouldn't this be done for ccViewer as well? @capCafu @rjanvier

Sorry for merging before you had a chance to look at it. ccViewer has same problem too, I will create new PR for it today.

capCafu:
> Oops, I didn't have the time to review this PR. Shouldn't this be done for ccViewer as well? @capCafu @rjanvier

I checked ccViewer and I could not reproduce the problem there. On an M2 Pro with macOS 15.7.8 the file opens correctly at launch.

The difference is the splash screen. In CloudCompare it lets the FileOpen event be delivered before the I/O filters are registered. ccViewer has no splash screen, so its first window is shown after the viewer is set and the event always arrives late enough.

The handler still returns false when the viewer is not set, so a file would be dropped if the event ever arrived earlier. I have a small change that keeps the file until the viewer is ready, like the one here. Do you want it as a PR, or would you rather leave ccViewer as it is since there is nothing to reproduce?

dgirardeau:
Yes, maybe out of consistency? And in case things behave a little bit differently on other OSes?

